PR #8, opened 2024-12-08
CCIP-4474 manual exec lbtc

bukata-sa:
Add lbtc manual exec feature.
If destTokenData is more than 32 bytes - attestation is disabled onchain
else - call lombard attestation api for a proof and include it as offchain token data


src/lib/offchain.ts
```diff
@@ -94,6 +109,79 @@ async function getUsdcTokenData(
...
+    )
+  }
+  const attestation = response.attestations.find((att) => att.message_hash === payloadHash)
+  if (attestation == null) {
+    throw new Error(
+      'Could not find requested LBTC attestation with hash:' +
+        payloadHash +
+        ' in response: ' +
+        JSON.stringify(response, null, 2),
+    )
```
andrevmatos, line 138: It's better to not throw inside this function, and instead return `undefined`, otherwise, a (malicious?) TokenPool emitting events similar to your LBTC's can DoS your manualExec algo. The USDC's approach above is: let's be optimist, do as many checks as you can/need, perform the request, but if anything fail, let it pass through undefined, and possibly (if it was a transient error) fail at estimation time (e.g. dest TP requires the attestation that we failed to fetch), but will eventually succeed if/when the user retries, if it should.

src/lib/offchain.ts
```diff
@@ -1,21 +1,36 @@
...
+  mainnet: 'https://bridge-manager.lombard.finance',
+  testnet: 'https://bridge-manager.staging.lombard.finance',
+}
 
 type AttestationResponse =
   | { error: 'string' }
   | { status: 'pending_confirmations' }
   | { status: 'complete'; attestation: string }
 
+type LombardAttestation = { status: string; message_hash: string; attestation: string }
```
andrevmatos, line 31: Nit: I see you compare `status` here with some constant. You don't need to extract that const in a global var, but if you can get a small list of possible status, it'd be nice to hardcode them as a union of string literals here, as it'll aid the constant comparison down there.
```suggestion
type LombardAttestation = { status: string; message_hash: string; attestation: string }
```

andrevmatos (changes requested):
Getting there!

src/lib/offchain.ts
```diff
@@ -114,21 +205,27 @@ export async function fetchOffchainTokenData(
     request.tx.logs.find(
       ({ topics, index }) => topics[0] === request.log.topics[0] && index < request.log.index,
     )?.index ?? -1
-  const requestLogs = request.tx.logs.filter(
+  const usdcRequestLogs = request.tx.logs.filter(
```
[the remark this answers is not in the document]
bukata-sa: This logic doesn't work for LBTC
DepositToBridge LBTC event can be before CCIPSendRequested (if transfer sent from CCIP Router) or after CCIPSendRequested (if transfer sent from Lombard's Bridge)
So we can't make any assumptions about relative log positions

src/lib/offchain.ts
bukata-sa: done

src/lib/offchain.ts
bukata-sa: done

src/lib/offchain.ts
andrevmatos: Is it? I still see a bunch of throws in this function

src/lib/offchain.ts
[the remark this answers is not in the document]
andrevmatos: Ok, I see. Eventually, we may move the assumption check of the log happening before the OnRamp event (i.e. TokenPool logs) to the USDC function; but it's fine to leave it like this for now. Thank you

andrevmatos (approved):
Good job on this implementation!

andrevmatos:
Can you add a CHANGELOG entry, please?

andrevmatos (approved):

PR #11, opened 2025-01-05
Supported tokens

aelmanaa:
# Add CCIP Supported Tokens Discovery Command

This PR adds a new command to discover and display tokens that are supported for CCIP cross-chain transfers between two chains.

## Features
- New `supported-tokens` command to check CCIP-supported tokens
- Pagination support for large token sets
- Parallel processing of pool support checks
- Token metadata fetching (name, symbol, decimals)
- Comprehensive error handling and reporting
- Multiple output formats (pretty, json, log)

## Implementation Details
- Moved command to dedicated file in `src/commands/supported-tokens.ts`
- Added configurable constants for performance tuning
- Structured metadata output including:
  - Chain information
  - Statistics
  - Timestamps
  - Success/failure counts
- Error collection for failed token checks
- Rate-limiting consideration with parallel pool checks

### Check supported tokens between chains
`npx tsx src/index.ts supported-tokens <source> <router> <dest> [--format <format>]`

### Example

`npx tsx src/index.ts supported-tokens ethereum-mainnet 0x80226fc0Ee2b096224EeAc085Bb9a8cba1146f7D avalanche-mainnet`


src/index.ts
```diff
@@ -357,6 +358,39 @@ async function main() {
           .finally(() => providers.destroy())
       },
     )
+    .command(
+      'supported-tokens <source> <router> <dest>',
```
andrevmatos, line 362: Can we use kamelCase for commands, for consistency?
I think yargs may allow both calls anyway

src/index.ts
```diff
@@ -357,6 +358,39 @@ async function main() {
...
+            type: 'string',
+            demandOption: true,
+            describe: 'router contract address on source',
+            coerce: getAddress,
+          })
+          .positional('dest', {
+            type: 'string',
+            demandOption: true,
+            describe: 'Destination chain name or id',
+            example: 'optimism-testnet-goerli',
```
andrevmatos, line 382:
```suggestion
            example: 'ethereum-testnet-sepolia-optimism-1',
```

src/abi/TokenPool_1_5.ts
andrevmatos, line 1: BurnMintTokenPool abi is already present, and is a super-set of TokenPool, I don't think this is needed

src/lib/types.ts
andrevmatos, line 1: If these types are not used inside the library, or somewhat needed here at least for completeness/consistency, I think it's better to put them at `commands/types.ts`.
`lib` is more intended for external consumers (e.g. Explorer, eventually), and these types look quite specific to the new command being added

andrevmatos (changes requested):
Some comments.
After they're addressed, we can merge, and I'll make a follow-up to migrate the rest of the commands to the commands module.
Thank you!

src/commands/supported-tokens.ts
```diff
@@ -0,0 +1,287 @@
...
+    `[INFO] Fetching all registered tokens from "${chainNameFromId(sourceChainId)}" using TokenAdminRegistry`,
+  )
+
+  do {
+    console.log(`[INFO] Fetching batch: offset=${startIndex}, limit=${maxCount}`)
+    tokensBatch = [...(await registry.getAllConfiguredTokens(startIndex, BigInt(maxCount)))]
+    totalScanned += tokensBatch.length
+    console.log(`[INFO] Found ${tokensBatch.length} tokens (total scanned: ${totalScanned})`)
+
+    if (tokensBatch.length > 0) {
```
andrevmatos, line 106: Could/should some of these "big steps" be split into local util functions (in same module, until they're needed by other modules).

src/commands/supported-tokens.ts
```diff
@@ -1,7 +1,9 @@
 /* eslint-disable @typescript-eslint/no-base-to-string */
-import { type Addressable, Contract } from 'ethers'
+import { type Addressable, type JsonRpcApiProvider, Contract } from 'ethers'
 import type { TypedContract } from 'ethers-abitype'
 
+import { chunk } from 'lodash-es'
```
andrevmatos, line 30: Could we please not pull whole `lodash-es` new dependency only for this small util? It may be easy enough to code it inline or write a function here for this

src/lib/types.ts
```diff
@@ -21,36 +25,129 @@ export type CCIPMessage = AbiParametersToPrimitiveTypes<
...
+export type CCIPContractTypeTokenPool =
+  | CCIPContractTypeBurnMintTokenPool
+  | CCIPContractTypeLockReleaseTokenPool
+  | CCIPContractTypeLockReleaseTokenPoolAndProxy
+  | CCIPContractTypeUSDCTokenPool
+  | CCIPContractTypeBurnWithFromMintRebasingTokenPool
+  | CCIPContractTypeBurnMintTokenPoolAndProxy
+  | CCIPContractTypeBurnMintTokenPoolWithTax
+  | CCIPContractTypeBurnWithFromMintTokenPoolAndProxy
+  | CCIPContractTypeBurnWithFromMintTokenPool
```
andrevmatos, line 91: Do we really need these contract types here?
When I created them, it was with the assumption that token pools were not strictly part of the ccip core contracts. They can even be implemented by 3Ps, and not comply with our interfaces at all.
I'd rather leave them to some self-contained command, and not tie them directly into lib.

andrevmatos (approved):

PR #19, opened 2025-03-12
Add Aptos execution hasher

RodrigoAD:
- Adds the ability to calculate the merkle proofs based on CCIP messages targeting Aptos. This enables users calling the Aptos Offramp `manuallyExecute` function to calculate part of the execution report.
- Also refactors the structure of the `/hasher` dir to be more explicit between chains, and adds some more tests
- Adds the Aptos Chain selectors

### DIFF
--- a/src/lib/execution.ts
+++ b/src/lib/execution.ts
@@ -10,7 +10,7 @@ import {
 import type { TypedContract } from 'ethers-abitype'
 
 import Router from '../abi/Router.js'
-import { Tree, getLeafHasher, proofFlagsToBits } from './hasher/index.js'
+import { Tree, getLeafHasher, proofFlagsToBits } from './hasher'
 import {
   type CCIPContract,
   type CCIPExecution,
@@ -42,20 +42,20 @@ import {
  * @param merkleRoot - Optional merkleRoot of the CommitReport, for validation
  * @returns ManualExec report arguments
  **/
-export function calculateManualExecProof(
-  messagesInBatch: readonly CCIPMessage[],
-  lane: Lane,
+export function calculateManualExecProof<V extends CCIPVersion>(
+  messagesInBatch: readonly CCIPMessage<V>[],
+  lane: Lane<V>,
   messageIds: string[],
   merkleRoot?: string,
 ): {
-  messages: CCIPMessage[]
+  messages: CCIPMessage<V>[]
   proofs: string[]
   proofFlagBits: bigint
 } {
   const leaves: string[] = []
   const hasher = getLeafHasher(lane)
   const prove: number[] = []
-  const messages: CCIPMessage[] = []
+  const messages: CCIPMessage<V>[] = []
   const seen = new Set<string>()
 
   messagesInBatch.forEach((message, index) => {
@@ -71,10 +71,6 @@ export function calculateManualExecProof(
       )
     }
     // Hash leaf node
-    // TODO: msg here could be of any type (1.5 or 1.6), but we are using a specific version hasher. This will throw at runtime if encounters different message versions
-    // hasher type should detect the msg version
-    // If we want to be able to hash any message type, we should have a hasher per version
-    // If we don't want to hash any message type, `messagesInBatch` should only accept one type of message
     leaves.push(hasher(msg))
     const msgId = msg.header.messageId
     seen.add(msgId)
--- a/src/lib/execution.ts
+++ b/src/lib/execution.ts
@@ -10,7 +10,7 @@ import {
 import type { TypedContract } from 'ethers-abitype'
 
 import Router from '../abi/Router.js'
-import { Tree, getLeafHasher, proofFlagsToBits } from './hasher'
+import { Tree, getLeafHasher, proofFlagsToBits } from './hasher/index.js'
 import {
   type CCIPContract,
   type CCIPExecution,
@@ -42,7 +42,7 @@ import {
  * @param merkleRoot - Optional merkleRoot of the CommitReport, for validation
  * @returns ManualExec report arguments
  **/
-export function calculateManualExecProof<V extends CCIPVersion>(
+export function calculateManualExecProof<V extends CCIPVersion = CCIPVersion>(
   messagesInBatch: readonly CCIPMessage<V>[],
   lane: Lane<V>,
   messageIds: string[],
--- a/src/lib/execution.ts
+++ b/src/lib/execution.ts
@@ -71,6 +71,10 @@ export function calculateManualExecProof(
       )
     }
     // Hash leaf node
+    // TODO: msg here could be of any type (1.5 or 1.6), but we are using a specific version hasher. This will throw at runtime if encounters different message versions
+    // hasher type should detect the msg version
+    // If we want to be able to hash any message type, we should have a hasher per version
+    // If we don't want to hash any message type, `messagesInBatch` should only accept one type of message
     leaves.push(hasher(msg))
     const msgId = msg.header.messageId
     seen.add(msgId)
--- a/src/lib/hasher/common.test.ts
+++ b/src/lib/hasher/common.test.ts
@@ -1,4 +1,4 @@
-import { hashInternal } from './common'
+import { hashInternal } from './common.js'
 
 describe('common hash', () => {
   it('should hash internal values', () => {
--- a/src/lib/hasher/common.test.ts
+++ b/src/lib/hasher/common.test.ts
@@ -1,0 +1,10 @@
+import { hashInternal } from './common'
+
+describe('common hash', () => {
+  it('should hash internal values', () => {
+    const a = '0x01'
+    const b = '0x02'
+    const result = hashInternal(a, b)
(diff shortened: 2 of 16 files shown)


src/lib/execution.ts
```diff
@@ -69,6 +69,10 @@ export function calculateManualExecProof(
       )
     }
     // Hash leaf node
+    // TODO: msg here could be of any type (1.5 or 1.6), but we are using a specific version hasher. This will throw at runtime if encounters different message versions
```
RodrigoAD, line 72: @andrevmatos I saw this issue when adding the Aptos type. As I mention in the comment, I'm not sure if we want to accept any type of CCIPMessage in the batch. If so, we should have the specific version hasher for each message version, if not, we should type `messagesInBatch` to accept only one version.

I test the issue [here](https://github.com/smartcontractkit/ccip-tools-ts/pull/19/files#diff-52fd748922ee0f1d394f7725994a3ed32265ba8f20405e25ab0e6230060c4366R166)

src/index.ts
```diff
@@ -22,7 +22,7 @@ import { Providers } from './providers.js'
 util.inspect.defaultOptions.depth = 6 // print down to tokenAmounts in requests
 // generate:nofail
 // `const VERSION = '${require('./package.json').version}-${require('child_process').execSync('git rev-parse --short HEAD').toString().trim()}'`
-const VERSION = '0.2.0-1149e42'
+const VERSION = '0.2.0-2c49195'
```
andrevmatos, line 25: Nit: No need to commit these bumps, they're mostly relevant for release PRs

src/lib/types.ts
```diff
@@ -108,6 +108,31 @@ export type CCIPMessage<V extends CCIPVersion = CCIPVersion> = V extends
...
 
+type AptosTokenTransfer = {
+  sourcePoolAddress: string
+  destTokenAddress: string
+  destGasAmount: bigint
+  extraData: string
+  amount: bigint
+}
+// TODO Rodrigo: Refactor it to be included in CCIPMessage type
+export type AptosCCIPMessage = {
```
andrevmatos, line 119: This type should be compatible with `CCIPMessage<V1_6>`, no need to have a specific one for Aptos; you can use `decodeMessage` to juggle properties around to be compatible here, if needed

src/lib/hasher/hasher.ts
```diff
@@ -168,14 +174,82 @@ export function getLeafHasher<V extends CCIPVersion = CCIPVersion>({
   destChainSelector,
   onRamp,
   version,
-}: Lane<V>): LeafHasher {
+}: Lane<V>): LeafHasher<V> {
   switch (version) {
     case CCIPVersion.V1_2:
     case CCIPVersion.V1_5:
-      return getV12LeafHasher(sourceChainSelector, destChainSelector, onRamp)
+      return getV12LeafHasher(sourceChainSelector, destChainSelector, onRamp) as LeafHasher<V>
```
andrevmatos, line 181: No need to cast here, your functions should be able to return the versioned LeafHashers on their own

src/lib/hasher/hasher.ts
```diff
@@ -168,14 +174,82 @@ export function getLeafHasher<V extends CCIPVersion = CCIPVersion>({
...
-      return getV16LeafHasher(sourceChainSelector, destChainSelector, onRamp)
+      return getV16LeafHasher(sourceChainSelector, destChainSelector, onRamp) as LeafHasher<V>
     default:
       throw new Error(`Unsupported CCIP version: ${version}`)
   }
 }
+
+export type AptosLeafHasher = (message: AptosCCIPMessage) => string
+
+export const makeAptosHasher =
```
andrevmatos, line 191: Nit: please, make naming compatible with the other leafhashers; you can assume you only need to implement it for V1_6+

src/lib/hasher/hasher.ts
```diff
@@ -168,14 +174,82 @@ export function getLeafHasher<V extends CCIPVersion = CCIPVersion>({
...
+  (message: AptosCCIPMessage): string =>
+    hashAptosMessage(message, hashAptosMetadata(sourceChainSelector, destChainSelector, onRamp))
+
+const encode = (target: string, value: string | bigint | number) =>
+  defaultAbiCoder.encode([target], [value])
+
+/**
+ * Encodes dynamic bytes without the extra length prefix.
+ * The default ABI encoding for type "bytes" is: 32-byte length prefix + the actual data padded.
+ * This helper strips the 32-byte length prefix, mimicking the Move implementation.
```
andrevmatos, line 202: EVM also has something similar, but this is usually taken care by the abi decoder;
i.e. the abi.encoded words also have the length in first word, then the rest of the array; but when abi decoding (e.g. from ethers), we get only the bytearray that we expect, as the client/parser takes care of this padding.

Shouldn't we expect Aptos client/decoder to also do this for us?

src/lib/hasher/hasher.ts
```diff
@@ -168,14 +174,82 @@ export function getLeafHasher<V extends CCIPVersion = CCIPVersion>({
...
+    encode('uint256', message.gasLimit),
+    encode('uint64', message.header.nonce),
+  ])
+
+  const tokenHash = concat([
+    encode('uint256', message.tokenAmounts.length),
+    ...message.tokenAmounts.map((token) =>
+      concat([
+        encodeRawBytes(token.sourcePoolAddress),
+        encode('bytes32', zeroPadValue(token.destTokenAddress, 32)),
```
andrevmatos, line 224: encode here is noop: the output is the same as zeroPadValue's

andrevmatos (commented):
Initial pass

src/lib/hasher/hasher.ts
RodrigoAD: They do, but `getLeafHasher` does not know if the function is actually returning the same version `V`

src/lib/execution.ts
```diff
@@ -42,20 +42,20 @@ import {
  * @param merkleRoot - Optional merkleRoot of the CommitReport, for validation
  * @returns ManualExec report arguments
  **/
-export function calculateManualExecProof(
-  messagesInBatch: readonly CCIPMessage[],
-  lane: Lane,
+export function calculateManualExecProof<V extends CCIPVersion>(
```
andrevmatos, line 45: Let's not require explicitly the template type to be passed
```suggestion
export function calculateManualExecProof<V extends CCIPVersion = CCIPVersion>(
```

src/lib/selectors.ts
```diff
@@ -382,4 +384,17 @@ const selectors: Record<number, { readonly selector: bigint; readonly name?: str
   },
   // end:generate
 }
+
+export const aptosSelectors: Selectors = {
```
andrevmatos, line 388: Please, add the `generate:` comments to generate from [here](https://github.com/smartcontractkit/chain-selectors/blob/main/selectors_aptos.yml), so we make sure these are kept up-to-date

src/lib/selectors.ts
```diff
@@ -382,4 +384,17 @@ const selectors: Record<number, { readonly selector: bigint; readonly name?: str
...
+  '1': { name: 'aptos-mainnet', selector: 4741433654826277614n },
+  '2': { name: 'aptos-testnet', selector: 743186221051783445n },
+  '4': { name: 'aptos-localnet', selector: 4457093679053095497n },
+}
+
+const selectors: Selectors = { ...evmSelectors, ...aptosSelectors }
```
andrevmatos, line 394: Hmm, Aptos chainIds will conflict with EVM's, and override them here.

We will also need to adjust all functions accepting a `chainId: number` (e.g. [chainSelectorFromId](https://github.com/smartcontractkit/ccip-tools-ts/blob/3af48d6774fe040c8b64d0038e84ccb7f43734d4/src/lib/utils.ts#L251-L253), to be able to handle conflicting chainIds on different families.

I think we could add a `ChainFamily` enum, and make the selectors `Record` have a `${ChainFamily}:${number}` key (`string`), e.g. `evm:1` for `ethereum-mainnet`.

But I really wanted to minimize architecture changes in this PR; it was mostly intended to add just an Aptos hasher, and is growing in scope and changes.

src/lib/hasher/hasher.ts
```diff
@@ -1,180 +1,29 @@
...
-      encodedTokens,
-      fixedSizeValues,
-      packedValues,
-      metadataInput,
-    })
-
-    return keccak256(packedValues)
-  }
-}
+import { isAptosChain } from '../selectors'
```
andrevmatos, line 1: Please, add `.js` suffixes to the import filepaths

src/lib/hasher/common.test.ts
```diff
@@ -0,0 +1,10 @@
+import { hashInternal } from './common'
```
andrevmatos, line 1: .js suffix

src/lib/selectors.ts
RodrigoAD: > Hmm, Aptos chainIds will conflict with EVM's, and override them here.
Right! Let me fix that

To avoid more major changes in this PR we can just export the `isAptosChain` function, as is the only one used, and keep `selectors` as evm only for now.

We can add support for different chain families later on

andrevmatos (approved):

PR #21, opened 2025-03-13
getProvider function exported from lib

marek-sadura:
Refactored logic to get a provider instance, exported from lib.


marek-sadura:
Turns out we will not need this

PR #22, opened 2025-03-17
Fetch offchain data aptos

RodrigoAD:


src/lib/offchain.ts
```diff
@@ -257,3 +268,134 @@ export async function fetchOffchainTokenData(
   }
   return offchainTokenData
 }
+
+// ----- Logic Replication using a chain agnostic input -----
+export type FetchOffchainTokenDataInput = {
```
RodrigoAD, line 273: New type for the `fetchOffchainTokenData` function that abstracts events part in a common type, [ChainEvent](https://github.com/smartcontractkit/ccip-tools-ts/pull/22/files#diff-a165b00264fe1582eba3cb59a6f93c8417e9f2a1cb952d30512d723617da4c19R2), that each chain specific event can adapt to

src/lib/offchain.ts
```diff
@@ -257,3 +268,134 @@ export async function fetchOffchainTokenData(
...
+  sourceChainSelector: bigint
+  ccipLog: Pick<ChainEvent, 'id' | 'index'>
+  txLogs: ChainEvent[]
+  tokenAmounts: TokenAmounts
+}
+
+export async function fetchOffchainTokenDataV2(
```
RodrigoAD, line 280: Replication of `fetchOffchainTokenData` function to validate refactor

src/lib/attestation/client.ts
```diff
@@ -0,0 +1,8 @@
+export type Attestation = {
+  attestation: string
+  messageHash: string
+}
+
+export interface AttestationClient {
```
RodrigoAD, line 6: Small abstraction over the atts. clients, in case more come in

RodrigoAD (commented):
@andrevmatos I would really appreciate an initial look to validate the approach

src/lib/attestation/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './usdc'
```
andrevmatos, line 1: .js suffixes; I'd appreciate a lint rule for those

andrevmatos: Adding this in #24 , will require rebasing and adjusting all imports to use `.ts` extension (going in the direction of modern node support for stripping TS types)

PR #23, opened 2025-03-18
Hasher for Solana

qs:
Hasher for Solana


src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,85 @@
...
+  ): LeafHasher<CCIPVersion.V1_6> =>
+  (message: CCIPMessage<CCIPVersion.V1_6>): string =>
+    hashSolanaMessage(message, hashSolanaMetadata(sourceChainSelector, destChainSelector, onRamp))
+
+const encode = (target: string, value: string | bigint | number) =>
+  defaultAbiCoder.encode([target], [value])
+
+const encodeSolanaAddress = (address: string): Uint8Array => {
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access
```
qs, line 21: `PublicKey` throws an error with missing type, I haven't found an easy way to address it with a type-safe wrapper.

src/lib/selectors.ts
```diff
@@ -400,8 +400,35 @@ export const aptosSelectors: Selectors = {
...
+  },
+  // end:generate
+}
+
 export const isAptosChain = (selector: bigint): boolean => {
   return Object.values(aptosSelectors).some((aptosSelector) => aptosSelector.selector === selector)
 }
 
+export const isSolanaChain = (selector: bigint): boolean => {
+  return Object.values(aptosSelectors).some(
```
RodrigoAD, line 429: Careful here, we could add a test for this fn
```suggestion
  return Object.values(solanaSelectors).some(
```

src/lib/selectors.ts
```diff
@@ -400,8 +401,35 @@ export const aptosSelectors: Selectors = {
...
+  //   .then((obj) => require('util').inspect(obj).split('\n').slice(1, -1))
+  '5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d': {
+    name: 'solana-mainnet',
+    selector: 124615329519749607n,
+  },
+  '4uhcVJyU9pJkvQyS88uRDiswHXSCkY3zQawwpjk2NsNY': {
+    name: 'solana-testnet',
+    selector: 6302590918974934319n,
+  },
+  EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG: {
```
bukata-sa, line 418: missed quotes?

bukata-sa (changes requested):

src/lib/selectors.ts
qs: they were removed by the linter

bukata-sa (approved):

src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,415 @@
...
+import type { LeafHasher } from './common.ts'
+
+// Message types from https://github.com/smartcontractkit/chainlink-ccip/blob/main/chains/solana/contracts/programs/ccip-offramp/src/messages.rs
+/**
+ * RampMessageHeader
+ * - message_id: fixed 32 bytes array.
+ * - source_chain_selector, dest_chain_selector, sequence_number, nonce: u64 values.
+ */
+class RampMessageHeader {
+  message_id: Uint8Array
```
andrevmatos, line 17: Why Uint8Array? We use strings everywhere in this codebase, and Atlas implementation is also siding with keeping bytes32 0x-encoded messageIds (since they are 32 bytes on every supported chain)

src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,415 @@
...
+  source_chain_selector: BN
+  dest_chain_selector: BN
+  sequence_number: BN
+  nonce: BN
+
```
andrevmatos, line 22: We also use `bigint` in this project

src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,415 @@
...
+  constructor(fields: {
+    message_id: Uint8Array
```
andrevmatos, line 24: And camelCase for property names

src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,415 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import { createHash } from 'crypto'
+import { PublicKey } from '@solana/web3.js'
```
andrevmatos, line 4: Please, refrain from pulling such big library only to convert base58

src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,415 @@
...
+import BN from 'bn.js'
```
andrevmatos, line 5: also, bn.js isn't needed, we prefer native bigint

src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,415 @@
...
+    this.compute_units = fields.compute_units
+    this.is_writable_bitmap = fields.is_writable_bitmap
+  }
+}
+
+/**
+ * CrossChainAmount
+ * - le_bytes: fixed 32 bytes array.
+ */
+class CrossChainAmount {
```
andrevmatos, line 57: This will already come as bigint for you from atlas; no need for a class only for this struct?

src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,415 @@
...
+/**
+ * Any2SVMRampMessage
+ * - header: RampMessageHeader
+ * - sender: Vec<u8> as Uint8Array
+ * - data: Vec<u8> as Uint8Array
+ * - token_receiver: Pubkey (32 bytes) as Uint8Array
+ * - token_amounts: array of Any2SVMTokenTransfer
+ * - extra_args: Any2SVMRampExtraArgs
+ */
+class Any2SVMRampMessage {
```
andrevmatos, line 104: This type should be mostly compatible with `CCIPMessage`; please, don't create a class only for it.
Also, the codebase is in functional style, and it's better not to use a class simply to hold some data, you can use a simple `interface` (or actually use the one we already have)

src/lib/hasher/hasher.ts
```diff
@@ -22,6 +23,12 @@ export function getLeafHasher<V extends CCIPVersion = CCIPVersion>({
           destChainSelector,
           onRamp,
         ) as LeafHasher<V>
+      } else if (isSolanaChain(destChainSelector)) {
+        return getV16SolanaLeafHasher(
+          sourceChainSelector,
+          destChainSelector,
+          onRamp,
+        ) as LeafHasher<V>
```
ernest-nowacki, line 31: Suggestion: Might be useful to have a hasher function getter extracted. 

Something like:

```ts
const getV16LeafHasherByChainSelector = (chainSelector: bigint) => {
  if (isAptosChain(chainSelector) {
    return getV16AptosLeafHasher;
  }
  
  if (isSolanaChain(chainSelector)) {
    return getV16SolanaLeafHasher;
  }
  
  return getV16LeafHasher;
} 
```

Than inside the switch statement:

```ts
    case CCIPVersion.V1_6: {
      const v16LeafHasher = getV16LeafHasherByChainSelector(destChainSelector);
      return v16LeafHasher(sourceChainSelector, destChainSelector, onRamp) as LeafHasher<V>
    }
```

src/lib/hasher/solana.test.ts
```diff
@@ -0,0 +1,207 @@
...
+  //     message,
+  //     message.header.sourceChainSelector,
+  //     message.header.destChainSelector,
+  //     message.header.onRamp,
+  //   )
+
+  //   expect(() => hashSolanaMessage(message, metadataHash)).toThrow(
+  //     'invalid DestTokenAddress length',
+  //   )
+  // })
```
ernest-nowacki, line 127: Suggestion: remove for now if it's not used or add description explaining what is needed for this test to be uncommented.

src/lib/hasher/solana.ts
```diff
@@ -0,0 +1,289 @@
...
+  console.log('serializeRampMessage', {
+    buffers,
+    extraArgsBuffer,
+    extraArgs,
+    message,
+    headerBuffer,
+    senderSizeBuffer,
+    dataSizeBuffer,
+    tokenAmountsBuffer,
+  })
```
ernest-nowacki, line 58: Is the log intentional? I saw `console.debug` being used elsewhere in the project maybe that's the preferred convention?

src/lib/hasher/hasher.ts
andrevmatos: This is the getter; this function contains no logic besides selecting the hasher function, so I think it's fine to leave it here, and there's no need for another indirection level

src/lib/hasher/solana.ts
andrevmatos: yep, this should be console.debug; we use all flavors, but fro logs like this, which aren't expected to surface unless user specifies `-v`, it should be debug

PR #34, opened 2025-05-16
Implement solana manual execution + solana library

PabloMansanet:


src/lib/solana/manuallyExecuteSolana.ts
```diff
@@ -0,0 +1,116 @@
...
+    sourceChainSelector: BigInt(ccipRequest.lane.sourceChainSelector),
+    message: ccipRequest.message as CCIPMessage<typeof CCIPVersion.V1_6>,
+    proofs,
+    // Offchain token data is unsupported for manual exec
+    offchainTokenData: ['0x'],
+  })
+
+  const payerAddress = destinationProvider.wallet.publicKey.toBase58()
+
+  const { executionReport, tokenIndexes, accounts, remainingAccounts, addressLookupTableAccounts } =
```
marek-sadura, line 55: From explorer perspective:
It would be great to have logic from line 55 moved to separate function that will take inputs:

```
{
      executionReportRaw,
      connection: destinationProvider.connection,
      offrampProgram,
      root: merkleRoot,
      senderAddress: payerAddress,
    }
    ```
    
    and return VersionedTransaction (just like it does now).
    
    With that I would be able to delete 95% of manual exec code from explorer. I owuld still build exec report using data from atlas and `calculateManualExecProof` on explorer and tx would be build in here

toblich (commented):
Overall looks good. I've left a few small comments, and it also looks like there may be a few lint issues here and there requiring fixing

agusaldasoro:
This PR has been migrated to https://github.com/smartcontractkit/ccip-tools-ts/pull/34

PR #35, opened 2025-05-16
Use buffering contract to mitigate "too large TX" errors when manually executing

PabloMansanet:
Note: due to limitations on the anchor/borsh serialization dependency (see [here](https://github.com/smartcontractkit/ccip-tools-ts/pull/36)) this branch requires some unseemly hacks to work, which really leave it under the standard of quality of the repository.

For that reason, I think this should not be merged, but rather kept as a side branch for oncall responders to fix the large TX issue. That issue is the edge case of an edge case, and should never happen unless our first line of defense (EVM transaction byte limits) as failed. Besides, the entire approach will be reworked with the next offramp update, which defeats the point of investing on this further.

Unless anyone has an argument to the contrary, I'll update the manual exec runbook to reference this branch for the cases where it's necessary.


src/commands/manual-exec.ts
```diff
@@ -107,21 +113,46 @@ export async function manualExec(
...
+      if (info?.confirmationStatus === 'finalized') {
+        return
+      }
+      await new Promise((res) => setTimeout(res, intervalMs))
+    }
+
+    throw new Error(`Transaction ${signature} not finalized after timeout`)
+  }
+
+  var signature
```
toblich, line 138: `var` is old JS syntax and should rarely be used. Please use `let signature: TransactionSignature` here instead

src/commands/manual-exec.ts
```diff
@@ -107,21 +113,46 @@ export async function manualExec(
...
+      const info = status.value[0]
+
+      if (info?.confirmationStatus === 'finalized') {
+        return
+      }
+      await new Promise((res) => setTimeout(res, intervalMs))
+    }
+
+    throw new Error(`Transaction ${signature} not finalized after timeout`)
+  }
```
toblich, line 136: I think it makes sense to move this function out to the main level in this file, so it can be re-used later if desired. It is pretty generic, I think it makes sense to not have it nested inside the `doManuallyExecute`

src/lib/solana/manuallyExecuteSolana.ts
```diff
@@ -97,7 +110,26 @@ export async function buildManualExecutionTxWithSolanaDestination<
...
+      serializedTokenIndexes,
+      accounts,
+      remainingAccounts,
+      computeBudgetIx,
+      blockhash,
+      addressLookupTableAccounts,
+    )
+  } else {
+    return [transaction]
+  }
```
toblich, line 132: Nit: since you're `return`ing inside the `if` branch, there's no need to have an `else` branch. You can just do 

```suggestion
  if (transaction.serialize().length > 1232 || forceBuffer) {
    console.log(
      `Execute report will be pre-buffered through the buffering contract ${bufferProgramAddress}. This may take some time`,
    )
    return bufferedTransactions(
      destinationProvider,
      bufferProgramAddress,
      serializedReport,
      serializedTokenIndexes,
      accounts,
      remainingAccounts,
      computeBudgetIx,
      blockhash,
      addressLookupTableAccounts,
    )
  } 
  
  return [transaction]
```

I haven't checked the linting rules configured in this repo and whether you've run the linter, but [this rule](https://eslint.org/docs/latest/rules/no-else-return) is related to this.

src/lib/solana/manuallyExecuteSolana.ts
```diff
@@ -114,3 +146,116 @@ export function newAnchorProvider(chainName: string) {
...
+  serializedTokenIndexes: Buffer<ArrayBuffer>,
+  originalManualExecAccounts: ManualExecAccounts,
+  originalManualExecRemainingAccounts: AccountMeta[],
+  computeBudgetIx: TransactionInstruction,
+  blockhash: string,
+  addressLookupTableAccounts: AddressLookupTableAccount[],
+): Promise<VersionedTransaction[]> {
+  // Arbitrary as long as there's consistency for all translations.
+  const bufferId = {
+    bytes: Array.from(randomBytes(32)),
```
toblich, line 178: As discussed offline, please either:
- Log the used ID
- Provide a CLI command to close a buffer (in which case you could even just use the message ID as the buffer ID, and just close it if it failed in the middle)

src/lib/solana/manuallyExecuteSolana.ts
```diff
@@ -114,3 +146,116 @@ export function newAnchorProvider(chainName: string) {
...
+  const [bufferAddress] = PublicKey.findProgramAddressSync(
+    [
+      Buffer.from('execution_buffer'),
+      destinationProvider.wallet.publicKey.toBuffer(),
+      Buffer.from(bufferId.bytes),
+    ],
+    new PublicKey(bufferProgramAddress),
+  )
+
+  const chunkSize = 250
```
toblich, line 189: Is there any reason to limit the chunk size to just `250` here? It feels to me like you should be able to fit more than that in each chunk, since the tx shouldn't have such a big overhead here (so something closer to 1K feels like it should be doable)

src/lib/solana/manuallyExecuteSolana.ts
```diff
@@ -114,3 +146,116 @@ export function newAnchorProvider(chainName: string) {
...
+
+  const bufferingProgram = new Program(
+    EXECUTION_BUFFER_IDL,
+    new PublicKey(bufferProgramAddress),
+    destinationProvider,
+  )
+
+  var transactions: VersionedTransaction[] = []
```
toblich, line 197: Please don't use `var`. Given that the array here is just then extended in place without needed to re-assign the variable reference, you should be able to use `const` instead.

PR #37, opened 2025-05-22
Add lookup-table support on Solana manual-exec

toblich:
Adds a `--solana-force-lookup-table` flag that, when set, will:
- Create a lookup table with all the accounts used for the manual exec
- Use that lookup table during the manual exec (whether it is using the intermediary buffer or not)
- Clean-up that lookup table. Note: this takes ~4 mins due to Solana's cooldown period on lookup tables

Other changes:
- Lint fixes, especially on imports
- Refresh the tx recent blockhash on each tx

### DIFF
--- a/src/lib/solana/manuallyExecuteSolana.ts
+++ b/src/lib/solana/manuallyExecuteSolana.ts
@@ -67,6 +67,8 @@ type ManualExecAlt = {
   closeTxs: web3.VersionedTransaction[]
 }
 
+export type ManualExecTxs = { transactions: VersionedTransaction[]; manualExecIdx: number }
+
 export async function buildManualExecutionTxWithSolanaDestination<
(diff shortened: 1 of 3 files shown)


src/lib/solana/manuallyExecuteSolana.ts
```diff
@@ -112,6 +120,75 @@ export async function buildManualExecutionTxWithSolanaDestination<
 
   const { blockhash } = await destinationProvider.connection.getLatestBlockhash()
 
+  console.log('ForceLookupTable', forceLookupTable)
```
PabloMansanet, line 123:
```suggestion
  console.debug('ForceLookupTable', forceLookupTable)
```

PabloMansanet (approved):
All good as far as I'm concerned!

PR #40, opened 2025-06-12
Apply chain updates

andrejrakic:
@zeuslawyer @andrevmatos I am opening this draft PR to add support for calling the `applyChainUpdates` function on the `TokenPool.sol` smart contract, given that some users find arguments for it complex to handle directly from Etherscan or similar tools.

I've added so far:
- JSON template to put arguments there (given the complexity I mentioned above, it's not very user friendly to pass array of Solidity structs directly via CLI)
- function to execute `applyChainUpdates` function
- function to generate calldata for `applyChainUpdates` function only, based on the JSON args. No need for provider or signer objects of any kind, useful for pasting to Gnosis Safe UI for example.
- functionality to detect Solana addresses based on regex and convert them to bytes32 hex values CCIP expects (is this assumption correct @andrevmatos ?)

This PR currently supports only v1.5 pools. I am opening this draft PR also to comment and validate approach I described. If we think this is the most user friendly approach, then I would like to add support for v1.4 pools as well. The core difference is that v1.4 pools have `applyChainUpdates(ChainUpdate[] calldata chains)` function signature with `0xdb6327dc` selector, while v1.5 pools have `applyChainUpdates(uint64[] calldata remoteChainSelectorsToRemove, ChainUpdate[] calldata chainsToAdd)` function signature with `0xe8a1da17` selector. Also `ChainUpdate` struct is different between these two versions.

What I suggest is to have another JSON args template, and based on the content of the JSON args "detect" whether the user wants to interact with v1.4 or v1.5 pool, instead of providing CLI flag (although I can do that as well). Any objections to this?

And do we want ccip-tools-ts to support the `applyRampUpdates` function from v1.2 token pools, which serves a similar purpose? 

Thank u

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,10 +45,10 @@
   },
   "dependencies": {
     "@inquirer/prompts": "7.5.1",
+    "@solana/web3.js": "^1.98.2",
     "@xlabs-xyz/ledger-signer-ethers-v6": "^0.0.1",
     "abitype": "1.0.8",
     "borsh": "^2.0.0",
-    "bs58": "^6.0.0",
     "ethers": "6.14.0",
     "tslib": "2.8.1",
     "yaml": "2.8.0",
--- a/src/commands/apply-chain-updates.ts
+++ b/src/commands/apply-chain-updates.ts
@@ -1,7 +1,7 @@
 import { Contract, Interface } from 'ethers'
 import { readFile } from 'fs/promises'
 import { createInterface } from 'readline/promises'
-import bs58 from 'bs58'
+import { PublicKey } from '@solana/web3.js'
 import TokenPoolABI from '../abi/TokenPool.ts'
 import { bigIntReplacer, chainIdFromName, getProviderNetwork } from '../lib/utils.ts'
 import type { Providers } from '../providers.ts'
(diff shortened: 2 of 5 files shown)


src/abi/TokenPool.ts
```diff
@@ -0,0 +1,1078 @@
+export default [
```
zeuslawyer, line 1: Should we add v number to name for disambiguation? eg `TokenPool_v1_5.ts`. If we need to add compat for v1.4 of applyChainUpdates() it may be clearer by adding v number in the name. WDYT?

src/commands/apply-chain-updates.ts
```diff
@@ -0,0 +1,372 @@
...
+ * If the address is not valid or cannot be decoded to 32 bytes, it returns the original address.
+ *
+ * @param address - A string representing a Solana address in Base58 format.
+ * @returns {string} - Returns the address as a 32-byte hex string prefixed with '0x'.
+ */
+function convertSolanaAddressToBytes32Hex(address: string): string {
+  try {
+    const decoded = bs58.decode(address)
+    if (decoded.length !== 32) {
+      console.warn(
```
zeuslawyer, line 61: Can we do the decoded length check in `isSolanaAddress`?  Ideally all validations should be in a single function that returns a boolean.  The convert function should only handle the conversion. That way the convert function is not returning an invalid address (which it does presently when len != 32).  Also the isSolanaAddress function could throw an error as (IIUC) the program should terminate with an invalid address.

src/commands/apply-chain-updates.ts
```diff
@@ -0,0 +1,372 @@
...
+  remoteChainSelectorsToRemove: bigint[]
+  chainsToAdd: ChainToAdd[]
+}> {
+  // ================================================================
+  // │              Extract Arguments from JSON File                │
+  // ================================================================
+  const argsContent = await readFile(argv.json_args, 'utf-8')
+  const args = JSON.parse(argsContent)
+
+  const remoteChainSelectorsToRemove = Array.isArray(args)
```
zeuslawyer, line 99: At this stage `const remoteChainSelectorsToRemove = Array.isArray(args)` -- wont args always be an object?  the example json file shows it as always an object?

zeuslawyer: right but isnt args in line 97 an object? its reading the whole file (which is the example file you added in this PR   https://github.com/smartcontractkit/ccip-tools-ts/pull/40/files#diff-7d9542aa81f373a120e8c1fea4b347c033a8ca62ff64b319c3f1ebb0db46aaac)

which means `remoteChainSelectorsToRemove` is a property on that object.   Currently the Array.isArray() will always be false when its `args` which is an object?  

I'll leave it to you to and @andrevmatos  -- but my suggestion is to require that the user only provides an object in the file (i.e. the file must never just be a JSON array) so we simple read `remoteChainSelectorsToRemove` and `chainsToAdd` from that object in which case the ternary check on  line 99 wont be needed etc.

package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@inquirer/prompts": "7.5.1",
+    "@solana/web3.js": "^1.98.2",
```
zeuslawyer, line 48: in the main PR description maybe add an "UPDATE" section and call out that this package may be deprecated and there are other packages involved but due to current dependencies we cannot update them all and maybe add that analyses in a new GH Issue and link to it in the Description?

src/commands/apply-chain-updates.ts
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/**
+ * Best-effort Solana address encoder: if the input looks like and is a valid Solana address,
+ * returns a 32-byte 0x-prefixed hex string; otherwise returns the original input.
+ * Only processes addresses that pass the Solana format check to avoid noise.
+ */
+function tryEncodeSolanaAddressToBytes32(address: string): string {
+  // Skip processing if it doesn't look like a Solana address
+  if (!looksLikeSolanaAddress(address)) {
```
zeuslawyer, line 44: shouldnt this logic be run outside of this function before invoking it?  it feels a bit odd that the function is silent if its not a solana address esp since the name is to encode a solana address

src/commands/apply-chain-updates.ts
```diff
@@ -0,0 +1,356 @@
...
+  const SOLANA_ADDRESS_REGEX = /^[1-9A-HJ-NP-Za-km-z]{32,44}$/
+  return SOLANA_ADDRESS_REGEX.test(address)
+}
+
+/**
+ * Best-effort Solana address encoder: if the input looks like and is a valid Solana address,
+ * returns a 32-byte 0x-prefixed hex string; otherwise returns the original input.
+ * Only processes addresses that pass the Solana format check to avoid noise.
+ */
+function tryEncodeSolanaAddressToBytes32(address: string): string {
```
zeuslawyer, line 42: is there a type alias for 0xstring that we should used instead of `string` return type? if not no worries

src/commands/apply-chain-updates.ts
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  console.log(`\nℹ️  Detected potential Solana address, attempting to convert it to bytes32`)
+
+
+  try {
+    const pubkey = new PublicKey(address)
+
+    // Check if the public key is on the ed25519 curve and warn if not
+    if (!PublicKey.isOnCurve(pubkey.toBytes())) {
```
zeuslawyer, line 55: curious - if address is not on the curve would it pass the conversion in line 59-63?  if it would not then maybe we dont need to double assert?

zeuslawyer (approved):
Left some small comments for your consideration in src/commands/apply-chain-updates.ts

PR #43, opened 2025-07-15
Add Aptos specific input formats

RodrigoAD:
## Why
Aptos message events have some different formats

## What
`networkInfo`: 
- Add support for chain selectors coming as strings.
- Unit Tests

`decodeMessage`:
- Add support for decoding messages with snake_case format. It converts every key in the message to camelCase
- Unit tests

Other fixes:
- Failing tests in Solana


andrevmatos (approved):
nice! thank you

PR #47, opened 2025-07-30
[SOLANA][Do not merge] Direct offramp buffering + account autoderivation

PabloMansanet:
Don't merge yet as this requires mainnet support for account autoderivation. It's OK to keep on a side branch for now.

PR #49, opened 2025-08-14
Manual exec from/to Solana with CCTP attestations

toblich:
Integrates the changes from #42 and #47, plus additional code. It adds support to:
- Manually exec on Solana when the message has a CCTP token (e.g. USDC) which requires fetching additional offchain token data such as the attestation. For this, it adds code to:
  - Parse the event from the source chain with the burn data from 1.6 pools
  - Encode the offchain token data with borsh for Solana
  - Derive the accounts for the manual exec by simulating transactions on Solana, instead of implementing the logic for each account on typescript. This is more generic and supports special cases such as CCTP tokens
- Manually exec on EVM when Solana is the source. For it, the new code:
  -	Supports reading non-hex transaction hashes (Solana uses base58 encoding)
  - Adds a new flag `--solana-router` for getting the address of the router in which to look for the event
  - Deserializes the message event and shapes it into something that resembles the evm-specific types that the rest of the code expects _(this is hacky)_
  - Checks for the presence of CCTP-specific events, in which case it fetches the offchain token data too
  - And then executes as normal on EVM chains


src/commands/manual-exec.ts
```diff
@@ -60,29 +70,44 @@ export async function manualExec(
     page: number
     wallet?: string
     solanaOfframp?: string
+    solanaRouter?: string
```
andrevmatos, line 73: Why `solanaRouter`? If this is on source, it's the OnRamp address, isn't it? I'd rather not add new cli parameters if we can auto-discover them

PR #227, opened 2026-04-14
Add missing axios dependency to ccip-sdk

ryannelsonsmartcontract:
## What
- Adds `axios: 1.15.0` as a dependency of `ccip-sdk`

## Why
- `axios` imports are found in `ccip-sdk > src > ton > index.ts`, but the package was not explicitly labeled as a dependency, creating module resolution errors for consumers


aelmanaa (dismissed):

aelmanaa (approved):

andrevmatos (approved):
Ok, I didn't notice we were importing axios directly;
We usually prefer raw `fetch` internally, but it seems TON requires axios to pass some rate-limited client; Should be fine as a minimal dependency surface

PR #231, opened 2026-04-19
Fixes on show, send and manualExec for EVM<->SOL

Farber98:
### Attestation lookup: normalize hash encoding
The attestation API expects hex-encoded hashes, but Solana messages encode the token's extra data as base64. The fix normalizes to hex before calling the API so both formats work transparently.

### EVM→Solana manual execution fixes

- **V0 transaction support for account derivation**: The derivation loop used legacy transactions (via Anchor's `.view()`), which don't support Address Lookup Tables (ALTs). Some TP's requires ALTs, so the legacy tx exceeded Solana's 1232-byte size limit. Switched to V0 transaction simulation which supports ALTs.

- **LBTC attestation encoding** : The offchain token data encoder only handled USDC (Borsh-wrapped). Added an LBTC branch that passes the raw attestation bytes directly.

- **Stale buffer cleanup on retry**: If a previous `manualExec` fails mid-way, it leaves a partially-filled buffer PDA on-chain, blocking retries. The fix detects this and automatically closes the stale buffer before retrying.

### SOL→EVM send fixes

Applied the same V0 simulation fix to the send path: both `getFee()` and account derivation (`deriveAccountsCcipSend`) now use V0 transactions instead of legacy, enabling fee estimation and sends for any token pool that uses ALTs.

### SOL→EVM manual execution fix

Fixed `invalid BytesLike value` error when executing Solana-sourced messages on EVM. Atlas returns an empty string (not `0x`) for the data field on messages with no calldata, which broke the ethers ABI encoder. The fix normalizes empty strings to valid hex bytes.

### Null-safe value transformation in `convertKeysToCamelCase`

The `convertKeysToCamelCase` utility recursively walks API response objects and applies a `mapValues` callback to transform leaf values (e.g. decoding addresses, casting to `BigInt`). When the CCIP API returns `null` for an address field, the null was passed directly to the callback, which called `decodeAddress(null)` → `Buffer.from(null)` → crash.

The fix short-circuits null/undefined values before they reach the callback, preserving them as-is in the output


ccip-sdk/src/solana/index.ts
```diff
@@ -1164,17 +1165,22 @@ export class SolanaChain extends Chain<typeof ChainFamily.Solana> {
...
-        // deactivation before they can be closed/recycled)
-        if (!opts.forceBuffer) opts = { ...opts, forceBuffer: true }
-        else if (!opts.forceLookupTable) opts = { ...opts, forceLookupTable: true }
-        else throw err
+        if (!(err instanceof Error)) throw err
+        if (err.message.includes('AlreadyContainsChunk')) {
+          // stale buffer from a previous failed attempt; close it and retry
+          if (!opts.clearLeftoverAccounts) {
+            opts = { ...opts, clearLeftoverAccounts: true }
+          } else throw err
```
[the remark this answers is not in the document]
Farber98: The flag is wired up in exec.ts, seems it got lost in the call chain

ccip-sdk/src/solana/exec.ts
```diff
@@ -438,19 +441,40 @@ async function autoDeriveExecutionAccounts({
...
+    if (!simResult.returnData?.data[0]) {
+      throw new CCIPError(
+        CCIPErrorCode.SOLANA_SIMULATION_NO_RETURN_DATA,
+        'No return data from deriveAccountsExecute simulation',
+      )
+    }
+    const response: IdlTypes<typeof CCIP_OFFRAMP_IDL>['DeriveAccountsResponse'] =
+      offramp.coder.types.decode(
+        'DeriveAccountsResponse',
+        Buffer.from(simResult.returnData.data[0], 'base64'),
```
andrevmatos: Nit: Maybe `bytesToBuffer` would fit better in this kind of case?

ccip-sdk/src/solana/send.ts
```diff
@@ -145,12 +169,23 @@ export async function getFee(
...
+    instructions: [ix],
+  })
+
+  if (!simResult.returnData?.data[0]) {
+    throw new CCIPSolanaFeeResultInvalidError('No return data from getFee simulation')
   }
 
+  const result: IdlTypes<typeof CCIP_ROUTER_IDL>['GetFeeResult'] = program.coder.types.decode(
+    'GetFeeResult',
+    Buffer.from(simResult.returnData.data[0], 'base64'),
```
andrevmatos: Same

ccip-sdk/src/solana/send.ts
```diff
@@ -159,37 +194,72 @@ async function deriveAccountsCcipSend({
...
+      throw new CCIPError(
+        CCIPErrorCode.SOLANA_SIMULATION_NO_RETURN_DATA,
+        'No return data from deriveAccountsCcipSend simulation',
+      )
+    }
+
+    const response: IdlTypes<typeof CCIP_ROUTER_IDL>['DeriveAccountsResponse'] =
+      router.coder.types.decode(
+        'DeriveAccountsResponse',
+        Buffer.from(simResult.returnData.data[0], 'base64'),
```
andrevmatos: Same

ccip-sdk/src/utils.ts
```diff
@@ -394,6 +394,7 @@ export function isBase64(data: unknown): data is string {
  */
 export function getDataBytes(data: BytesLike | readonly number[]): Uint8Array {
   if (Array.isArray(data)) return new Uint8Array(data)
+  if (typeof data === 'string' && data === '') return new Uint8Array(0)
```
andrevmatos: If you're already doing a strict equality check, no need to also guard on type
```suggestion
  if (data === '') return new Uint8Array(0)
```

ccip-sdk/src/utils.ts
```diff
@@ -482,7 +483,8 @@ export function convertKeysToCamelCase(
     return obj.map((v) => convertKeysToCamelCase(v, mapValues, key))
   }
 
-  if (obj == null || typeof obj !== 'object') return mapValues ? mapValues(obj, key) : obj
+  if (obj == null) return obj
+  if (typeof obj !== 'object') return mapValues ? mapValues(obj, key) : obj
```
andrevmatos, line 487: Why this change? They're equivalent?
Prior to this change, we entered the conditional in either of the branches, but in the former, `mapValues` is always falsy, which goes directly to the `else` branch, which is effectively the same action taken by the first `if`

Farber98: Old clause:
```
if (obj == null || typeof obj !== 'object') return mapValues ? mapValues(obj, key) : obj
```
When `convertKeysToCamelCase({ receiver: null }, mapValues)` recurses into the null value:
- `obj == null → true`, short-circuits the || 
- But then checks `mapValues ? mapValues(null, key) : null`
- `mapValues` is truthy → calls `mapValues(null, "receiver")`→ crashes

New code returns null directly without ever calling `mapValues`

See this [TS Playground](https://www.typescriptlang.org/play/?ssl=21&ssc=36&pln=21&pc=45#code/PTAEHkBkBFQYwPYFsBGBLAdgUwCagOYCuAhgE44BcoGhANrQVgC4DOoADsSy7qEwqCTF2ANWK1CWFgCgAZoQxwmaBBlAJaOAMKqAbllJMAFAhQArKgoDWGBAHcMAGkHCxEqQH4qR3ZYw37J1ArL1AWJlJMfABKUABeAD5Qa1sHZyssAE9Q8MiMGL8Ah1AAb2lQUDRZUCMAQVJSYkyAOjQWesbMk3No2NJmQlI1UzNmoXYfeKSNbT0DY11ncbdJFnSs3vLK6u6zeLjqOgYAH2O+TPYsBGqR0ABCOIOAchGsJSe+gaGXUXFV0A8PxWUl260ysSoIy2IAqsLh8IRiKRyJRCIAehjMVjsRj4OJaLw7GgmAALQ70O5bRAYcLwOaGXBUABKbwQ5AAPLkos4UoEkgcSgBfLayNk1am0gDaVmcugAuupquBzG8mM0sBgImgQSNenSMPoGThpQqDjMdAb5j4lq4-lJ0tEtv0mIM1NTDUxcNJhdIYQA5ACiAHUwlhOI1PeSGEQyJQo6AUBcuDw2Ms7WwNcp+rRMnIFEoVGpsHYLR7doVUkE0+4WKEfBXAukclr8rFEsl-JWwc28gUO0U1GUKlUah0mq12g0mrs9c7XepzGNhJN28XS1bFkD02DNsOdrdHlHPi7vrdkTDAAmE8bnQ14ODQ-SUOZtvxr1Cwhrx9C9e5qTAuVw3OY9yHi8KrvMe87Vv8gLQTq5g7qAkLmFSqi0u68yMqALKIByXL5DynZ8vEpTChUoqkOKaFMKA0qygq1wQOBaqZpE8FmHqGFGiaJFrvSCwvsCazBI6FQ3m6-Fej6MIsGgSB0MQkZ2CSimgP0ACOqysM0rCgDgCBSFQcFsHA+JsDgrIWbUOA4P03DAAAQmg+AAJKauoai6Om0gSjRETEDSFFICR9b9pWbZJEO2w1Lo+xHnwJKkPY752KAAYNGyRgAAYAKrYAAHpcSi8DQ9AcMmvD8HwjSBWySB3FlolqV8nnetI0gRJkpRbOa-FGCUzVwFgaCGlQpUMIKzj+bVpBII6gp4kwcBkkYWCxFFEoaFgzS0Ag+BGE8UCwHAjQsCSFBPM4q2gFwaUZaQ0RjFILDEPga3QmA7YAER5VghWqiVRzldwlUCNNLBBXcX1tT5aFbTte0HYGIb6VI1AIDRJ1cOdl2gAAUgAyuAfrNPh+BVF0fGWoY-WDcNo3xpN1UBRDdW9I6MLtiUX2PvTBhfRQ42CkAA) example

Farber98: CI was failing [here](https://github.com/smartcontractkit/ccip-tools-ts/actions/runs/24678530260/job/72169169415?pr=231): 

```
test at src/evm/fork.test.ts:168:5
✖ should return native and token balances for CCIP transfer participants (210.837879ms)
  TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received null
      at Buffer.from (node:buffer:348:9)
      at decodeBase64 (file:///home/runner/work/ccip-tools-ts/ccip-tools-ts/node_modules/ethers/lib.esm/utils/base64.js:27:32)
      at getAddressBytes (file:///home/runner/work/ccip-tools-ts/ccip-tools-ts/ccip-sdk/src/utils.ts:459:25)
      at EVMChain.getAddress (file:///home/runner/work/ccip-tools-ts/ccip-tools-ts/ccip-sdk/src/evm/index.ts:569:18)
      at decodeAddress (file:///home/runner/work/ccip-tools-ts/ccip-tools-ts/ccip-sdk/src/utils.ts:307:16)
      at file:///home/runner/work/ccip-tools-ts/ccip-tools-ts/ccip-sdk/src/requests.ts:92:11
      at convertKeysToCamelCase (file:///home/runner/work/ccip-tools-ts/ccip-tools-ts/ccip-sdk/src/utils.ts:486:66)
      at convertKeysToCamelCase (file:///home/runner/work/ccip-tools-ts/ccip-tools-ts/ccip-sdk/src/utils.ts:493:27)
      at file:///home/runner/work/ccip-tools-ts/ccip-tools-ts/ccip-sdk/src/utils.ts:483:27
      at Array.map (<anonymous>) {
    code: 'ERR_INVALID_ARG_TYPE'
  }
```

andrevmatos (approved):

PR #236, opened 2026-04-24
align gas estimate with ccip v2

aelmanaa:
This pull request refines the way gas limits are calculated for EVM chain transactions in the `EVMChain` class. The new approach provides a more accurate and robust estimate by accounting for various worst-case gas consumption scenarios that may occur during execution, rather than relying on a fixed overhead. This helps ensure transactions have sufficient gas under adversarial but valid conditions.

**Gas estimation improvements:**

* Enhanced the calculation of `execTx.gasLimit` in `EVMChain` to include worst-case gas usage from ERC165 interface probes, additional buffer for state updates, and EIP-150 forwarding overhead, based on the presence of token transfers and receiver payloads.

### DIFF
--- a/ccip-sdk/src/evm/index.ts
+++ b/ccip-sdk/src/evm/index.ts
@@ -18,7 +18,6 @@ import {
   isBytesLike,
   isError,
   isHexString,
-  keccak256,
   toBeHex,
   toBigInt,
 } from 'ethers'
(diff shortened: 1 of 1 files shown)


andrevmatos (commented):
I wonder if we could/should replace the `executeSingleMessage` with just `execute` estimation.

Rationale: we used `executeSingleMessage` back in the day, because it was the internal method which always reverted upon failure (required for accurate estimation), while the outer `execute` always try/catched.

After pointing this to the on-chain team, they seem to have implemented the CCIPv1 behavior: if it's the first execution and it fails, catch the error and store the failure, otherwise bubble up the error (actually [NoStateProgressMade](https://github.com/smartcontractkit/chainlink-ccip/blob/60cdf88568fc63be41e4ad17960c19081381eca0/chains/evm/contracts/offRamp/OffRamp.sol#L252), but that also works).

Now, this should work for non-first execution, but we're still left with an algorithm which doesn't work on first (as due to the try/catch, it'll under-estimate). Maybe we could lowerbound execTx.gasLimit to `message.executionGasLimit`, which is the gasLimit always used by exec plugin.

aelmanaa:
> I wonder if we could/should replace the `executeSingleMessage` with just `execute` estimation.
> 
> Rationale: we used `executeSingleMessage` back in the day, because it was the internal method which always reverted upon failure (required for accurate estimation), while the outer `execute` always try/catched.
> 
> After pointing this to the on-chain team, they seem to have implemented the CCIPv1 behavior: if it's the first execution and it fails, catch the error and store the failure, otherwise bubble up the error (actually [NoStateProgressMade](https://github.com/smartcontractkit/chainlink-ccip/blob/60cdf88568fc63be41e4ad17960c19081381eca0/chains/evm/contracts/offRamp/OffRamp.sol#L252), but that also works).
> 
> Now, this should work for non-first execution, but we're still left with an algorithm which doesn't work on first (as due to the try/catch, it'll under-estimate). Maybe we could lowerbound execTx.gasLimit to `message.executionGasLimit`, which is the gasLimit always used by exec plugin.

@andrevmatos check now please

andrevmatos (dismissed):

andrevmatos (approved):

PR #243, opened 2026-04-30
feat(cli): add safe-propose command for Safe multisig CCIP proposals

SyedAsadKazmi:
Adds `ccip-cli safe-propose` — proposes a CCIP send to the Safe Transaction Service queue without on-chain execution, so other Safe owners can review and sign in the Safe UI.

- Proposes ERC-20 approvals as separate transactions with sequential nonces ahead of `ccipSend`
- Signs with EIP-712 via `signTransaction()`
- Warns if Safe balance is below estimated fee at proposal time
- Outputs Safe UI queue URL and safeTxHashes; `--format json` supported
- `parseExtraArgs` moved to `utils.ts`; `RPCS_RE` exported from `providers/index.ts`
- Adds `safe-propose.mdx` reference page, sidebar entry, and commands table row

### Dependencies added (`ccip-cli`)

| Package | Version |
|---|---|
| `@safe-global/protocol-kit` | `^7.1.0` |
| `@safe-global/api-kit` | `^4.1.0` |


ccip-cli/src/commands/safe-propose.ts
andrevmatos, line 1: I think this deserves to be an option for tx-sending commands, instead of its own command;
so probably a `--safe <safeAddress>` option side-by-side with wherever `--wallet` is defined/called.

In practice, you still need the current/existing `--wallet` logic, to load the account/signer with which you'll effectively sign the proposal and send the on-chain tx to Safe's inbox.

But then, the tx-sending logic can/should be "wrapped" by safe logic, caching its nonce locally and, instead of sending the tx to e.g. Router, it'd wrap the unsigned tx and send it to Safe instead as a proposal.

Everything should work mostly as is, with the only considerations being, when you read back the result, we may need to adjust the expectation that the tx contains a CCIP request (e.g. in the end of the `EVMChain.sendMessage` method), and also that simulation/estimateGas will always succeed or provide accurate gasLimit (due to ccipSend being simulated strictly after approvals are mined). But I think it should be feasible to work around or special those cases when `--safe` is provided

ccip-cli/src/commands/safe-propose.ts
```diff
@@ -0,0 +1,571 @@
...
+}) => SafeApiKitInstance
+
+export const command = 'safe-propose'
+export const describe = 'Propose a CCIP send as a Safe multisig transaction'
+
+/**
+ * Extracts the first available RPC URL from CLI args, RPC_* env vars, or rpcsFile.
+ * Needed to initialise the Safe Protocol Kit, which requires a single concrete URL.
+ */
+function extractFirstRpcUrl(argv: { rpcs?: string[]; rpcsFile?: string }): string {
```
andrevmatos, line 108: Please, don't. Prefer to use `source.provider` directly if needed. Does safe libs have some form to integrate directly with ethers' `provider`s? Would be even better than assuming a RPC url.

ccip-cli/src/commands/safe-propose.ts
```diff
@@ -0,0 +1,571 @@
...
+  const SafeApiKitSDK = apiKitModule.default as SafeApiKitConstructor
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+  const getEip3770Prefix = safeModule.getEip3770NetworkPrefixFromChainId as (
+    chainId: bigint,
+  ) => string
+
+  // 12. Initialise the Safe Protocol Kit
+  const protocolKit = await SafeSDK.init({
+    provider: rpcUrl,
+    signer: proposerPrivateKey,
```
andrevmatos, line 445: Shouldn't either assume a private key; wallet could be a Ledger wallet.
Instead, let's make sure this works and simply wraps the discovered signer.

ccip-cli/src/commands/safe-propose.ts
```diff
@@ -0,0 +1,571 @@
...
+
+  // 10. Get a concrete RPC URL for the Safe Protocol Kit
+  const rpcUrl = extractFirstRpcUrl(argv)
+
+  // 11. Lazy-load the Safe SDK.
+  // Dynamic imports with `as any` casts are necessary to work around a TypeScript NodeNext
+  // CJS interop issue: @safe-global packages have no "type":"module", causing TS to
+  // resolve the module namespace as the default export type, making it circular.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment
+  const safeModule = (await import('@safe-global/protocol-kit')) as any
```
andrevmatos, line 430: Please, let's also avoid those mixed imports; there's usually some way to import them in nodejs, even if it requires importing or accessing explicit `default`s properties

ccip-cli/src/commands/send.ts
```diff
@@ -178,47 +177,6 @@ export async function handler(argv: Awaited<ReturnType<typeof builder>['argv']>
...
 
-/**
- * Parse --extra=key=value entries into a record.
- * Values are parsed as JSON with bigint support for integers; fallback to string.
- * Repeated keys are accumulated into an array.
- *
- * @example
- * `assert.eq(parseExtraArgs(['ccvs=["0xvalue1", "0xvalue2"]', 'finality=safe', 'flag=true']), { ccvs: ["0xvalue1", "0xvalue2"], finality: 'safe', flag: true })`
- */
-function parseExtraArgs(extra: readonly string[] | undefined): Record<string, unknown> {
```
andrevmatos, line 189: Shouldn't need this move if safe is built as a wrapper for wallet instead

andrevmatos (changes requested):

PR #244, opened 2026-05-04
Fix TON > Solana manual exec - sender address encoding

krebernisak:
Example issue when running manual exec for a TON > Solana message: https://ccip-ui-chore-bump-ccip-sdk-1-5-0-chainlinklabs.vercel.app/api/h/ccipData/buildManuallyExecuteTx?destChainSelectorName=solana-devnet&messageId=0xcb612421de1c4aa30fa9502b7e74bdc402b8aa93404232e3f4c61f86e57038bf&senderAddress=DqsjtWF3S45F4rNfo61MQ9dwBw95qcP9GKTGbGuHUaQZ&gasLimitOverride=10000

```
Unsupported data format: "0:358280f2b46935d7470439a34fd234cc8617f2019018545383a74b03b9035174"
```

This PR introduces a fix where now `getAddressBytes` is used to decode `message.sender` during manual execution encoding vs. previously we used `bytesToBuffer` which threw the above error.


ccip-sdk/src/ton/types.ts
```diff
@@ -2,9 +2,12 @@ import { type Builder, Address, Cell, beginCell } from '@ton/core'
...
 
 import { CCIPDataFormatUnsupportedError } from '../errors/specialized.ts'
-import type { CCIPMessage_V1_6_EVM } from '../evm/messages.ts'
+import type { EVMExtraArgsV2 } from '../extra-args.ts'
 import type { CCIPMessage_V1_6, ChainFamily, ExecutionInput } from '../types.ts'
-import { bytesToBuffer } from '../utils.ts'
+import { bytesToBuffer, getAddressBytes } from '../utils.ts'
+
+/** TON-specific CCIP v1.6 message type. TON uses the same extra-args layout as EVM. */
+export type CCIPMessage_V1_6_TON = CCIPMessage_V1_6 & EVMExtraArgsV2
```
andrevmatos: I don't think you need the new type;
If the type is effectively the same as EVM, I think it's fine to just reuse it;
At most, we could make a type alias (which would NOT yield a different type than just reusing the EVM type), but that wouldn't be useful either, as anywhere accepting one, would accept the other.
Let's keep using `_EVM` for TON, until we have any real differentiation in it.

ccip-sdk/src/types.ts
```diff
@@ -158,7 +160,7 @@ export interface Lane<V extends CCIPVersion = CCIPVersion> {
 export type CCIPMessage<V extends CCIPVersion = CCIPVersion> = V extends typeof CCIPVersion.V2_0
   ? CCIPMessage_V2_0
   : V extends typeof CCIPVersion.V1_6
-    ? CCIPMessage_V1_6_EVM | CCIPMessage_V1_6_Solana | CCIPMessage_V1_6_Sui
+    ? CCIPMessage_V1_6_EVM | CCIPMessage_V1_6_Solana | CCIPMessage_V1_6_Sui | CCIPMessage_V1_6_TON
```
andrevmatos: No need, no value, pls

andrevmatos (changes requested):

andrevmatos (approved):

PR #250, opened 2026-05-06
fix: manualExec v1.5 messages from API (messageId) from cli

andrevmatos:
- API v1.5 messages lack some required v1.2..v1.5 fields in main `/messages` endpoint: `strict` and `sourceTokenData`
- It errors when trying to `manualExec` one of those messages from CLI while passing a `messageId`, as it tries to specify a complete `input` to `execute`, but with API's message payload, which lacks those fields, and errors when trying to `manuallyExecute.populateTransaction`
- So, in this case, also fetch `/execution-inputs`, which contain raw data and those required fields
- These bugs shouldn't affect codepaths used by Explorer or other clients, which'll usually execute passing `messageId` to execute, not a whole `input` fetched from API

### DIFF
--- a/ccip-sdk/src/chain.ts
+++ b/ccip-sdk/src/chain.ts
@@ -700,8 +700,7 @@ export abstract class Chain<F extends ChainFamily = ChainFamily> {
       ])
       if (trans.timestamp <= finalizedTs) return true
     }
-    const signals = [this.abort, abort].filter(Boolean) as AbortSignal[]
-    const watch = signals.length ? AbortSignal.any(signals) : true
+    const watch = abort ? AbortSignal.any([this.abort, abort]) : this.abort
     for await (const l of this.getLogs({
       address: log.address,
       startBlock: log.blockNumber,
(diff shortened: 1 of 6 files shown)


ccip-sdk/src/chain.ts
```diff
@@ -1183,11 +1183,23 @@ export abstract class Chain<F extends ChainFamily = ChainFamily> {
     opts: ExecuteOpts,
   ): Promise<Extract<ExecuteOpts, { input: unknown }>> {
     let opts_: Extract<typeof opts, { input: unknown }>
-    if ('input' in opts) {
+    if (
+      'input' in opts &&
+      ('encodedMessage' in opts.input || // v2
+        'feeValueJuels' in opts.input.message || // v1.6
+        'strict' in opts.input.message) // <=v1.5 with all required fields (API /messages lack some)
```
andrevmatos, line 1189: If this field is lacking for a v1.5 message (i.e. message coming from API), skip this branch to fetch `/execution-inputs` from API, which comes with this field.

aelmanaa (approved):

PR #257, opened 2026-05-15
Add Canton commands

RodrigoAD:
- Add support for Canton in `manual-exec` and `send` commands. Adds a new flag `canton-config` to point to a Canton config file
- Add support for external signing


ccip-cli/src/commands/manual-exec.ts
```diff
@@ -118,6 +120,11 @@ export const builder = (yargs: Argv) =>
         string: true,
         example: '--receiver-object-ids 0xabc... 0xdef...',
       },
+      'canton-config': {
+        type: 'string',
+        describe:
+          'Path to Canton config JSON file (party, ccipParty, jwt, edsUrl, transferInstructionUrl, etc.)',
```
RodrigoAD: @andrevmatos Adding all the necessary values to get Canton working from a config file instead of adding 4-5 flags

ccip-cli/src/commands/manual-exec.ts
```diff
@@ -204,59 +211,73 @@ async function manualExec(
...
+                2,
+              ),
+            )
+          }
+          return
+        }
+      } catch (err) {
+        if (err instanceof CCIPNotImplementedError) {
+          logger.warn('Gas estimation not supported for this destination chain; skipping')
+          if (argv.onlyEstimate) return
```
andrevmatos: This `if` estimate branch is entered only if the user explicitly asked for `--estimate-gas-limit`;
We shouldnn't just log and return, or even worse, log and move on, if this errors; let it blow up.
If the user entered this path, they deserve to see any error that this throws; if they want it to not error, they should explicitly omit the option in the next run instead.
Please, undo this try/catch

ccip-cli/src/commands/send.ts
```diff
@@ -228,7 +233,9 @@ async function sendMessage(
   const destNetwork = networkInfo(argv.dest)
   const getChain = fetchChainsFromRpcs(ctx, argv)
   const source = await getChain(sourceNetwork.name)
-  decodeAddress(argv.router, sourceNetwork.family)
+  if (sourceNetwork.family !== ChainFamily.Canton) {
+    decodeAddress(argv.router, sourceNetwork.family)
```
andrevmatos: `decodeAddress` calls `CantonChain.getAddress`; any chance of handling this there, instead of special-casing it here?

ccip-cli/src/commands/send.ts
```diff
@@ -350,6 +357,65 @@ async function sendMessage(
...
+          party: string,
+        ): Promise<{ feeToken: string; extraArgs: Record<string, unknown> }>
+      }
+    ).discoverSendArgs(walletAddress)
+
+    // Merge: discovered defaults < code-generated extraArgs (which already includes --extra overrides)
+    const cantonExtraArgs = {
+      ...discovered.extraArgs,
+      ...extraArgs,
+    }
```
andrevmatos: Oh, no. Please, don't.
All the parameters collected during the "non-canton" flow below (which should actually be the only one) are passed directly to `CantonChain.sendMessage`. Any custom-Canton-specific "discovery" you need to do, you should do there instead, so we don't need this big branch, and from `cli`ents perspective, behavior is the same, which is the whole point of the SDK.

ccip-cli/src/commands/send.ts
```diff
@@ -350,6 +357,65 @@ async function sendMessage(
...
+
+    if (argv.onlyGetFee) {
+      logger.warn('Fee estimation is not supported for Canton; skipping --only-get-fee')
+      return
+    }
```
andrevmatos: You can just return some fixed value in `getFee`, e.g. `0`; it'll just print and skip on `--only-get-fee`, no need to special-case it

ccip-cli/src/commands/send.ts
```diff
@@ -350,6 +357,65 @@ async function sendMessage(
     ...parseExtraArgs(argv.extra),
   }
 
+  // ─── Canton-specific send flow ──────────────────────────────────────────────
+  if (sourceNetwork.family === ChainFamily.Canton) {
```
andrevmatos: You can achieve the same behavior of this branch, while adjusting code only on your CantonChain class.
I agree the config may be inevitable, but let's do the MINIMUM special-casing for specific chain families in the CLI

ccip-cli/src/providers/index.ts
```diff
@@ -55,13 +57,42 @@ async function collectEndpoints(
   return endpoints
 }
 
+/**
+ * Load and validate a Canton config JSON file.
+ * @param configPath - Path to JSON file, or undefined if not provided.
+ * @param logger - Logger for debug output.
+ * @returns Parsed CantonConfig or undefined.
+ */
+function loadCantonConfig(
```
andrevmatos: Please, move this to your `canton.ts` file

ccip-cli/src/providers/index.ts
```diff
@@ -246,6 +279,29 @@ export async function loadChainWallet(
...
+      const privateKey = typeof argv.wallet === 'string' ? argv.wallet : undefined
+      if (privateKey && /^(0x)?[0-9a-fA-F]{64}$/.test(privateKey)) {
+        const signer = new Ed25519TransactionSigner(privateKey, party)
+        logger?.debug(
+          `Canton wallet: external signer created (fingerprint=${signer.getFingerprint()})`,
+        )
+        return [party, { party, signer }] as const
+      }
+
+      return [party, { party }] as const
```
andrevmatos: Please, move all of this to some function in your canton.ts file; the parameters here in the index.ts should be the minimal which requires knowledge this case is dealing with a canton source; everything els ehere should be put in there.

ccip-cli/src/providers/index.ts
```diff
@@ -55,13 +57,42 @@ async function collectEndpoints(
...
+  }
+
+  logger?.debug('Loaded Canton config from', configPath, 'for party', parsed['party'])
+  return parsed as unknown as CantonConfig
+}
+
 export function fetchChainsFromRpcs(
   ctx: Ctx,
-  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api'>,
+  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api'> & { cantonConfig?: string },
```
andrevmatos: `cantonConfig` may deserve to become a `GlobalOpts`, since it's shared between more than 1 command

ccip-sdk/src/canton/index.ts
```diff
@@ -1750,6 +1829,16 @@ function stripHexPrefix(hex: string): string {
   return hex.startsWith('0x') ? hex.slice(2) : hex
 }
 
+/** Decode a base64-encoded string into a Uint8Array. */
+function base64ToUint8Array(b64: string): Uint8Array {
```
andrevmatos: Use `getDataBytes`

ccip-sdk/src/canton/index.ts
andrevmatos: We have a `sleep` utility somewhere, which should do the same

andrevmatos (changes requested):
cli code should be as agnostic of family specificities as possible;
let's adjust it to contain the Canton-specific stuff on the existing codepaths

ccip-cli/src/commands/manual-exec.ts
andrevmatos: Good call, thank you for this.

ccip-cli/src/providers/index.ts
andrevmatos:
```suggestion
  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api' | 'cantonConfig'>,
```

ccip-cli/src/providers/index.ts
```diff
@@ -57,11 +58,11 @@ async function collectEndpoints(
 
 export function fetchChainsFromRpcs(
   ctx: Ctx,
-  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api'>,
+  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api'> & { cantonConfig?: string },
 ): ChainGetter
 export function fetchChainsFromRpcs(
   ctx: Ctx,
-  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api'>,
+  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api'> & { cantonConfig?: string },
```
andrevmatos: Same

ccip-cli/src/commands/manual-exec.ts
```diff
@@ -204,7 +205,12 @@ async function manualExec(
 
   let inputs
   if (source) {
-    offRamp ??= await discoverOffRamp(source, dest, request.lane.onRamp, source)
+    // Canton destinations don't use offRamp/onRamp discovery
+    if (dest.network.family === ChainFamily.Canton) {
+      offRamp ??= ''
+    } else {
+      offRamp ??= await discoverOffRamp(source, dest, request.lane.onRamp, source)
```
andrevmatos: Hmm, I'd rather have this special case inside `discoverOffRamp`; there, you can just check, if dest family is Canton, just return `''` before the discover loop

ccip-cli/src/providers/index.ts
```diff
@@ -76,9 +77,10 @@ export function fetchChainsFromRpcs(
  */
 export function fetchChainsFromRpcs(
   ctx: Ctx,
-  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api'>,
+  argv: Pick<GlobalOpts, 'rpcs' | 'rpcsFile' | 'api'> & { cantonConfig?: string },
```
andrevmatos: .

andrevmatos (approved):
Good job! The PR ended up as clean as spring water! LGTM

PR #268, opened 2026-05-28
fix(ton): prevent polling loop from exiting due to unref'd AbortSignal.timeout timers

nicolasgnr:
AbortSignal.timeout() uses unref'd timers internally in Node.js, which do not keep the event loop alive. This caused the TON transaction polling loop to exit prematurely while awaiting the next poll interval.

This change replaces timeout-based polling delays with a standard setTimeout() wrapped in Promise.race(), preserving AbortSignal cancellation semantics while ensuring the polling loop correctly keeps the process alive until completion or cancellation.`

### DIFF
--- a/ccip-cli/src/index.ts
+++ b/ccip-cli/src/index.ts
@@ -28,7 +28,7 @@ Error.stackTraceLimit = 50 // show more stack frames for better debugging
 
 // generate:nofail
 // `const VERSION = '${require('./package.json').version}-${require('child_process').execSync('git rev-parse --short HEAD').toString().trim()}'`
-const VERSION = '1.7.1-90c0438'
+const VERSION = '1.7.1-2807dab'
 // generate:end
 
 const require = createRequire(import.meta.url)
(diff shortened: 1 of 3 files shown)


andrevmatos (changes requested):
Unfortunately, I don't think we can accept this PR.

The behavior of AbortSignal not holding the event loop is intentional. It prevents those cascade of signals from leaking memory, and dangling contexts potentially forever (which is very bad in long-lived scripts), which is not the case with `Promise.race` (search for _Promise.race leaks_).

Instead, you should explicitly keep your event loop for the duration of your main task.
This is how we do it in cli:
https://github.com/smartcontractkit/ccip-tools-ts/blob/2807dab7854032374d5f754469a51853cbd117f7/ccip-cli/src/index.ts#L161-L169

nicolasgnr:
Thanks for the explanation and feedback @andrevmatos. It's not be best user experience as a consumer of the SDK, however, I understand the concern and the proposal / workaround.

PR #273, opened 2026-06-12
feat(cct-sdk): Init CCT SDK with setPool (EVM+Solana)

apedrob:
## What
- [DAPP-10175](https://smartcontract-it.atlassian.net/browse/DAPP-10175): Add the CCT SDK at `@chainlink/ccip-sdk/cct`
- Introduce `EVMTokenManager` with `generateUnsignedSetPool` (build unsigned tx) and `setPool` (sign + submit)
- Add `setPool` operation module: validates params, encodes `setPool(localToken, pool)`, discovers `TokenAdminRegistry` via router
- Add `Operation` / `EVMOperation` lifecycle abstraction (validate → encode → submit)
- Add shared CCT infra: `TokenManager` base, submit action, address validations, and error types 

## Why
- TMEM's Set Pool flow currently builds calldata client-side via `ccip-contracts`. This adds a first-class SDK path for CCT admin ops so consumers (partners / dapp team) can build and submit txs server-side with consistent validation and error handling

# Testing
- CI

## Notes

- https://github.com/smartcontractkit/explorer/pull/10928 vendors this SDK and wires it into our server-side endpoint
- [CCT SDK Architecture & Requirements (DD)](https://docs.google.com/document/d/1-4Q-l5bc_kX5olWH3JWOT2Sl5oqaESUFlum2LQNxs4Y/edit?usp=sharing)

[DAPP-10175]: https://smartcontract-it.atlassian.net/browse/DAPP-10175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/evm/operations/set-pool.ts
```diff
@@ -0,0 +1,84 @@
...
+/** Result of `setPool`. */
+export type SetPoolResult = CctTxResult
+
+/**
+ * Validates {@link SetPoolParams} before any RPC.
+ * @throws {@link CCIPCctParamsInvalidError} if any address is invalid
+ */
+function validate(params: SetPoolParams): void {
+  validateAddress(OPERATION, 'tokenAddress', params.tokenAddress)
+  validateAddress(OPERATION, 'poolAddress', params.poolAddress)
```
apedrob: Should we go deeper and validate whether the address is a token or pool contract through `typeAndVersion`?

ccip-sdk/src/cct/evm/operations/set-pool.ts
apedrob, line 1: Thinking of a generic abstract class, e.g. `Operation` that defines function signatures for `validate`, `generate` and `execute`. What are your thoughts?

ccip-sdk/src/cct/token-manager.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/** Result of any single-transaction CCT write. */
+export interface CctTxResult {
+  txHash: string
+}
+
+/** Base for a chain-family CCT manager; subclasses hold the concrete `chain`. */
+export abstract class TokenManager<F extends ChainFamily = ChainFamily> {
+  abstract readonly chain: Chain<F>
+}
```
apedrob, line 18: Should we add the function signatures for the chain-agnostic operations, i.e. `generateUnsignedSetPool` and `setPool`?

ccip-sdk/package.json
```diff
@@ -27,6 +27,10 @@
       "types": "./dist/all-chains.d.ts",
       "default": "./dist/all-chains.js"
     },
+    "./cct/evm": {
```
apedrob, line 30: Exposed under the subpath `/evm` for now; we probably want to expose just `/cct` with a barrel export

ccip-sdk/src/cct/evm/operations/set-pool.ts
apedrob: Added!

ccip-sdk/src/cct/token-manager.ts
mervin-link: I’d avoid it for now. The params are different per chain, and this keeps us from ending up with a huge union of chain-specific params. It’ll be easier to maintain and extend.

ccip-sdk/src/cct/evm/submit.ts
```diff
@@ -0,0 +1,77 @@
...
+      isTransient: isTransientError(error),
+    })
+  }
+
+  chain.logger.debug(`${operation}: waiting for confirmation, tx =`, response.hash)
+
+  let receipt
+  try {
+    receipt = await response.wait(1, CONFIRM_TIMEOUT_MS)
+  } catch (error) {
```
aelmanaa, line 70: **(blocker — correctness).** In ethers v6, `response.wait()` **throws `CALL_EXCEPTION`** on a mined revert (`status === 0`) — it does not resolve a status-0 receipt. This `catch` maps every throw to `CCTTxNotConfirmedError` (`isTransient: true, retryAfterMs: 5000`), so a permanent revert (e.g. `onlyAdministrator`) is reported as retryable and retry-on-transient logic will resubmit a reverting tx repeatedly. Consequently the `if (receipt.status === 0)` branch below (line 69) is dead code.

Fix: discriminate in the catch — `if (isError(error, 'CALL_EXCEPTION'))` throw a permanent revert error (reuse `CCIPExecTxRevertedError`, keep `cause`); only `TIMEOUT`/`TRANSACTION_REPLACED` → `CCTTxNotConfirmedError`. Delete the dead `status === 0` branch. And add a unit test whose fake `wait()` **throws** — `submit.test.ts` currently resolves `wait` to `{ status: 0 }` (v5-style), which is why this slipped through.

ccip-sdk/src/cct/evm/submit.ts
```diff
@@ -0,0 +1,77 @@
...
+  operation: string,
+): Promise<TransactionHash> {
+  if (!isSigner(wallet)) throw new CCIPWalletInvalidError(wallet)
+
+  chain.logger.debug(`${operation}: submitting...`)
+
+  let response: TransactionResponse
+  try {
+    let tx: TransactionRequest = { ...unsigned.transactions[0]! }
+    tx = await wallet.populateTransaction(tx)
```
aelmanaa, line 54: **(major — correctness).** This relies on `wallet.populateTransaction` deriving the nonce from the pending count and never touches the SDK's nonce cache. `sendMessage` (`evm/index.ts:1603/1618`) and `execute` (`:1830`) set `tx.nonce = await this.nextNonce(sender)` — an optimistic per-address cache that allows back-to-back sends. Single standalone `setPool` is fine, but batch/interleaved admin ops on a shared `EVMChain` (or mixing with `sendMessage`/`execute`, which bump `this.nonces`) will collide.

Fix: set `tx.nonce = await chain.nextNonce(sender)` (with rollback on failure — note core `execute` currently omits the rollback that `sendMessage` does, so standardize on the `sendMessage` model). Ideally via a shared `submitAndConfirm` helper.

ccip-sdk/src/cct/errors.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @packageDocumentation
+ */
+
+import { type CCIPErrorOptions, CCIPError, CCIPErrorCode } from '../errors/index.ts'
+
+// Parameter validation
+
+/** Thrown before any RPC when operation params fail validation. Permanent. */
+export class CCTParamsInvalidError extends CCIPError {
```
aelmanaa, line 26: **(blocker — DX).** These three CCT error classes are re-exported by neither `errors/index.ts` nor `cct/evm/index.ts`, so a consumer can't `import { CCTTxFailedError }` to `instanceof`-narrow, and the `@throws {@link ...}` JSDoc links resolve to unimportable symbols. Same for `SetPoolParams`/`TransactionHash` in the public signatures. Every other CCIP error is public — this is the sole exception.

Fix: re-export the classes + types; ideally fold this file into `errors/index.ts` so `instanceof` works from the main barrel. Do this after the error-reuse cleanup (see the comment below) so you only export the surface that survives — the reused `CCIPExecTx*` errors are already public.

ccip-sdk/src/cct/evm/token-admin/operations/set-pool.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { UnsignedEVMTx } from '../../../../evm/types.ts'
+import { ChainFamily } from '../../../../networks.ts'
+import { EVMOperation } from '../../operation.ts'
+import { validateAddress } from '../../validate.ts'
+
+/** Parameters for `setPool`. Zero `poolAddress` delists the token. */
+export interface SetPoolParams {
+  tokenAddress: string
+  poolAddress: string
+  routerAddress: string
```
aelmanaa: **(major — DX).** `getTokenAdminRegistryFor(address, ...)` is deliberately polymorphic — it accepts Router / OnRamp / OffRamp / TokenPool / TokenAdminRegistry (`evm/index.ts:1135-1149`) and is named `address` on every family for that reason. Narrowing to `routerAddress` forces a caller holding the chain-global registry address to mislabel it, and the router path costs extra hops (router → getOnRamp(dest) → staticConfig) that fail without a configured lane.

Fix: name it `address` (matching the SDK) since actually one can either use the router or the token admin registry to call setPool

ccip-sdk/src/cct/evm/token-admin/operations/set-pool.ts
```diff
@@ -0,0 +1,45 @@
...
+  protected validate(p: SetPoolParams): void {
+    validateAddress(this.name, 'tokenAddress', p.tokenAddress)
+    validateAddress(this.name, 'poolAddress', p.poolAddress)
+    validateAddress(this.name, 'routerAddress', p.routerAddress)
+  }
+
+  /** Encodes `setPool` on the TokenAdminRegistry discovered from the router. */
+  protected async encode(chain: EVMChain, p: SetPoolParams): Promise<UnsignedEVMTx> {
+    const to = await chain.getTokenAdminRegistryFor(p.routerAddress)
+    const data = new Interface(TokenAdminRegistryABI).encodeFunctionData('setPool', [
```
aelmanaa: **(DRY).** This builds `new Interface(TokenAdminRegistryABI)` on every `encode()`. The SDK keeps a built-once instance: `interfaces.TokenAdminRegistry` (`evm/const.ts:51`, used at `evm/index.ts:1863/1891`).

Fix: `import { interfaces }` and call `interfaces.TokenAdminRegistry.encodeFunctionData('setPool', [p.tokenAddress, p.poolAddress])`.

Two small extras nearby: add a one-line comment that TAR `setPool` encoding is version-stable 1.5→2.0, so nobody copies this no-dispatch pattern into a version-sensitive op; and the zero-`poolAddress`-delists behavior is documented only on the internal params interface — surface it on the public method JSDoc.

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  /** Run {@link validate} and {@link encode}, applying optional `sender`; no signing. */
+  async generate(chain: EVMChain, params: P): Promise<UnsignedEVMTx> {
+    this.validate(params)
+    const unsigned = await this.encode(chain, params)
+    if (params.sender && unsigned.transactions[0]) unsigned.transactions[0].from = params.sender
+    return unsigned
+  }
+
+  /** {@link generate}, then sign and submit via {@link submit}; returns once confirmed. */
```
aelmanaa, line 34: **(nit — consistency).** The core signed paths override the caller's sender with the wallet before generating: `sendMessage` (`evm/index.ts:1589`) and `execute` (`:1820`) both do `sender = await wallet.getAddress()`, so `from` always matches the signer. Here `params` is passed straight through, so a user `sender !== wallet` trips ethers' from/signer mismatch in `populate()` (surfaced as an opaque failure), and `sender === wallet` is a silent no-op. `sender` only meaningfully applies to `generateUnsignedSetPool`.

Fix: in `execute`, `const sender = await wallet.getAddress()` and pass `{ ...params, sender }` to `generate`. Same root cause as the nonce-cache issue — the durable fix is a shared `submitAndConfirm` helper.

ccip-sdk/src/cct/evm/validate.ts
```diff
@@ -0,0 +1,23 @@
...
+import { isAddress } from 'ethers'
+
+import { CCTParamsInvalidError } from '../errors.ts'
+
+/**
+ * Asserts `value` is a valid EVM address.
+ * @throws {@link CCTParamsInvalidError} if it is not
+ */
+export function validateAddress(operation: string, param: string, value: unknown): void {
+  if (typeof value !== 'string' || !isAddress(value)) {
```
aelmanaa: **(DRY / consistency).** This is the only place in the SDK validating an address with bare `isAddress`. The canonical family-aware primitive is validate-by-convert via `getAddress` (throws `CCIPAddressInvalidError`; used explicitly for validation, e.g. `chain.ts:2161`). Delegate to `getAddress` + reuse `CCIPAddressInvalidError`/`ADDRESS_INVALID`, keeping a thin `CCTParamsInvalidError` wrapper only if you want the `{ operation, param }` context.

ccip-sdk/src/cct/evm/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Registers a pool, signing + submitting with `opts.wallet` (the token admin).
+   * @throws {@link CCIPWalletInvalidError} if `wallet` is not a valid signer
+   * @throws {@link CCTParamsInvalidError} if any param is invalid
+   * @throws {@link CCTTxFailedError} if the tx reverts or fails
+   */
+  setPool(opts: SetPoolParams & { wallet: unknown }): Promise<TransactionHash> {
+    return this.#setPool.execute(this.chain, opts)
+  }
+
```
aelmanaa: **(blocker — CI).** Stray blank line before the closing `}` — prettier flags this file (and `index.test.ts:132`), which is why `build-and-test` is red. Run `npm run lint:fix`.

pnpm-lock.yaml
```diff
@@ -0,0 +1,1891 @@
+lockfileVersion: '9.0'
```
aelmanaa: **(blocker).** Wrong lockfile for this repo — it uses npm (`package-lock.json`). This `pnpm-lock.yaml` was added in `1c36578` and no new dependency requires it (CCT uses `ethers`, already present). It drifts silently and can mislead CI/agents into `pnpm install`.

Fix: `git rm pnpm-lock.yaml` and add it to `.gitignore`.

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -0,0 +1,36 @@
...
+import { submit } from './submit.ts'
+
+/** EVM CCT write base. Subclasses supply {@link validate} and {@link encode}. */
+export abstract class EVMOperation<P extends { sender?: string }> extends Operation<
+  EVMChain,
+  P,
+  UnsignedEVMTx
+> {
+  /** Encode calldata into an unsigned tx; versioned ops resolve their encoder here. */
+  protected abstract encode(chain: EVMChain, params: P): Promise<UnsignedEVMTx> | UnsignedEVMTx
```
mervin-link: `encode` is too EVM-specific function name. I would propose `build` or `buildUnsigned`.

ccip-sdk/src/cct/evm/submit.ts
apedrob: Yep, agreed; didn't consider batched transactions here. I'll add the `nextNouce` together with a rollback on failure

ccip-sdk/src/cct/evm/token-admin/operations/set-pool.ts
apedrob: Makes sense, renamed to `address` (if it gets too generic we can edit to e.g. contractAddress after)

pnpm-lock.yaml
apedrob: Removed it and added it to `.gitignore`; committed the pnpm-lock out of habit from working in the explorer repo

ccip-sdk/src/cct/evm/validate.ts
apedrob: Although `EVMChain.getAddress` is static, importing it will drag the entire `EVMChain` class for a leaf validator. I will add `CCIPAddressInvalidError` as a cause to this throw.

aelmanaa: Agreed. Chaining `CCIPAddressInvalidError` as the cause keeps the canonical error without pulling the whole `EVMChain` in for a leaf check. Resolved.

ccip-sdk/src/cct/evm/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Registers a pool, signing + submitting with `opts.wallet` (the token admin).
+   * A zero/empty `poolAddress` delists the token from the registry.
+   * @throws {@link CCIPWalletInvalidError} if `wallet` is not a valid signer
+   * @throws {@link CCTParamsInvalidError} if any param is invalid
+   * @throws {@link CCTTxFailedError} if the tx reverts or fails
+   */
+  setPool(opts: SetPoolParams & { wallet: unknown }): Promise<TransactionHash> {
```
aelmanaa, line 89: Missing `@example` on `setPool` / `generateUnsignedSetPool` and on the CCT error classes. Every error in `specialized.ts` has one, and TypeDoc pulls these into the generated `ccip-api-ref` SDK reference. A short snippet covering the must-be-admin precondition and the zero `poolAddress` delist would cover it. Only open item on my side.

aelmanaa:
CCT will add many more write ops beyond `setPool`. The populate/nonce/submit/wait/classify flow is already repeated in `sendMessage`, `execute`, and this `submit`, each handling reverts a bit differently. Might be worth one shared `submitAndConfirm` utility so all write paths (and the future CCT ops) behave the same. Not blocking this PR.

ccip-sdk/src/cct/evm/index.ts
apedrob: Good to know, added!

aelmanaa (dismissed):

mervin-link (dismissed):

mervin-link (approved):

PR #274, opened 2026-06-18
logs streaming improvements

andrevmatos:
- getLogs: improve and fix several cases of `block range too wide`-like RPC errors and range detection
- ton: fetch InProgress logs, return masterchain `seqno` as `blockNumber` and `lt` as `log.index`
- allow bigints in public interfaces accepting numeric types (to allow easier [de]serialization)
- solana: optimize streaming `ExecutionStateChanged` from offramps (skip commit-only txs)
- cli: ignore commented out RPCs in `--rpcs-file`


ccip-sdk/src/ton/index.ts
Farber98: should we re-key cache here by and use seq no instead of LT?

Farber98: Also, should we update [this](https://github.com/smartcontractkit/ccip-tools-ts/blob/87ce705f89c114fa0d6c138d1c0ea7f9545ac708/ccip-cli/src/commands/show.test.ts#L573) assertion?

andrevmatos: Good call; I'm actually thinking on getting rid of this whole `getTransactions` cache, or at least simplifying it, since we can now properly find timestamps for blocks (using `getBlockHeader` call)

ccip-sdk/src/ton/index.ts
```diff
@@ -375,7 +450,7 @@ export class TONChain extends Chain<typeof ChainFamily.TON> {
         blockNumber: res.blockNumber, // Note: This is lt (logical time), not block seqno
```
Farber98: Comment seems stale

ccip-sdk/src/ton/index.ts
Farber98: comment seems stale

ccip-sdk/src/ton/index.ts
```diff
@@ -251,6 +269,66 @@ export class TONChain extends Chain<typeof ChainFamily.TON> {
...
+      )
+    }
+    const { result } = (await res.json()) as {
+      result: Awaited<ReturnType<TONChain['getMCBlockHeader']>>
+    }
+    return result
+  }
+
   /**
    * Fetch the timestamp for a given logical time (lt) or finalized block.
```
Farber98: comment seems stale

ccip-sdk/src/ton/index.ts
Farber98: comment seems stale

Farber98: comment seems stale

Farber98 (approved):

PR #278, opened 2026-06-25
fix(canton): prod testnet send and manual-exec for ccip-cli and ccip-sdk

SyedAsadKazmi:
Fix Canton prod testnet CCIP flows (Sepolia ↔ Canton) in ccip-tools-ts. Canton support already exists; this patch set addresses gaps that blocked prod testnet lanes.

### SDK fixes

- Correct ACS template package names for prod testnet (`ccip-runtime`, etc.)
- Fix CCIPReceiver resolution and `requiredCCVs` matching on execute
- Fix CCV override handling for token-pool manual exec
- Fix Canton send/execute details (fee token config, gas limit encoding, ledger scoping)
- Honor `chainId` in canton-config when synchronizer alias is generic (`global`) so prod testnet resolves to `canton:TestNet` instead of defaulting to `canton:DevNet`

### CLI fixes

- Resolve `-r` / `--indexer` / `ccvs` from canton-config where appropriate
- Filter mixed `--rpc` lists so EVM endpoints aren’t raced against Canton ledger URLs
- Pass message receiver hash to Canton `manual-exec` by default
- Avoid redundant source tx re-fetch in `show` after `send`

### Known limitations / workarounds

- **`--only-get-fee` on Canton source** — not useful today; Canton has no scalar upfront fee (`getFee` returns `0`). Real cost is computed in the on-chain send command. 
     - [ ] **[TODO: implement Canton fee quoting or document EDS/registry-based estimate]**
      
- **Gas estimation on Canton source** — `estimateReceiveExecution` calls `typeAndVersion` on the source OnRamp/router; **`CantonChain.typeAndVersion` is not implemented**, so positive `--estimate-gas-limit` fails for Canton→EVM sends. **
     - [ ] **[TODO: implement Canton OnRamp `typeAndVersion` or a Canton-specific gas estimator]**

- **Workaround:** pass `--estimate-gas-limit -101` (skips estimation when `<= -100`) and set gas explicitly, e.g. `-x gasLimit=200000`

### Tested
- Sepolia → Canton: data + token send and manual exec
- Canton → Sepolia: data + token send (with `--estimate-gas-limit -101` + `-x gasLimit=…`)

### DIFF
--- a/ccip-cli/src/commands/send.ts
+++ b/ccip-cli/src/commands/send.ts
@@ -41,8 +41,7 @@ import type { GlobalOpts } from '../index.ts'
 import { showRequests } from './show.ts'
 import { type Ctx, Format } from './types.ts'
 import { getCtx, logParsedError, parseTokenAmounts } from './utils.ts'
-import { loadCantonConfig, resolveCliRouter } from '../providers/canton.ts'
-import { fetchChainsFromRpcs, loadChainWallet } from '../providers/index.ts'
+import { fetchChainsFromRpcs, loadChainWallet, resolveRouter } from '../providers/index.ts'
 
 export const command = 'send'
 export const describe = 'Send a CCIP message from source to destination chain'
@@ -228,12 +227,7 @@ async function sendMessage(
   const { output, logger } = ctx
   const sourceNetwork = networkInfo(argv.source)
   const destNetwork = networkInfo(argv.dest)
-  const cantonConfig = loadCantonConfig(argv.cantonConfig, logger)
-  const router = resolveCliRouter(
-    argv.router,
-    cantonConfig,
-    sourceNetwork.family === ChainFamily.Canton,
-  )
+  const router = resolveRouter(argv, sourceNetwork, logger)
   if (!router) {
     throw new CCIPArgumentInvalidError(
       'router',
--- a/ccip-cli/src/commands/send.ts
+++ b/ccip-cli/src/commands/send.ts
@@ -41,6 +41,7 @@ import type { GlobalOpts } from '../index.ts'
 import { showRequests } from './show.ts'
 import { type Ctx, Format } from './types.ts'
 import { getCtx, logParsedError, parseTokenAmounts } from './utils.ts'
+import { loadCantonConfig, resolveCliRouter } from '../providers/canton.ts'
 import { fetchChainsFromRpcs, loadChainWallet } from '../providers/index.ts'
 
 export const command = 'send'
@@ -68,8 +69,8 @@ export const builder = (yargs: Argv) =>
     .option('router', {
       alias: 'r',
       type: 'string',
-      demandOption: true,
-      describe: 'Router contract address on source',
+      describe:
+        'Router contract address on EVM source, or CCIPSender instance id on Canton source (defaults to canton-config senderInstanceId)',
     })
     .options({
       receiver: {
@@ -227,9 +228,24 @@ async function sendMessage(
   const { output, logger } = ctx
   const sourceNetwork = networkInfo(argv.source)
   const destNetwork = networkInfo(argv.dest)
+  const cantonConfig = loadCantonConfig(argv.cantonConfig, logger)
+  const router = resolveCliRouter(
+    argv.router,
+    cantonConfig,
+    sourceNetwork.family === ChainFamily.Canton,
+  )
+  if (!router) {
+    throw new CCIPArgumentInvalidError(
+      'router',
+      sourceNetwork.family === ChainFamily.Canton
+        ? 'required on Canton source: pass -r or set senderInstanceId in canton-config'
+        : 'required: pass -r with the source router contract address',
+    )
+  }
+
   const getChain = fetchChainsFromRpcs(ctx, argv)
   const source = await getChain(sourceNetwork.name)
-  decodeAddress(argv.router, sourceNetwork.family)
+  decodeAddress(router, sourceNetwork.family)
 
   let data: BytesLike | undefined
   if (argv.data) {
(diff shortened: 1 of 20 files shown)


ccip-cli/src/commands/send.ts
```diff
@@ -227,9 +228,24 @@ async function sendMessage(
   const { output, logger } = ctx
   const sourceNetwork = networkInfo(argv.source)
   const destNetwork = networkInfo(argv.dest)
+  const cantonConfig = loadCantonConfig(argv.cantonConfig, logger)
+  const router = resolveCliRouter(
```
RodrigoAD: It's probably best to not have chain specific logic inside the CLI commands. This affects to other `resolve...` functions.

Something like this, that follows other patterns like `fetchChainsFromRpcs` might be better
```
// providers/index.ts
function resolveRouter(argv, sourceNetwork) {
   const cantonConfig = loadCantonConfig(argv.cantonConfig)
   // ...
```
@andrevmatos can you give us your thoughts here?

ccip-sdk/src/canton/explicit-disclosures/acs.ts
```diff
@@ -37,6 +40,55 @@ function extractInstanceId(createArgument: unknown): string | null {
   return extractStringField(createArgument, 'instanceId')
 }
 
+/** Length of a list field on createArgument (empty list → 0). */
+function extractListLength(createArgument: unknown, fieldName: string): number {
+  return extractStringListField(createArgument, fieldName).length
+}
+
+/** Extract RawInstanceAddress.unpack strings from a list field on createArgument. */
+function extractStringListField(createArgument: unknown, fieldName: string): string[] {
```
RodrigoAD: This function is quite hard to understand, what's its purpose? Can we simplify it?

ccip-sdk/src/canton/ccv-addresses.ts
```diff
@@ -0,0 +1,92 @@
+import { id as keccak256Utf8 } from 'ethers'
+
+/** Normalize a hex CCV InstanceAddress for comparison (lowercase, no 0x). */
+export function normalizeCcvHex(value: string): string {
```
RodrigoAD: I've seen this in `acs.ts` too. Let's avoid the duplication. Also, I think there's some general utils in the repo for this

ccip-sdk/src/canton/index.ts
```diff
@@ -187,10 +210,35 @@ export class CantonChain extends Chain<typeof ChainFamily.Canton> {
     tokenMetadataClient: TokenMetadataClient,
     ccipParty: string,
     indexerUrl = MAINNET_INDEXER_URLS[0]!,
+    ledgerParty?: string,
```
RodrigoAD: We should probably make this required, defaulting to `ccipParty` doesn't make too much sense for users

ccip-sdk/src/utils.ts
```diff
@@ -371,6 +375,11 @@ export function getAddressBytes(address: BytesLike | readonly number[]): Uint8Ar
   return bytes
 }
 
+/** Daml party ID: `hint::1220<64-hex-fingerprint>` (not a 3-part instrument id). */
+function isCantonPartyId(address: string): boolean {
```
RodrigoAD: Same as other utilities, I've seen them duplicated in different places.

RodrigoAD (commented):
Thanks @SyedAsadKazmi for the fixes! Left you some comments. Although we should get a review from @andrevmatos

ccip-cli/src/commands/send.ts
SyedAsadKazmi: Moved router/indexer resolution into `providers/index.ts` as `resolveRouter` / `resolveIndexer`, following the same pattern as `fetchChainsFromRpcs` (commands pass argv + network/chain context; canton-config loading stays in the providers layer). Low-level helpers remain in `canton.ts` for unit tests.

ccip-sdk/src/canton/explicit-disclosures/acs.ts
SyedAsadKazmi: Split the monolithic `extractStringListField` into smaller helpers:

- `readCreateArgumentField` - flat object or `{ fields: [{ label, value }] }`
- `asLedgerList` - bare array vs `{ list }` / `{ elements }`
- `extractRawInstanceAddressUnpack` - one `RawInstanceAddress` entry -> `unpack` hex
- `extractRequiredCCVs` - composes the above for CCIPReceiver `requiredCCVs`

Removed the separate `requiredCCVsLength` helper; callers use `extractRequiredCCVs(...).length` when needed.

ccip-sdk/src/canton/ccv-addresses.ts
SyedAsadKazmi: Consolidated the duplicated helpers into `utils.ts`:

- Exported `normalizeHex`, `hashedUtf8Hex`, and `isCantonPartyId`
- `acs.ts` and `ccv-addresses.ts` now import these instead of local copies (removed the duplicate `isCantonPartyId` / `normalizeHex` in `acs.ts` and `normalizeCcvHex` in `ccv-addresses.ts`)
- `getAddressBytes` uses `hashedUtf8Hex` for Canton party hashing as well

ccip-sdk/src/utils.ts
SyedAsadKazmi: Resolved with [this](https://github.com/smartcontractkit/ccip-tools-ts/pull/278#discussion_r3480964112).

ccip-sdk/src/canton/index.ts
SyedAsadKazmi: Made `ledgerParty` required on the constructor and `fromClient`, removed the `?? ccipParty` fallback, and updated the `CantonConfig` docs to clarify `party` (user ledger / actAs) vs `ccipParty` (CCIP operator). `fromUrl` now validates that `ctx.cantonConfig.party` is set and passes it through explicitly.

RodrigoAD (approved):
@SyedAsadKazmi LGTM, let's wait for @andrevmatos 👍

andrevmatos (approved):

PR #282, opened 2026-07-02
fix(canton): send/execute reliability, config defaults, and update-ID…

SyedAsadKazmi:
Canton CCIP support fixes and SDK exports so Canton→EVM flows work end-to-end without adding Canton-specific logic to generic CLI commands.

### Fixes
- **Manual exec:** Accept Canton ledger update IDs; fetch transactions via `update-by-id`; decode `CCIPMessageSent` from ledger events (selectors, OffRamp, `0x`-prefixed `encodedMessage`)
- **Canton send:** Auto-provision router/sender disclosures; support `executorMode=none` for manual EVM execution; config-driven default gas limit
- **Canton execute:** CCV address decoding/matching; update receiver `requiredCCVs` and return new receiver contract ID; improved ACS/EDS disclosure handling
- **EVM → Canton:** Apply no-execution executor default on the EVM send path only when `extraArgs.executor` is unset
- **Config:** `CantonConfig` + `Chain.cantonConfig` for send gas, fee transfer amount, and executor defaults (overridable via `canton-config.json` / `extraArgs`)

### SDK exports
- `canton/defaults.ts` resolvers and `applyCantonDestExecutorDefault`
- `parseCantonInstrumentId`, `parseInstrumentId`, `isCantonWallet`
- `readMessageV1ChainSelectors`, `readMessageV1OffRampAddress`
- `isCantonUpdateId` / `normalizeCantonUpdateId` (via `isSupportedTxHash`)

### CLI
- `manual-exec` validation/help only — handler stays chain-agnostic (`fetchChainsFromRpcs` → `getMessagesInTx` → `execute`)

### Tests
- New/updated unit tests for defaults, update IDs, CCV addresses, types, and message decoding helpers


ccip-sdk/src/canton/ccv-addresses.ts
```diff
@@ -1,5 +1,49 @@
...
+ * The indexer returns UTF-8 bytes as a hex string; Go uses `string(vr.VerifierDestAddress)`.
+ */
+export function decodeCantonVerifierDestAddress(destAddress: string): string {
+  const trimmed = destAddress.trim()
+  if (trimmed.includes('@')) return trimmed
+
+  const hex = trimmed.startsWith('0x') ? trimmed.slice(2) : trimmed
+  if (/^[0-9a-fA-F]+$/.test(hex) && hex.length >= 2 && hex.length % 2 === 0) {
+    try {
+      const decoded = Buffer.from(hex, 'hex').toString('utf8')
```
RodrigoAD: `Buffer` is NodeJS specific, this will break in browser envs. I'm sure we have cross-env supported utilities for this

ccip-sdk/src/canton/defaults.ts
```diff
@@ -0,0 +1,99 @@
...
+/** Default gas for Canton → destination sends when `extraArgs.gasLimit` is omitted. */
+export const DEFAULT_CANTON_SEND_GAS_LIMIT = 50_000n
+
+/** Transfer-factory preview amount for fee payments. */
+export const DEFAULT_CANTON_FEE_TRANSFER_FACTORY_AMOUNT = '1.0'
+
+/** CCIPSender `instanceId` when creating a missing sender contract. */
+export const DEFAULT_CANTON_SENDER_INSTANCE_ID = 'ccipsender'
+
+/** @deprecated Use {@link DEFAULT_CANTON_NO_EXECUTION_EXECUTOR} */
```
RodrigoAD: I'd say it's safe to rename this without a deprecation strategy. I don't think anyone is using this currently

ccip-sdk/src/canton/update-id.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/** Full canonical update ID length: prefix (4) + digest (64) = 68 hex chars. */
+export const CANTON_UPDATE_ID_HEX_LENGTH = 68
+
+/**
+ * Returns true when `id` looks like a Canton ledger update ID (not a CCIP message ID).
+ *
+ * Message IDs use `0x` + 64 hex. Update IDs use `1220` + 64 hex (optionally with `0x`).
+ */
+export function isCantonUpdateId(id: string): boolean {
```
RodrigoAD, line 27: We probably have utilities for this already. Let's try to avoid duplication

ccip-sdk/src/chain.ts
```diff
@@ -722,6 +742,7 @@ export abstract class Chain<F extends ChainFamily = ChainFamily> {
       )
     this.network = network as NetworkInfo<F>
     this.logger = logger
+    this.cantonConfig = cantonConfig
```
RodrigoAD: Why do we need this on the base class?

SyedAsadKazmi: The config is Canton-related, but the consumer is whichever chain is the **source** when dest is Canton, today that's `EVMChain`, tomorrow potentially others. Keeping `cantonConfig` on `ChainContext` / `Chain` is a generic passthrough so future non-EVM -> Canton sources can apply the same dest defaults without re-plumbing CLI context. 

Happy to scope to `EVMChain` only if we prefer YAGNI until a second source family lands.

ccip-sdk/src/canton/update-id.ts
SyedAsadKazmi: Refactored to use `normalizeHex` + ethers `isHexString` instead of custom regex/hex stripping. No pre-existing `isCantonUpdateId` utility, the duplication was in the hex parsing/validation primitives. Canton-specific update-ID semantics remain in `canton/update-id.ts`.

ccip-sdk/src/canton/defaults.ts
SyedAsadKazmi: Removed the deprecated lines.

ccip-sdk/src/canton/ccv-addresses.ts
SyedAsadKazmi: Replaced Node-specific `Buffer` decoding with cross-env helpers.

ccip-sdk/src/utils.ts
```diff
@@ -259,6 +260,7 @@ export function decodeAddress(address: BytesLike, family: ChainFamily = ChainFam
  * @throws {@link CCIPChainFamilyUnsupportedError} if specified chain family is not supported
  */
 export function isSupportedTxHash(txHash: unknown, family?: ChainFamily): txHash is string {
+  if (typeof txHash === 'string' && isCantonUpdateId(txHash)) return true
```
andrevmatos: This should be part of `Canton.isTxHash`, look at the code below, not a separate, "special-case `if`" in the top of this file

ccip-sdk/src/messages.ts
```diff
@@ -277,3 +277,34 @@ export function decodeMessageV1(encodedMessage: BytesLike): MessageV1 {
...
+  if (encoded.length < 71) throw new CCIPMessageDecodeError('MESSAGE_MIN_SIZE')
+  let offset = 69
+  const onRampAddressLength = encoded[offset++]!
+  offset += onRampAddressLength
+  if (offset >= encoded.length) throw new CCIPMessageDecodeError('MESSAGE_OFFRAMP_ADDRESS_LENGTH')
+  const offRampAddressLength = encoded[offset++]!
+  if (offset + offRampAddressLength > encoded.length) {
+    throw new CCIPMessageDecodeError('MESSAGE_OFFRAMP_ADDRESS_CONTENT')
+  }
+  return hexlify(dataSlice(encoded, offset, offset + offRampAddressLength))
```
andrevmatos: You shouldn't `hexlify` this, you cannot be sure the format of the address; I'd suggest also reading `destChainSelector` (or maybe using above's function), and then using `decodeAddress` with dest's family instead of hexlify; even if it's EVM, hexlify will return lowercase, which messes with our standard for EVM of preferring ethers' default checksummed address foramt

ccip-cli/src/commands/send.ts
```diff
@@ -198,12 +198,16 @@ function parseExtraArgs(extra: readonly string[] | undefined): Record<string, un
...
-    let value: unknown = entry.substring(eqIdx + 1)
+    const raw = entry.substring(eqIdx + 1)
+    let value: unknown = raw
 
-    try {
-      value = jsonParse<unknown>(value as string)
-    } catch {
-      // pass
+    // Bare 0x literals are hex (addresses, calldata); yaml/json parse would coerce them to bigint.
+    if (!/^0x[0-9a-fA-F]+$/i.test(raw)) {
```
andrevmatos: `0x` may also be useful
```suggestion
    if (!/^0x[0-9a-fA-F]*$/i.test(raw)) {
```

ccip-sdk/src/extra-args.ts
```diff
@@ -16,6 +16,21 @@ export const SVMExtraArgsV1Tag = id('CCIP SVMExtraArgsV1').substring(0, 10) as '
...
+export const NO_EXECUTION_TAG = '0xeba517d2' as const
+
+const noExecutionAddressBytes = new Uint8Array(20)
+noExecutionAddressBytes.set([0xeb, 0xa5, 0x17, 0xd2], 0)
+
+/**
+ * EVM executor sentinel for GenericExtraArgsV3: skip automated destination execution.
+ * Maps to `Client.NO_EXECUTION_ADDRESS` in chainlink-ccip EVM contracts.
+ */
+export const NO_EXECUTION_ADDRESS = getAddress(hexlify(noExecutionAddressBytes))
```
andrevmatos: Nit, for extra expressivity as code, and for consistency with similar uses elsewhere:
```suggestion
export const NO_EXECUTION_TAG = dataSlice(id('NO_EXECUTION_TAG'), 0, 4) as '0xeba517d2'

/**
 * EVM executor sentinel for GenericExtraArgsV3: skip automated destination execution.
 * Maps to `Client.NO_EXECUTION_ADDRESS` in chainlink-ccip EVM contracts.
 */
export const NO_EXECUTION_ADDRESS = getAddress(zeroPadBytes(NO_EXECUTION_TAG, 20))
```

andrevmatos (commented):
Almost there; nice additions!!

andrevmatos (approved):

PR #289, opened 2026-07-07
feat(cct-sdk): Add version dispatch + Transfer Ownership

apedrob:
## What
- [DAPP-10418](https://smartcontract-it.atlassian.net/browse/DAPP-10418): Add version-aware dispatch for CCT EVM pool operations
- Add `transferOwnership` operation on EVMTokenManager with `generateUnsignedTransferOwnership` (build unsigned tx) and `transferOwnership` (sign + submit) to showcase the dispatch logic
- Add CCT version-dispatch error types: `CCTContractTypeInvalidError`, `CCTTokenPoolVersionUnsupportedError`, `CCTOperationUnsupportedError`

## Why
- CCT pool contracts span multiple versions (1.5.0, 1.5.1, 1.6.1, 2.0.0) with divergent encodings. 
- This adds a reusable dispatch so each op registers encoders only where its calldata changes, resolving the right ABI and encoder at runtime, keeping core `CCIPVersion` logic untouched by isolating the new `TokenPoolVersion` within cct/

## Testing
- CI

## Notes
- [CCT SDK Architecture & Requirements (DD)](https://docs.google.com/document/d/1-4Q-l5bc_kX5olWH3JWOT2Sl5oqaESUFlum2LQNxs4Y/edit?tab=t.0) updated with version dispatch, error handling and op lifecycle sections

[DAPP-10418]: https://smartcontract-it.atlassian.net/browse/DAPP-10418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/evm/token-pool/operations/transfer-ownership.ts
```diff
@@ -0,0 +1,58 @@
...
+export interface TransferOwnershipParams {
+  poolAddress: string
+  newOwner: string
+  sender?: string
+}
+
+/** Encodes `transferOwnership` calldata against the resolved pool ABI. */
+type Encoder = (abi: InterfaceAbi, params: TransferOwnershipParams) => UnsignedEVMTx
+
+const encodeTransferOwnership: Encoder = (abi, { newOwner, poolAddress }) => {
```
mervin-link, line 27: What is the plan for multiple versions with a long implementations? Should we cram everything in here or create another file for each implementations.

ccip-sdk/src/cct/errors.ts
```diff
@@ -65,3 +65,62 @@ export class CCTTxNotConfirmedError extends CCIPError {
...
+        ...options,
+        isTransient: false,
+        context: { ...options?.context, address, expected, actual },
+      },
+    )
+  }
+}
+
+/** Thrown when a token pool reports a version string the SDK does not recognize. Permanent. */
+export class CCTTokenPoolVersionUnsupportedError extends CCIPError {
```
mervin-link: Should we generalize this to `CCTVersionUnsupportedError`? Since we do have a message containing specific version and contract type (`Unsupported token pool version: ${version}`).

apedrob: I can make it as `CCTContractVersionUnsupportedError`, similarly to the error above

ccip-sdk/src/cct/evm/token-pool/operations/transfer-ownership.ts
apedrob: For now we can keep it within the file; if it grows beyond, we can move it around.

ccip-sdk/src/cct/evm/index.ts
```diff
@@ -89,6 +94,28 @@ export class EVMTokenManager extends TokenManager<typeof ChainFamily.EVM> {
...
     return this.#setPool.execute(this.chain, opts)
   }
+
+  /**
+   * Builds an unsigned pool `transferOwnership` tx (for multisig / offline signing).
+   * @throws {@link CCTParamsInvalidError} if any param is invalid
+   * @throws {@link CCTContractTypeInvalidError} if the pool is not a recognised pool type
+   * @throws {@link CCTContractVersionUnsupportedError} if the pool version is unsupported
+   */
+  generateUnsignedTransferOwnership(opts: TransferOwnershipParams): Promise<UnsignedEVMTx> {
```
mervin-link, line 113: Missing `@example` for both `generateUnsignedTransferOwnership` and `transferOwnership`.

ccip-sdk/src/cct/evm/token-pool/version.ts
```diff
@@ -0,0 +1,125 @@
...
+import TokenPool_2_0 from '../../../evm/abi/TokenPool_2_0.ts'
+import type { EVMChain } from '../../../evm/index.ts'
+import {
+  CCTContractTypeInvalidError,
+  CCTContractVersionUnsupportedError,
+  CCTOperationUnsupportedError,
+} from '../../errors.ts'
+
+/** Supported pool contract types; unsupported values fail in {@link parseTokenPoolVersion}. */
+export const TOKEN_POOL_TYPES = ['BurnMintTokenPool', 'LockReleaseTokenPool'] as const
```
mervin-link, line 22: Should we also add support for `BurnFromMintTokenPool` and `BurnWithFromMintTokenPool`?

apedrob: I'm deferring it to the deploy pool work where I'll introduce these variants abis and bytecodes from contracts-ccip

ccip-sdk/src/cct/evm/index.ts
apedrob: Yes, added!

mervin-link (approved):

ccip-sdk/src/cct/evm/token-pool/version.ts
aelmanaa, line 22: This allowlist only recognizes `BurnMintTokenPool` and `LockReleaseTokenPool`, but there are four other mainstream pool types in the standard burn-mint/lock-release family that share the same `TokenPool` base surface and are common deployments: `BurnFromMintTokenPool`, `BurnWithFromMintTokenPool`, `BurnToAddressTokenPool`, and `SiloedLockReleaseTokenPool`. All four ship across 1.5.1 / 1.6.1 / 2.0.0 (verified against the `chainlink-ccip` pool sources at those tags). Because `parseTokenPoolVersion` throws `CCTContractTypeInvalidError` for any unlisted type, and every op routed through `resolveTokenPool` inherits that rejection, a caller cannot use `transferOwnership` on any of these pools — even though `transferOwnership(address)` is byte-identical on all of them (it lives on `Ownable2StepMsgSender`, which the `TokenPool` base inherits).

Two options:
- Broaden this to the six standard-family types (`BurnMintTokenPool`, `LockReleaseTokenPool`, plus the four above), with a comment noting that specialized pools (`CrossChainPoolToken`, the USDC/CCTP/Lombard family) are phased to a later milestone and `BurnMintFastTransferTokenPool` is out of scope. This documents the boundary as a product decision rather than an oversight.
- For version- and type-independent ops like `transferOwnership`, don't gate on pool type at all — resolve metadata for the ABI but don't reject when the encoder floor-matches. This covers the standard six and the phased seven for free for the ops where type genuinely doesn't matter.

ccip-sdk/src/cct/evm/token-pool/version.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      TokenPoolVersion.V1_5_1,
+      TokenPoolVersion.V1_6_1,
+      TokenPoolVersion.V2_0_0,
+    ])
+  })
+
+  it('isTokenPoolVersion narrows known versions and rejects others', () => {
+    assert.equal(isTokenPoolVersion(TokenPoolVersion.V1_5_1), true)
+    assert.equal(isTokenPoolVersion(TokenPoolVersion.V2_0_0), true)
+    assert.equal(isTokenPoolVersion('1.6.0'), false)
```
aelmanaa, line 48: This assertion encodes the claim that `1.6.0` is never a real pool version. That isn't fully accurate: at the `contracts-ccip-v1.6.0` package tag, `SiloedLockReleaseTokenPool` is stamped `"SiloedLockReleaseTokenPool 1.6.0"` (the other pools at that tag are stamped `1.5.1`, but this one was bumped). So a `SiloedLockReleaseTokenPool` deployed from the v1.6.0 package reports `1.6.0` on-chain and would be rejected by the exact-match version allowlist — on top of already being rejected by the two-entry type allowlist.

This is the same pool family already flagged on the type allowlist. If the version allowlist stays exact-match, consider whether `1.6.0` should be admitted for `SiloedLockReleaseTokenPool`, or whether the floor-match approach used for encoder selection should also apply to version acceptance. At minimum, the test comment should note that `1.6.0` is a real on-chain string for one pool so this rejection is a deliberate scoping choice, not a statement that no such pool exists.

ccip-sdk/src/cct/evm/token-pool/operations/transfer-ownership.ts
```diff
@@ -0,0 +1,58 @@
...
+    validateAddress(this.name, 'newOwner', newOwner)
+  }
+
+  /** Reads the pool's type-and-version, then floor-matches the encoder and its ABI. */
+  protected async buildUnsigned(
+    chain: EVMChain,
+    { poolAddress, newOwner }: TransferOwnershipParams,
+  ): Promise<UnsignedEVMTx> {
+    const { version, abi } = await resolveTokenPool(chain, poolAddress)
+    return resolveEncoder(this.encoders, version, this.name)(abi, { poolAddress, newOwner })
```
aelmanaa, line 56: `transferOwnership` registers a single encoder at `V1_5_0` that floor-matches to every version — the encoding is version-stable — yet `buildUnsigned` still forces a full `resolveTokenPool` round-trip (an RPC `typeAndVersion` call plus the type+version allowlist check). This is inconsistent with the sibling `setPool` op, which documents "TAR.setPool encoding is version-stable across v1.5–v2.0; no version dispatch needed" and does zero `typeAndVersion` probing.

The probe imports failure modes the op doesn't have: the type allowlist can reject valid pools (see the comment on `TOKEN_POOL_TYPES`), and the RPC + ABI load is wasted work for a payload that ignores both. Two options:
- Skip `resolveTokenPool` for this op — validate addresses, encode `transferOwnership(newOwner)` against a cached ABI, submit. This matches the `setPool` precedent and removes the type-gate blast radius.
- Keep the probe but make it non-fatal for version-independent ops: resolve for the ABI, but don't reject unknown types/versions when the encoder floor-matches.

Whichever rule is chosen, document it next to the encoder table so the next op author copies the right pattern.

ccip-sdk/src/cct/evm/token-pool/version.ts
```diff
@@ -0,0 +1,125 @@
...
+    )
+  if (!isTokenPoolVersion(version))
+    throw new CCTContractVersionUnsupportedError(contractType, version, { context: { address } })
+  return { type: contractType, version }
+}
+
+/** Vendored pool ABIs keyed by {@link TokenPoolVersion}.
+ * TODO: split per type once BurnMint ABIs are imported from `@chainlink/contracts-ccip`  */
+export const TOKEN_POOL_ABIS: Record<TokenPoolVersion, InterfaceAbi> = {
+  [TokenPoolVersion.V1_5_0]: LockReleaseTokenPool_1_5,
```
aelmanaa, line 80: The ABI map routes 1.5.x and 1.6.1 to `LockReleaseTokenPool_*` ABIs and 2.0.0 to the generic `TokenPool_2_0`, and `tokenPoolAbi` ignores `type` entirely. This is harmless for `transferOwnership` because `Ownable2Step` lives in the shared `TokenPool` base, so the calldata encodes correctly regardless of which concrete ABI is used. But the moment a per-type op lands (e.g. `getSupportedTokens`, rate-limit config), a `BurnMintTokenPool` handed a `LockReleaseTokenPool` ABI will mis-encode.

The existing TODO on line 78 already flags this ("split per type once BurnMint ABIs are imported"). Worth adding a guard or test that forces the first per-type op to do the split rather than silently inheriting the LockRelease ABI — otherwise the TODO is easy to miss when the per-type op is written.

ccip-sdk/src/cct/evm/token-pool/operations/transfer-ownership.ts
```diff
@@ -0,0 +1,58 @@
...
+  poolAddress: string
+  newOwner: string
+  sender?: string
+}
+
+/** Encodes `transferOwnership` calldata against the resolved pool ABI. */
+type Encoder = (abi: InterfaceAbi, params: TransferOwnershipParams) => UnsignedEVMTx
+
+const encodeTransferOwnership: Encoder = (abi, { newOwner, poolAddress }) => {
+  const data = new Interface(abi).encodeFunctionData('transferOwnership', [newOwner])
```
aelmanaa, line 28: `new Interface(abi)` is reconstructed on every call. The sibling `setPool` op encodes against a shared cached `interfaces.TokenAdminRegistry` from `evm/const.ts`. Once the ABI is version-resolved, cache the `Interface` per ABI (or per version) instead of reconstructing it per invocation, to match the repo's cached-interface convention.

ccip-sdk/src/cct/evm/token-pool/operations/transfer-ownership.ts
```diff
@@ -0,0 +1,58 @@
...
+import { ChainFamily } from '../../../../networks.ts'
+import { EVMOperation } from '../../operation.ts'
+import { validateAddress } from '../../validate.ts'
+import { TokenPoolVersion, resolveEncoder, resolveTokenPool } from '../version.ts'
+
+/** Parameters for {@link TransferOwnership}. */
+export interface TransferOwnershipParams {
+  poolAddress: string
+  newOwner: string
+  sender?: string
```
aelmanaa, line 21: `sender` has no TSDoc on the field itself. The facade `@example` explains it ("`sender` must be the pool's current owner"), but a consumer using autocomplete on the param object sees a bare `sender?: string`. Add a one-line TSDoc: current pool owner; sets `tx.from` for offline/multisig signing.

PR #290, opened 2026-07-07
feat(cct-sdk): Add solana set pool feature

mervin-link:
## What

- [DAPP-10460](https://smartcontract-it.atlassian.net/browse/DAPP-10460)
- Add the CCT SDK for solana at `ccip-sdk/cct/solana`
- Add Solana CCT `TokenAdminRegistry` `setPool` support
- Add PDA derivation helpers for Solana router programs

## Why

- Enables Solana CCT users to generate and submit `setPool` transactions through the SDK

[DAPP-10460]: https://smartcontract-it.atlassian.net/browse/DAPP-10460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token-admin-registry/operations/set-pool.ts
```diff
@@ -0,0 +1,76 @@
...
+    const payer = new PublicKey(opts.payer)
+    const authority = new PublicKey(opts.authority ?? opts.payer)
+    const lookupTable = new PublicKey(opts.poolLookupTableAddress)
+
+    const routerProgram = createRouterProgram(chain, router, payer)
+    const config = deriveRouterConfigPda(router)
+    const tokenAdminRegistry = deriveTokenAdminRegistryPda(router, tokenMint)
+
+    const instruction = await routerProgram.methods
+      .setPool(Buffer.from([3, 4, 7]))
```
aelmanaa: **(blocker — correctness).** The `writable_indexes` payload is hardcoded to `[3, 4, 7]`. That value is correct only for the standard 9-account BurnMint/LockRelease pool lookup table. `writable_indexes` is fundamentally caller-specific — it's the set of positions in the caller's *own* pool lookup table that the Router must mark writable during a transfer, and that layout varies per deployment. A pool whose ALT carries additional accounts (custom/extended pools) needs those extra slots flagged writable, which is impossible while the value is a constant → such a pool is mis-registered and its CPI fails at execution time. The value belongs to the caller, not the SDK (the reference multisig generator takes it as a required, 0–255-validated argument).

Fix: add `writableIndexes?: number[]` to the params, default to a named `DEFAULT_WRITABLE_INDEXES = [3, 4, 7]` for the standard-pool convenience case, validate (non-empty, each an integer 0–255), and TSDoc that it indexes the caller's pool ALT and must be extended for custom/larger lookup tables.

ccip-sdk/src/cct/solana/operation.ts
```diff
@@ -0,0 +1,50 @@
...
+    const { wallet } = params
+    if (!isWallet(wallet)) throw new CCIPWalletInvalidError(wallet)
+
+    params.payer = wallet.publicKey.toBase58()
+    const unsigned = await this.generate(chain, params)
+
+    try {
+      return { hash: await simulateAndSendTxs(chain, wallet, unsigned) }
+    } catch (error) {
+      throw new CCTTxFailedError(
```
aelmanaa: **(major — correctness).** Every `simulateAndSendTxs` failure is funneled into `CCTTxFailedError`, whose constructor defaults `isTransient` to `false` — so all failures are asserted *permanent*. A blockhash expiry / RPC blip / send timeout is therefore reported as non-retryable, and retry-on-transient consumers won't retry a recoverable send. `CCTTxNotConfirmedError` (transient, with `retryAfterMs`) already exists but is never thrown here.

Fix: discriminate — a program/simulation revert → permanent `CCTTxFailedError`; blockhash/RPC/timeout → transient `CCTTxNotConfirmedError`. Mirror the classification the EVM `submit` path uses so both families behave the same.

ccip-sdk/src/cct/solana/operation.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.validate(params)
+    return this.encode(chain, params)
+  }
+
+  /** Generate, sign, simulate, send, and confirm with wallet.publicKey as payer. */
+  async execute(chain: SolanaChain, params: P & { wallet: unknown }): Promise<TransactionHash> {
+    const { wallet } = params
+    if (!isWallet(wallet)) throw new CCIPWalletInvalidError(wallet)
+
+    params.payer = wallet.publicKey.toBase58()
```
aelmanaa: **(minor — DX).** `execute` overwrites `params.payer` with the wallet's key and mutates the caller's object in place. Meanwhile the public `setPool` / `generateUnsigned*` facade requires `payer` in its params type — so the signed path demands a field it then ignores, and the mutation is an observable side effect on the caller's input.

Fix: make `payer` optional on the signed path (or use a distinct params type for it), and derive locally without mutating — e.g. `const payer = wallet.publicKey.toBase58()` then pass `{ ...params, payer }` to `generate`.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { SolanaChain } from '../../solana/index.ts'
+import type { UnsignedSolanaTx } from '../../solana/types.ts'
+import { TokenManager } from '../token-manager.ts'
+import { type SerializedSolanaTxEncoding, serializeUnsignedSolanaTx } from './serialize.ts'
+import { SolanaTokenAdminRegistryClient } from './token-admin-registry/index.ts'
+
+/** CCT admin facade for Solana; grouped clients own contract/program operations. */
+export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana> {
+  readonly chain: SolanaChain
+  readonly tokenAdminRegistry: SolanaTokenAdminRegistryClient
```
aelmanaa: **(major — DX / cross-family consistency).** The operation is reached as `manager.tokenAdminRegistry.setPool(...)`, but the EVM manager exposes it flat as `manager.setPool(...)` — two different call shapes for the same op across families (and EVM keeps `token-admin/` purely as internal folder structure, which is the shape to match). A single operation doesn't justify the extra facade hop.

Suggest flattening to `SolanaTokenManager.setPool()` / `.generateUnsignedSetPool()`, identical to EVM, keeping `token-admin-registry/` as internal structure. If a second program's operations ever land, add sub-clients on both families together.

ccip-sdk/src/cct/solana/serialize.ts
```diff
@@ -0,0 +1,46 @@
...
+    throw new CCTParamsInvalidError(
+      'serializeUnsignedTx',
+      'lookupTables',
+      'legacy-message serialization does not support address lookup tables',
+    )
+  }
+
+  const payerKey = typeof payer === 'string' ? new PublicKey(payer) : payer
+  const { blockhash } = await connection.getLatestBlockhash()
+  const serialized = Buffer.from(
```
aelmanaa, line 29: **(nit).** `Buffer` is used here (and in `programs/router.ts`, e.g. the PDA seeds) via the Node global rather than an explicit import. The rest of the Solana code imports it (`import { Buffer } from 'buffer'`) for portability across bundlers/runtimes — the sibling `set-pool.ts` already does. One-line fix: add the import to both files.

aelmanaa (commented):
### Solana CCT `setPool` — review (draft PoC)

Reviewed at head `214b7cd`. Structurally this is in good shape: it adopts the shared `Operation`/`TokenManager` bases, returns the shared `{ hash }` result, derives the payer from the wallet, delegates address validation to `SolanaChain.getAddress`, and the offline serializer (base58/base64/hex) is a genuinely useful surface the EVM side lacks. A full signed `setPool` cycle (delist → non-default ALT → delist) was verified end-to-end on devnet.

Not mergeable as-is — details inline:
- **blocker:** the `writable_indexes` payload is hardcoded (`[3, 4, 7]`) — correct only for the standard pool layout, wrong for custom/extended pools; it must be a caller parameter.
- **major:** every send failure is classified permanent, so transient failures (blockhash expiry / RPC / timeout) won't be retried.
- **major (DX):** the facade shape (`manager.tokenAdminRegistry.setPool(...)`) diverges from EVM's flat `manager.setPool(...)` for the same operation.
- **minor:** the signed path requires a `payer` it immediately overwrites (and mutates the caller's object); `Buffer` is used via the global in two files.

Also **inherited from the base branch**: the shared CCT error classes (`CCTTxFailedError`, `CCTParamsInvalidError`, `CCTTxNotConfirmedError`) aren't re-exported from any public entry point (`errors/index.ts`, the root, or the `./cct/*` subpaths), so consumers can't `instanceof`-catch them — the same export gap already flagged on the EVM PR. Fixing it there fixes it here.

Nits: no `fromUrl` constructor (only `fromChain`, though `SolanaChain` has it); params fields lack TSDoc/`@example` — the ALT prerequisite and zero-pool delist semantics are worth documenting.

ccip-sdk/src/cct/solana/serialize.ts
mervin-link: Fixed.

ccip-sdk/src/cct/solana/index.ts
mervin-link: I talked to Pedro and we decided to group these ops by contract/program. There are some duplicated functions from different contracts, like `transferOwnership`, which exist for both the token and the token pool.

ccip-sdk/src/cct/solana/validate.ts
```diff
@@ -0,0 +1,46 @@
...
+  try {
+    SolanaChain.getAddress(value)
+  } catch (error) {
+    throw new CCTParamsInvalidError(
+      operation,
+      param,
+      `must be a valid Solana public key, got ${String(value)}`,
+      { cause: error instanceof Error ? error : undefined },
+    )
+  }
```
apedrob, line 28: I was facing same approach in EVM and found that this pulls the entire `SolanaChain` class into this lightweight validator. We could use `new PublicKey(value) from `@solana/web3.js`. However, if you find it's not a big deal having the chain, we can go with this approach and I change it in the EVM approach as well. 

Btw, in EVM, I've added if it make sense here:
```
{ cause: new CCIPAddressInvalidError(String(value), ChainFamily.EVM), }
```

ccip-sdk/src/cct/solana/token-admin-registry/operations/set-pool.ts
```diff
@@ -0,0 +1,78 @@
...
+  /** Builds the unsigned Solana `setPool` instruction set. */
+  protected async encode(
+    chain: SolanaChain,
+    opts: GenerateSetPoolParams,
+  ): Promise<UnsignedSolanaTx> {
+    const routerAddress = await chain.getTokenAdminRegistryFor(opts.address)
+    const router = new PublicKey(routerAddress)
+    const tokenMint = new PublicKey(opts.tokenAddress)
+    const payer = new PublicKey(opts.payer)
+    const authority = new PublicKey(opts.authority ?? opts.payer)
```
apedrob, line 83: Although it might be fine for EOA, in multisig flow the fee payer != admin as we faced out in TMEM. Suggest making authority required or document this default

ccip-sdk/src/cct/solana/token-admin-registry/__tests__/index.test.ts
apedrob, line 1: The core sdk uses flat in same dir `*.tests.ts` instead of `__tests__` subdir

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  /** Creates a Solana CCT manager for an existing chain. */
+  constructor(chain: SolanaChain) {
+    super()
+    this.chain = chain
+    this.tokenAdminRegistry = new SolanaTokenAdminRegistryClient(chain)
+  }
+
```
apedrob, line 33: We should also offer also offer `fromUrl`/`fromProvider` constructors

ccip-sdk/src/cct/solana/submit.ts
```diff
@@ -0,0 +1,64 @@
...
+export function createCCTSubmitError(
+  operation: string,
+  error: unknown,
+): CCTTxFailedError | CCTTxNotConfirmedError {
+  const signature = getSignature(error)
+  if (signature) {
+    return new CCTTxNotConfirmedError(operation, signature, {
+      cause: error instanceof Error ? error : undefined,
+    })
+  }
```
apedrob: Note: Given tx simulates first, a tx that passes simulation then fails on-chain (state changed between sim and send) would carry a signature and can be misclassified here

apedrob (commented):
Following up our convo, I believe we can clean up some test files given the operations tests will carry the most of the logic too

ccip-sdk/src/cct/solana/token-admin-registry/operations/set-pool.ts
mervin-link: I will just add a comment for document. Payer should always be the authority for single-signer transactions. Multisig users will use authority to override it.

ccip-sdk/src/cct/solana/token-admin-registry/__tests__/index.test.ts
mervin-link: They are using `__tests__` in solana. Do we want all our tests per op inside `operations`?

ccip-sdk/src/cct/solana/index.ts
aelmanaa: The core issue here is developer-experience consistency: someone using the SDK to call `setPool` should do it the same way regardless of chain. Today they can't. EVM is `manager.setPool(...)`, Solana is `manager.tokenAdminRegistry.setPool(...)`. Same operation, two different call shapes, is a real DevX split for anyone integrating across chains, and `setPool` is the only op we're shipping.

On the grouping rationale: `transferOwnership` is the generic `OwnerIsCreator` function, uniform across every CCIP contract (router, token pool, registry, ramps) with the same signature. That makes it a flat, address-keyed op too, `manager.transferOwnership({ address, newOwner })` pointed at any contract, the same polymorphic-address convention `getTokenAdminRegistryFor` and `setPool`'s `address` param already use. So it doesn't require per-program sub-clients either.

Proposal: keep `setPool` flat on both facades, matching EVM, so the call is identical across chains. If a real need for namespacing shows up later, add it to both families in one change. The thing to avoid is shipping one operation two different ways.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  serializeUnsignedTx(
+    unsigned: Pick<UnsignedSolanaTx, 'instructions' | 'lookupTables'>,
+    payer: string,
+    encoding?: SerializedSolanaTxEncoding,
+  ): Promise<string> {
+    return serializeUnsignedSolanaTx(this.provider, unsigned, payer, encoding)
+  }
+}
+
+export type { GenerateSetPoolParams, SetPoolParams } from './token-admin-registry/index.ts'
```
aelmanaa: The shared CCT error classes are re-exported from `cct/evm/index.ts` (`export * from '../errors.ts'`) but not from here. So a `./cct/solana` consumer can't `instanceof`-catch `CCTTxFailedError` / `CCTParamsInvalidError` / `CCTTxNotConfirmedError`, while a `./cct/evm` consumer can. Could we add the same re-export here (plus the params/result types) so both subpaths are consistent?

ccip-sdk/src/cct/solana/index.ts
apedrob: @aelmanaa here we are proposing the both EVM and Solana will follow the grouping pattern, i.e. `manager.tokenAdminRegistry.setPool(...)`, given the conflict of same method names on different contracts, such as the `transferOwnership` that exist on both token and token pool contracts

ccip-sdk/src/cct/solana/token-admin-registry/operations/set-pool.ts
```diff
@@ -0,0 +1,96 @@
...
+
+/** Standard BurnMint/LockRelease pool ALT writable positions. */
+export const DEFAULT_WRITABLE_INDEXES = [3, 4, 7] as const
+
+/** Parameters shared by Solana TokenAdminRegistry `setPool` generation and execution. */
+type SetPoolParams = {
+  tokenAddress: string
+  address: string
+  poolLookupTableAddress: string
+  writableIndexes?: number[]
```
aelmanaa, line 40: **(nit — DX, follow-up to the resolved `writableIndexes` thread).** The param + `DEFAULT_WRITABLE_INDEXES` + validation all landed as prescribed, but the TSDoc that thread also asked for ("indexes the caller's pool ALT and must be extended for custom/larger lookup tables") was never added — and this is the one field where the miss has teeth: a wrong/short value on a custom pool passes validation but fails at message-execution time, and the `[3,4,7]` default masks the requirement. Suggest:

```ts
/**
 * Positions in the pool's own Address Lookup Table the Router marks writable during a
 * transfer. Defaults to {@link DEFAULT_WRITABLE_INDEXES} for standard BurnMint/LockRelease
 * pools; custom pools with extra accounts MUST extend this or the pool CPI gets wrong
 * write-permissions and fails at execution. Each entry is a byte (0–255).
 */
writableIndexes?: number[]
```

ccip-sdk/src/cct/solana/token-admin-registry/operations/set-pool.ts
```diff
@@ -0,0 +1,96 @@
...
+} from '../../programs/router.ts'
+import { validatePublicKey, validateWritableIndexes } from '../../validate.ts'
+
+/** Standard BurnMint/LockRelease pool ALT writable positions. */
+export const DEFAULT_WRITABLE_INDEXES = [3, 4, 7] as const
+
+/** Parameters shared by Solana TokenAdminRegistry `setPool` generation and execution. */
+type SetPoolParams = {
+  tokenAddress: string
+  address: string
```
aelmanaa, line 31: **(nit — DX / cross-family parity).** `address` is opaque at the field level — it's fed to `getTokenAdminRegistryFor(opts.address)`, i.e. it resolves the TokenAdminRegistry/Router from any CCIP contract. EVM documents this field; the Solana side has no TSDoc. One line restores parity:

```ts
/**
 * CCIP contract to resolve the TokenAdminRegistry/Router from — the registry itself,
 * a Router, OnRamp, OffRamp, or TokenPool address all work.
 */
address: string
```

ccip-sdk/src/cct/solana/token-admin-registry/operations/set-pool.ts
```diff
@@ -0,0 +1,96 @@
...
+  poolLookupTableAddress: string
```
aelmanaa, line 33: **(nit — DX).** Non-obvious provenance — this is the pool's ALT produced by the separate `createLookupTable` op, not an arbitrary address. A one-liner helps chaining the two ops:

```ts
/** The pool's Address Lookup Table address, produced by the `createLookupTable` op. */
poolLookupTableAddress: string
```

apedrob (approved):

aelmanaa (approved):

PR #291, opened 2026-07-09
feat(cct-sdk): Add create and append lookup table solana op

mervin-link:
## What

- [DAPP-10480](https://smartcontract-it.atlassian.net/browse/DAPP-10480) and [DAPP-10500](https://smartcontract-it.atlassian.net/browse/DAPP-10500)
- Add Solana CCT `createLookupTable` and `appendToLookupTable` operation for generating pool lookup table instructions
- Support typed Solana unsigned results with extra metadata such as `lookupTableAddress`
- Add optional `computeUnits` support to Solana CCT execute params

## Why

- Solana token pool setup needs a reusable SDK operation to build ALT instructions and return the derived lookup table address
- Future Solana CCT operations may also need generated metadata results alongside unsigned transaction instructions

[DAPP-10480]: https://smartcontract-it.atlassian.net/browse/DAPP-10480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DAPP-10500]: https://smartcontract-it.atlassian.net/browse/DAPP-10500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token-admin-registry/operations/create-lookup-table.ts
```diff
@@ -0,0 +1,132 @@
...
+    const poolProgram = new PublicKey(opts.poolProgramAddress)
+    const tokenMint = new PublicKey(opts.tokenAddress)
+    const payer = new PublicKey(opts.payer)
+    const authority = new PublicKey(opts.authority ?? opts.payer)
+    const additionalAddresses = (opts.additionalAddresses ?? []).map((a) => new PublicKey(a))
+
+    const [createIx, lookupTableAddress] = AddressLookupTableProgram.createLookupTable({
+      authority,
+      payer,
+      recentSlot: await chain.connection.getSlot(),
```
aelmanaa: **(blocker — correctness).** `getSlot()` here inherits `SolanaChain`'s `Connection` default commitment of `confirmed`. The Address Lookup Table program requires `recentSlot` to already be in the `SlotHashes` sysvar, and a `confirmed` tip often isn't yet, so create fails intermittently with `InvalidInstructionData` (reproduced live on devnet: 1 of 2 back-to-back runs failed). Canonical tooling uses `finalized` (`lookuptable.go` does `GetSlot(CommitmentFinalized)` then `slot - 1`, with a comment that the most recent slot sometimes errors), and web3.js defaults to `finalized`.

Fix: `await chain.connection.getSlot('finalized')` (optionally `- 1`). Note the same latent issue exists in core `solana/exec.ts` (`getSlot('confirmed')`), so this is a good candidate for a shared helper that fixes both.

ccip-sdk/src/cct/solana/token-admin-registry/operations/create-lookup-table.ts
```diff
@@ -0,0 +1,132 @@
...
+/** Parameters for unsigned Solana lookup table generation. */
+export type GenerateCreateLookupTableParams = SolanaGenerateParams<CreateLookupTableParams>
+
+/** Unsigned create lookup table result, including the derived ALT address. */
+export type GenerateCreateLookupTableResult = UnsignedSolanaTx & {
+  lookupTableAddress: string
+}
+
+/** Submitted create lookup table result. */
+export type CreateLookupTableResult = TransactionHash
```
aelmanaa: **(blocker — DX / op-chaining).** The signed path returns a bare `TransactionHash` (`CreateLookupTableResult = TransactionHash`), so `createLookupTable(...)` yields `{ hash }` with no `lookupTableAddress`. But the ALT address is the whole point of this op and the exact input the next step needs (`setPool({ poolLookupTableAddress })`). It's a PDA of `[authority, recentSlot]` derived inside `buildUnsigned`, so after `execute` returns the caller can't re-derive it (the `recentSlot` isn't exposed) and would have to scan the confirmed tx's accounts.

The unsigned result (`GenerateCreateLookupTableResult`) already includes `lookupTableAddress`; the signed one should too. `SolanaOperation` added a `Tx` generic but not the `Result` generic the base `Operation<Chain, Params, Tx, Result>` already provides. Fix: thread `Result` through so `execute` returns `{ hash, lookupTableAddress }`, and make `CreateLookupTableResult = TransactionHash & { lookupTableAddress: string }`. Worth doing before the bare alias is published, since widening it later is a breaking change.

ccip-sdk/src/cct/solana/token-admin-registry/operations/create-lookup-table.ts
```diff
@@ -0,0 +1,148 @@
...
+
+const MAX_ALT_ADDRESSES = 256
+const EXTEND_CHUNK_SIZE = 30
+
+/** Parameters shared by Solana TokenAdminRegistry `createLookupTable` generation and execution. */
+type CreateLookupTableParams = {
+  tokenAddress: string
+  poolProgramAddress: string
+  additionalAddresses?: string[]
+  authority?: string
```
aelmanaa: **(design scope — vault-owned ALT / multisig).** The signed one-shot here only works for an EOA-owned ALT. `execute` bundles `createLookupTable` + `extendLookupTable` in one transaction signed by the wallet, but the two instructions differ: `create` does not require the `authority` to sign, while `extend` does. So `authority === wallet` works, but `authority` = a Squads vault fails at broadcast (the extend is unsigned; simulation passes since `sigVerify` is off, so the failure is confusing). There's also a ~60-90s fuse, since the ALT address is a PDA of `[authority, recent_slot]` and the create instruction re-checks the slot at execution time.

We do want to support vault-owned ALTs, i.e. the Squads vault is the ALT authority so it controls extend/deactivate/close, which the one-shot can't express. Our own tool already implements exactly this pattern: https://github.com/smartcontractkit/ccip-solana-bs58-generator

Two pieces are needed alongside the current one-shot:

1. A create-empty mode: create the ALT with `authority = vault`, paid and signed by an EOA, without bundling `extend`. This is legal because `create` needs no authority signature, so an EOA can create a vault-owned table. It must return the ALT address so the table can be populated later (same return-the-address fix already noted on this file).
2. An `appendToLookupTable` op that builds `extend` with `authority = vault`, reachable via the unsigned/serialize path so Squads signs and executes it. `extend` has no slot fuse, so multisig approval latency is fine.

The current bundled create+extend can stay as the EOA-owned convenience path (`authority = wallet`). The key split to get right: create-empty is EOA-signable even when the authority is the vault, but append must go through the unsigned/Squads path since only the vault can sign the extend.

mervin-link: I added a `mode` field (`createAndExtend` or `createEmpty`) to support both EOA and vault flows. I also added a check to ensure `authority === payer` when `mode` is `createAndExtend`.

I will create another PR for `appendToLookupTable` op.

ccip-sdk/src/cct/solana/token-admin-registry/operations/create-lookup-table.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  /** Builds unsigned ALT create instructions, optionally with extend instructions. */
+  protected async buildUnsigned(
+    chain: SolanaChain,
+    opts: GenerateCreateLookupTableParams,
+  ): Promise<GenerateCreateLookupTableResult> {
+    const payer = new PublicKey(opts.payer)
+    const authority = new PublicKey(opts.authority ?? opts.payer)
+
+    const [createIx, lookupTableAddress] = AddressLookupTableProgram.createLookupTable({
```
aelmanaa: **(blocker — correctness, `createEmpty` mode).** The `createEmpty` mode added for vault-owned ALTs can't actually create one. `buildUnsigned` builds the create instruction via `AddressLookupTableProgram.createLookupTable(...)` here, and web3.js marks the `authority` account as `isSigner: true` on that instruction (pre-relaxation account metas). So `createEmpty` with `authority = vault` emits a transaction that demands the vault's signature, which the paying EOA can't provide, and it fails. The mode therefore only works when `authority = wallet` (where `createEmpty` is pointless) and fails for the vault-owned case it exists for — the same case the `createAndExtend` guard points people to ("Use mode: 'createEmpty' for vault-owned ALTs").

The on-chain ALT program dropped the authority-must-sign requirement on create years ago via the `relax_authority_signer_check_for_lookup_table_creation` feature gate, but the web3.js helper still emits the old metas. This is exactly why the reference tool (`smartcontractkit/ccip-solana-bs58-generator`, `scripts/create-alt.ts`) hand-builds the create instruction with `authority: isSigner: false`.

**Fix** — hand-build the create ix (byte-identical to the web3.js helper except the single `isSigner` flag on the authority):

```ts
import {
  AddressLookupTableProgram,
  PublicKey,
  SystemProgram,
  TransactionInstruction,
} from '@solana/web3.js'
import { Buffer } from 'buffer'

function buildCreateLookupTableIx(
  authority: PublicKey,
  payer: PublicKey,
  recentSlot: bigint,
): { instruction: TransactionInstruction; lookupTableAddress: PublicKey } {
  const slotBuf = Buffer.alloc(8)
  slotBuf.writeBigUInt64LE(recentSlot)
  // ALT address is a PDA of [authority, recent_slot_le] — same derivation web3.js uses
  const [lookupTableAddress, bump] = PublicKey.findProgramAddressSync(
    [authority.toBuffer(), slotBuf],
    AddressLookupTableProgram.programId,
  )
  const data = Buffer.alloc(13)
  data.writeUInt32LE(0, 0) // CreateLookupTable discriminator
  data.writeBigUInt64LE(recentSlot, 4) // recent_slot: u64
  data.writeUInt8(bump, 12) // bump_seed: u8
  const instruction = new TransactionInstruction({
    programId: AddressLookupTableProgram.programId,
    keys: [
      { pubkey: lookupTableAddress, isSigner: false, isWritable: true },
      { pubkey: authority, isSigner: false, isWritable: false }, // web3.js hardcodes true
      { pubkey: payer, isSigner: true, isWritable: true },
      { pubkey: SystemProgram.programId, isSigner: false, isWritable: false },
    ],
    data,
  })
  return { instruction, lookupTableAddress }
}
```

Strictly only `createEmpty` needs this (the `createAndExtend` guard forces `authority = wallet`, which signs anyway), but using it for both modes keeps one code path and changes nothing behaviorally: a signature that was required-and-provided just becomes not-required-but-provided; account order and data are identical. Worth a test asserting the built create ix has `authority.isSigner === false` (this is the regression the web3.js helper would silently reintroduce).

**Devnet proof (head `dae8d03`, same wallet/vault/finalized-slot source, only the `isSigner` flag differs between the failing and passing runs):**
- **[A] static** — `generateUnsignedCreateLookupTable({ mode: 'createEmpty', authority: <vault>, payer: <wallet> })` returns a create ix with `authority.isSigner = true` (inspected from `instructions[0].keys`).
- **[B] bug on-chain** — `createLookupTable({ mode: 'createEmpty', authority: <vault>, wallet })` failed with `CCTTxFailedError` (simulation rejected; the tx message demands the vault's signature the wallet can't provide).
- **[C] fix on-chain** — same params, hand-built ix (`authority: isSigner: false`): success. Tx `2VwfyN4nEEgXQ3uBgUy5K26bSTFtwBtiZ4cJEFCSHLboXnZwH3RbT37Pry4kemJJ7wAuhZJAvhE5qMTHdc26jgn` — ALT `FzfqnkwMuLrgvZSqh5H5Uy5FkLCGwuvAg1yQ7HCvPVfv`, 0 entries, `authority = F1rZWMSWuAQRqtS5SYVirQ7GAsPigAgsmVdWxUs3SBRa` (a fresh keypair that never signed anything).

**Real Squads v4 end-to-end (with the `appendToLookupTable` op from the child PR):** created a real multisig `G4Av8VixyZ7rzzfnfoT51EDHkbaWkAMPQ6XDNz9bjZCH` (vault PDA `JbyFektCPqiEFD9jcGcbzc7Q4nVRCWf9Q3oVdmeKUmB`); the hand-built create produced a vault-owned empty ALT `HbK7RXWGiq9b2gPCvftq3NK4pRPHzsiTNUixxbiwvADf` (authority = vault) with a payer-only signature; the append op then populated it through the full Squads proposal lifecycle (`vaultTransactionCreate` -> `proposalCreate` -> `proposalApprove` -> `vaultTransactionExecute`, the vault signing extend via `invoke_signed`) to the 10 canonical entries, authority = vault.

This gates the whole vault-owned flow: the append op works end-to-end, but there is no vault-owned ALT to append to until this create fix lands. Natural home is a shared `programs/alt.ts` next to the append helper.

mervin-link: Updated.

I created `buildCreateLookupTableInstruction` in `programs/alt.ts` to be used by both modes. Also tested it manually to make sure it works when the authority is different from the payer.

ccip-sdk/src/cct/solana/programs/fee-quoter.ts
```diff
@@ -0,0 +1,9 @@
+import { PublicKey } from '@solana/web3.js'
+
+/** Derives the FeeQuoter billing token config PDA for a mint. */
+export function deriveFeeBillingTokenConfigPda(feeQuoter: PublicKey, mint: PublicKey): PublicKey {
+  return PublicKey.findProgramAddressSync(
+    [Buffer.from('fee_billing_token_config'), mint.toBuffer()],
```
aelmanaa, line 8: **(nit — portability / convention).** This file (and `programs/token-pool.ts`) uses the global `Buffer` without importing it: `Buffer.from('fee_billing_token_config')` here, and `Buffer.from('ccip_tokenpool_config')` / `Buffer.from('ccip_tokenpool_signer')` in `token-pool.ts`.

It compiles and passes lint because `Buffer` is a Node global and `import-x/no-nodejs-modules` only inspects imports — but it breaks the SDK's browser-portability contract, where `Buffer` isn't global and needs an explicit import for bundlers to polyfill. Every sibling module follows the convention: `programs/router.ts`, the new `programs/alt.ts`, and `operations/set-pool.ts` all do `import { Buffer } from 'buffer'`.

Fix: add `import { Buffer } from 'buffer'` to both `programs/fee-quoter.ts` and `programs/token-pool.ts`.

ccip-sdk/src/cct/solana/token-admin-registry/operations/create-lookup-table.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    if (addresses.length > MAX_ALT_ADDRESSES) {
+      throw new CCTParamsInvalidError(
+        this.name,
+        'additionalAddresses',
+        `ALT cannot exceed ${MAX_ALT_ADDRESSES} addresses; requested ${addresses.length}`,
+      )
+    }
+
+    const extendIxs = []
```
aelmanaa: **(nit — TS).** `const extendIxs = []` is an untyped evolving array. Prefer an explicit element type for clarity: `const extendIxs: TransactionInstruction[] = []`.

aelmanaa (approved):

ccip-sdk/src/cct/solana/operation.ts
```diff
@@ -17,40 +17,44 @@ export type SolanaGenerateParams<P extends object> = P & { payer: string }
...
-  protected abstract buildUnsigned(
-    chain: SolanaChain,
-    params: SolanaGenerateParams<P>,
-  ): Promise<UnsignedSolanaTx>
+  protected abstract buildUnsigned(chain: SolanaChain, params: SolanaGenerateParams<P>): Promise<Tx>
+
+  /** Adds generated operation metadata to the submit result. */
+  protected resultFromGenerated(hash: TransactionHash, _tx: Tx): Result {
+    return hash as Result
+  }
```
apedrob: I was facing the same situation when [deploying an EVM token](https://github.com/smartcontractkit/ccip-tools-ts/pull/297), because the output was not just the transaction hash. What I found that worked was to have the override execute on the operation and return a type that extends the transaction hash with additional fields (e.g., `& { lookupTableAddress  }`). This satisfies type checking and eliminates the need for these helper functions (`resultFromGenerated`)

apedrob (approved):

PR #293, opened 2026-07-10
feat(cct-sdk): Add append to lookup table solana op

mervin-link:
## What

- [DAPP-10500](https://smartcontract-it.atlassian.net/browse/DAPP-10500)
- Add Solana CCT SDK support for appending addresses to an existing lookup table
- Share CCIP lookup table address derivation between create and append operations
- Add validation for ALT existence, authority mismatch, address sources, and 256-address capacity limits

## Why

 - Vault-owned ALTs need to be created empty and populated later through the vault authority, which requires an extend-only operation that can be serialized for external signing

[DAPP-10500]: https://smartcontract-it.atlassian.net/browse/DAPP-10500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token-admin-registry/operations/append-to-lookup-table.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (opts.tokenAddress && opts.poolProgramAddress) {
+      const poolProgram = new PublicKey(opts.poolProgramAddress)
+      const tokenMint = new PublicKey(opts.tokenAddress)
+      const ccipAddresses = await deriveCcipLookupTableAddresses(chain, {
+        lookupTableAddress,
+        tokenMint,
+        poolProgram,
+        authority,
+      })
+      addresses.unshift(...ccipAddresses)
```
aelmanaa, line 128: **(major — correctness).** When `tokenAddress`/`poolProgramAddress` are provided, this branch auto-derives the 10 canonical CCIP entries and `unshift`es them, but it never checks whether those entries are **already present** in the table. The existing table is read (`lookupTable.value.state.addresses`) only for the 256-cap check, not for de-duplication.

A pool ALT has a fixed, order-sensitive layout where indices `[3,4,7]` are exactly the ones `setPool` marks writable via its `writableIndexes` bitmap. If someone calls `appendToLookupTable({ tokenAddress, poolProgramAddress })` against a table that already holds the canonical block (e.g. one built by `createLookupTable` in `createAndExtend` mode), this silently appends a **second** copy at indices 10–19, corrupting the layout while `writableIndexes: [3,4,7]` still points at the first copy. The auto-derive path is meant for populating an ALT created **empty** via `createEmpty`, but the API can't distinguish that from misuse on an already-populated table.

**Fix:** when auto-deriving, guard the canonical block — detect that it already exists (e.g. index 0 self-ref / entries `[0..9]` already present) and either skip it or throw a teaching `CCTParamsInvalidError`. Scope the guard to the CCIP-derived block only; leave the `additionalAddresses`-only path untouched so callers can still freely append custom (non-CCIP) accounts for custom pools. Duplicates among plain `additionalAddresses` are wasteful but non-corrupting, so at most a debug log there — no hard block, to avoid fighting repair/retry flows.

ccip-sdk/src/cct/solana/token-admin-registry/operations/append-to-lookup-table.ts
```diff
@@ -0,0 +1,171 @@
...
+  ): Promise<ExecuteAppendToLookupTableResult> {
+    const { wallet, computeUnits, ...rest } = params
+    if (!isWallet(wallet)) throw new CCIPWalletInvalidError(wallet)
+
+    const payer = wallet.publicKey.toBase58()
+    if (params.authority && !new PublicKey(params.authority).equals(wallet.publicKey)) {
+      throw new CCTParamsInvalidError(
+        this.name,
+        'authority',
+        'appendToLookupTable requires authority to be the executing wallet.',
```
aelmanaa: **(nit — DX).** Unlike the equivalent guard in `create-lookup-table.ts` (`"... Use mode: 'createEmpty' for vault-owned ALTs."`), this message doesn't tell the user what to do instead. For a vault-owned ALT the caller must go through the unsigned-generate + Squads path — worth pointing them there, e.g. `"appendToLookupTable requires authority to be the executing wallet. Use generateUnsignedAppendToLookupTable(...) for vault-owned ALTs (sign/execute via the vault)."`

ccip-sdk/src/cct/solana/token-admin-registry/operations/append-to-lookup-table.ts
```diff
@@ -0,0 +1,171 @@
...
+    const totalAddressesAfterAppend = lookupTable.value.state.addresses.length + addresses.length
+    if (totalAddressesAfterAppend > MAX_ALT_ADDRESSES) {
+      throw new CCTParamsInvalidError(
+        this.name,
+        'additionalAddresses',
+        `ALT cannot exceed ${MAX_ALT_ADDRESSES} addresses; requested ${totalAddressesAfterAppend}`,
+      )
+    }
+
+    const instructions = []
```
aelmanaa: **(nit — TS).** `const instructions = []` is an untyped evolving array. Prefer an explicit element type for clarity: `const instructions: TransactionInstruction[] = []`.

aelmanaa (approved):

apedrob (approved):

PR #294, opened 2026-07-13
feat(cct-sdk): Add deploy token solana op

mervin-link:
## What

- [DAPP-10498](https://smartcontract-it.atlassian.net/browse/DAPP-10498)
- Add Solana CCT SDK support for deploying SPL token mints, with optional metaplex metadata
- Support both SPL Token and Token-2022 programs for mint creation

## Why

- Solana CCT users need an SDK operation to create token mints before configuring token pools and lookup tables

[DAPP-10498]: https://smartcontract-it.atlassian.net/browse/DAPP-10498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (!params.symbol || params.symbol.length > 10) {
+      throw new CCTParamsInvalidError(
+        this.name,
+        'symbol',
+        'is required and must be <= 10 characters when withMetaplex is true',
+      )
+    }
+    if (params.uri !== undefined && typeof params.uri !== 'string') {
+      throw new CCTParamsInvalidError(this.name, 'uri', 'must be a string when provided')
+    }
```
apedrob: Suggestion: Extract validators into a single file to be reusable

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,182 @@
...
+        ...createMetadataInstructions(chain, mint, payer, tokenProgram, params.decimals, {
+          name: params.name,
+          symbol: params.symbol,
+          uri: params.uri ?? '',
+        }),
+      )
+
+    chain.logger.debug(
+      `${this.name}: mint = ${mint.toBase58()}, tokenProgram = ${tokenProgram.toBase58()}`,
+    )
```
apedrob, line 339: Doubt: Should we keep these debuggers?

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,182 @@
...
+    return {
+      family: ChainFamily.Solana,
+      instructions,
+      mainIndex: 0,
+      tokenAddress: mint.toBase58(),
+    }
+  }
+}
+
+function createMetadataInstructions(
```
apedrob: Readability: Since it's called in the above function, move it to the top of that function

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
apedrob, line 1: Q: Shouldn't we override the `execute` function here, since we need to return the contract address as well?

ccip-sdk/package.json
```diff
@@ -75,6 +75,10 @@
   "dependencies": {
     "@aptos-labs/ts-sdk": "^6.3.1",
     "@coral-xyz/anchor": "^0.29.0",
+    "@metaplex-foundation/mpl-token-metadata": "^3.4.0",
+    "@metaplex-foundation/umi": "^1.5.1",
+    "@metaplex-foundation/umi-bundle-defaults": "^1.5.1",
+    "@metaplex-foundation/umi-web3js-adapters": "^1.5.1",
```
apedrob: Concern: We should stick with a hardcoded version (without `^`)

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
mervin-link: I think so because these are useful for CLI/debug flows.

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
mervin-link: I’d keep this local for now. These checks are specific to deployToken/metaplex params. If another operation needs the same token metadata validation later, we can extract it then.

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,193 @@
+import { TokenStandard, createV1, mplTokenMetadata } from '@metaplex-foundation/mpl-token-metadata'
```
aelmanaa: Eagerly importing `@metaplex-foundation/*` libraries at the top level of this file introduces a very large bundle footprint (~7.8 MB and ~19 transitive dependencies) that propagates to the barrel exports of `@chainlink/ccip-sdk/cct/solana`. As a result, non-deploy users (calling `setPool` or `createLookupTable`) are forced to load these dependencies.

To address this, consider:
- **Option A:** Hand-rolling the Metaplex Create-V1 instruction (discriminator 42, `CreateArgs::V1`, borsh `AssetData`), guarded by a dev-only dependency and golden test asserting byte-for-byte serialization equality with Umi to prevent layout drift.
- **Option B:** Using dynamic imports (`await import(...)`) inside `createMetadataInstructions` (when `withMetaplex: true` is executed) and exporting `DeployToken` lazily from the barrel files.

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,193 @@
...
+  SolanaOperation,
+} from '../../operation.ts'
+import { submit } from '../../submit.ts'
+import { validatePublicKey } from '../../validate.ts'
+
+type BaseDeployTokenParams = {
+  /** Mint decimals. Must be an integer between 0 and 255. */
+  decimals: number
+  /** Token program that owns the mint: classic SPL Token or Token-2022. */
+  tokenProgram: 'spl-token' | 'token-2022'
```
aelmanaa: Consider defaulting `tokenProgram` to `'spl-token'` to avoid requiring this parameter for the most common use path.

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,193 @@
...
+export type GenerateDeployTokenParams = SolanaGenerateParams<DeployTokenParams>
+
+/** Unsigned token deploy transaction plus the created mint address. */
+export type GenerateDeployTokenResult = UnsignedSolanaTx & { tokenAddress: string }
+
+/** Parameters for executing Solana token deploy. */
+export type ExecuteDeployTokenParams = SolanaExecuteParams<DeployTokenParams>
+
+/** Result of executing Solana token deploy. */
+export type ExecuteDeployTokenResult = TransactionHash & { tokenAddress: string }
```
aelmanaa: Consider returning `metadataAddress` in the result when `withMetaplex: true`. Since the metadata PDA is created during deployment, the caller needs its address to update/verify metadata later. This can be represented in the discriminated union type of the result.

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,193 @@
...
+  /** Builds the unsigned Solana mint creation instruction set. */
+  protected async buildUnsigned(
+    chain: SolanaChain,
+    params: GenerateDeployTokenParams,
+  ): Promise<GenerateDeployTokenResult> {
+    const payer = new PublicKey(params.payer)
+    const tokenProgram =
+      params.tokenProgram === 'token-2022' ? TOKEN_2022_PROGRAM_ID : TOKEN_PROGRAM_ID
+
+    const seed = `mint_${Date.now()}_${Math.random().toString(36).slice(2, 8)}`
```
aelmanaa: The seed is currently hardcoded as `mint_${Date.now()}_...`. Consider accepting an optional caller-supplied/derived seed parameter (and defaulting to this random string if omitted). This makes deployments reproducible and resumable (re-signing the same transaction bytes produces the same mint address), which is crucial for offline/multisig signer graphs.

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,193 @@
...
+      SystemProgram.createAccountWithSeed({
+        fromPubkey: payer,
+        newAccountPubkey: mint,
+        basePubkey: payer,
+        seed,
+        lamports,
+        space: mintSpace,
+        programId: tokenProgram,
+      }),
+      createInitializeMintInstruction(mint, params.decimals, payer, payer, tokenProgram),
```
aelmanaa: Consider using `createInitializeMint2Instruction` instead of the legacy `createInitializeMintInstruction` (which requires the rent sysvar).

Additionally, regarding authorities and supply:
1. **Nullable freeze authority**: Expose an optional freeze authority parameter that defaults to the payer but can be `null`, allowing deployers to renounce freeze authority at creation time for better centralization hygiene.
2. **Configurable mint authority**: Allow a configurable `mintAuthority` (defaulting to the payer) so that self-served pool initialization and later transfers of authority to the pool PDA function correctly.
3. **Pre-mint / Initial Supply**: Expose an optional, explicit `preMint`/`preMintRecipient` parameter to mint initial supply within the same deployment transaction (using idempotent ATA creation + `mintTo`). If a non-payer `mintAuthority` is specified, ensure it is surfaced as a required signer for transaction assembly.

aelmanaa (commented):
Feedback and suggestions for the Solana CCT deployToken implementation.

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,193 @@
...
+      throw new CCTParamsInvalidError(this.name, 'decimals', 'must be an integer between 0 and 255')
+    }
+    if (!['spl-token', 'token-2022'].includes(params.tokenProgram)) {
+      throw new CCTParamsInvalidError(this.name, 'tokenProgram', 'must be spl-token or token-2022')
+    }
+    if (typeof params.withMetaplex !== 'boolean') {
+      throw new CCTParamsInvalidError(this.name, 'withMetaplex', 'must be a boolean')
+    }
+    if (!params.withMetaplex) return
+    if (!params.name || params.name.length > 32) {
```
aelmanaa: The metadata name and symbol constraints should be validated using their UTF-8 byte lengths instead of JS `string.length` (which measures UTF-16 code units). For multi-byte characters (e.g. emojis, accents), a string might pass the JS length check but revert on-chain because its UTF-8 encoding exceeds the Metaplex byte limits (32 bytes for name, 10 bytes for symbol).

Consider checking the byte size using a helper like `new TextEncoder().encode(val).length`.

aelmanaa (commented):
Add comment regarding UTF-8 byte length validation.

ccip-sdk/src/cct/solana/token/operations/deploy-token.ts
mervin-link: I opted for Option B. Replaced with dynamic import inside `deploy-token` op and `SolanaTokenManager`.

apedrob (approved):

PR #296, opened 2026-07-14
feat(cct-sdk): Add deploy token pool solana op

mervin-link:
## What

- [DAPP-10507](https://smartcontract-it.atlassian.net/browse/DAPP-10507)
- Add Solana CCT SDK support for deploying/initializing token pools
- Expose `generateUnsignedDeployTokenPool` and `deployTokenPool` methods on `SolanaTokenManager`
- Support optional allowlist configuration during token pool initialization

## Why

 - Solana CCT users need an SDK operation to initialize token pools after deploying a mint and before registering the pool with the TokenAdminRegistry

[DAPP-10507]: https://smartcontract-it.atlassian.net/browse/DAPP-10507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,133 @@
...
+export type GenerateDeployTokenPoolParams = SolanaGenerateParams<DeployTokenPoolParams>
+
+/** Unsigned Solana token pool deploy result. */
+export type GenerateDeployTokenPoolResult = UnsignedSolanaTx
+
+/** Parameters for executing Solana token pool deploy. */
+export type ExecuteDeployTokenPoolParams = SolanaExecuteParams<DeployTokenPoolParams>
+
+/** Result of executing Solana token pool deploy. */
+export type ExecuteDeployTokenPoolResult = TransactionHash
```
aelmanaa: Currently, `deployTokenPool` and `generateUnsignedDeployTokenPool` do not return the derived pool state PDA (`poolAddress`). Although the pool state PDA is deterministic and can be computed from the inputs by downstream operations, it is not returned in the result. In addition, the PDA derivation helper is not exported on the public SDK surface.

Consider returning `poolAddress` (mirroring `deployToken`'s behavior) or exporting the PDA derivation helper on the public surface to improve discoverability.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -155,6 +163,44 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
...
+   * ```ts
+   * const cct = SolanaTokenManager.fromChain(chain)
+   * await cct.deployTokenPool({
+   *   tokenAddress: mint,
+   *   poolProgramAddress: poolProgram,
+   *   wallet,
+   * })
+   * ```
+   */
+  deployTokenPool(opts: ExecuteDeployTokenPoolParams): Promise<ExecuteDeployTokenPoolResult> {
```
aelmanaa, line 210: There is an undocumented multi-step setup journey here. The `initialize` instruction assumes the global config PDA is already initialized, requires the `authority` to be the upgrade or mint authority, and does not create the pool signer PDA's ATA (which is deferred and handled separately).

Without documentation, callers might experience opaque reverts/errors. A potential improvement is to add a `@remarks` section outlining the required ordering and prerequisites (e.g., initializing global config, setting up authorities, and creating the pool signer ATA before configuring/registering the pool).

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,133 @@
...
+} from '../../programs/token-pool.ts'
+import { submit } from '../../submit.ts'
+import { validatePublicKey } from '../../validate.ts'
+
+/** Parameters for initializing a Solana token pool, optionally with an allowlist. */
+type DeployTokenPoolParams = {
+  /** Token mint address this pool manages. */
+  tokenAddress: string
+  /** Token pool program address, e.g. BurnMint or LockRelease token pool program. */
+  poolProgramAddress: string
```
aelmanaa: The input `poolProgramAddress` is a free-form string. Because there are only two shared canonical CCT pool programs (`'burn-mint'` and `'lock-release'`), accepting a free-form string makes the API less discoverable and prone to errors.

Consider replacing `poolProgramAddress: string` with a closed selector like `poolType: 'burn-mint' | 'lock-release'` mapping to their respective canonical on-chain addresses. This would structurally prevent invalid program addresses and make the options discoverable out-of-the-box.

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,133 @@
...
+  /** Generate, sign, simulate, send, and confirm with wallet.publicKey as payer. */
+  override async execute(
+    chain: SolanaChain,
+    params: ExecuteDeployTokenPoolParams,
+  ): Promise<ExecuteDeployTokenPoolResult> {
+    const { wallet, computeUnits, ...rest } = params
+    if (!isWallet(wallet)) throw new CCIPWalletInvalidError(wallet)
+
+    const payer = wallet.publicKey.toBase58()
+    if (params.authority && !new PublicKey(params.authority).equals(wallet.publicKey)) {
```
aelmanaa: If `params.authority` is passed but is not a valid PublicKey string, the `new PublicKey(params.authority)` constructor will throw a raw web3.js error before reaching the validation method `validate()` on line 130. Please ensure `validatePublicKey` is called on the authority string before performing the equality check.

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,133 @@
...
+          config: deriveTokenPoolGlobalConfigPda(poolProgram),
+        })
+        .instruction(),
+    ]
+
+    const allowlist = (opts.allowlist ?? []).map((a) => new PublicKey(a))
+    if (allowlist.length) {
+      instructions.push(
+        await program.methods
+          .configureAllowList(allowlist, true)
```
aelmanaa, line 108: Currently, configuring the allowlist is only supported during token pool deployment. It would be highly beneficial to support post-deploy lifecycles by adding standalone operations like `configureAllowList` and `removeFromAllowList` to match the on-chain program capabilities.

Also, it would be helpful to document the `enabled = true` parameter to clarify that seeding the list automatically toggles enforcement.

aelmanaa (commented):
Detailed review feedback on the Solana CCT deployTokenPool PR.

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
mervin-link: Added comment for enable param and I'll do a follow up PRs for `configureAllowList` and `removeFromAllowList` ops. This will be added on milestone 3.

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
mervin-link: @aelmanaa So this means we are not supporting custom pools? Are we doing the same for create lookup table and append lookup table when deriving CCIP addresses? In there, we also have a free-form string of `poolProgramAddress`

ccip-sdk/src/cct/solana/index.ts
mervin-link:
```@remarks
This only sends the pool initialize instruction. The signer must be allowed to initialize the pool. This does not create the pool signer PDA's associated token account.
```

@aelmanaa I added this remarks above to both `generateUnsignedDeployTokenPool` and `deployTokenPool` clarifying that this only builds/sends the pool initialize instruction, the authority/signer must be allowed to initialize the pool, and the pool signer ATA is not created here.

 I didn’t call out global config initialization as a caller prerequisite since we now only target canonical CCIP pool programs, so my understanding is that config is CCIP-managed. Let me know if I’m wrong.

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,146 @@
...
+  poolAddress: string
+}
+
+/** Parameters for executing Solana token pool deploy. */
+export type ExecuteDeployTokenPoolParams = SolanaExecuteParams<DeployTokenPoolParams>
+
+/** Result of executing Solana token pool deploy plus the derived pool state PDA. */
+export type ExecuteDeployTokenPoolResult = TransactionHash & {
+  poolAddress: string
+}
```
apedrob, line 59: I don't know if it make sense for Solana, but I've defined in `evm/operation.ts` this type can will reuse across different deployments
```
export type DeployResult = TransactionResult & { contractAddress: string }
```

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.validate(generateParams)
+
+    const authority = params.authority ? new PublicKey(params.authority) : undefined
+    if (authority && !authority.equals(wallet.publicKey)) {
+      throw new CCTParamsInvalidError(
+        this.name,
+        'authority',
+        'deployTokenPool requires authority to be the executing wallet. Use generateUnsignedDeployTokenPool for vault-owned pools and have the vault sign/execute it.',
+      )
+    }
```
apedrob, line 145: Should this be in the `validate`

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,146 @@
...
+  override async execute(
+    chain: SolanaChain,
+    params: ExecuteDeployTokenPoolParams,
+  ): Promise<ExecuteDeployTokenPoolResult> {
+    const { wallet, computeUnits, ...rest } = params
+    if (!isWallet(wallet)) throw new CCIPWalletInvalidError(wallet)
+
+    const payer = wallet.publicKey.toBase58()
+    const generateParams: GenerateDeployTokenPoolParams = { ...rest, payer }
+    this.validate(generateParams)
```
apedrob, line 135: Confirm if you can call `this.generate` function from SolanaOperation instead of `buildUnsigned` to inherit its inner `this.validate`

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
mervin-link: Currently, deploys in Solana are different. Deploy token returns two additonal params which are `tokenAddress` and `metadataAddress` while deploy token pool returns `poolAddress` only.

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
mervin-link: That’s what I tried before, but I need to verify that `params.authority` exists and is a valid `PublicKey` before checking it against wallet. Using `this.generate` would repeat that validation, so I used `buildUnsigned` instead.

ccip-sdk/src/cct/solana/index.ts
aelmanaa: Correct: with `poolType` limiting to the canonical CCIP pool programs, global config is CCIP-managed, so it is not a caller prerequisite. The `@remarks` cover the rest. Resolving — I will open a separate doc-only follow-up to add the ordered deploy→ATA→setPool journey + a DAPP-10522 cross-link.

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
aelmanaa: Good call — I checked, and CCIP-Solana **does** support custom pool programs (`set_pool` validates the ALT's derived PDAs structurally, not against an allow-list — `cctp-token-pool` is a real third program). So we should NOT collapse to closed `poolType` everywhere. The nuance: `deployTokenPool` can only `initialize()` programs whose IDL we ship, so `poolType` is correct *here* and this PR is good as-is. But `createLookupTable`/`appendToLookupTable` are program-agnostic (they just derive PDAs), so they should keep custom support — I will open a follow-up to give them a `{ poolType } | { poolProgramAddress }` union (adds the discoverable shortcut, keeps today's free-form path). Not a blocker here.

ccip-sdk/src/cct/solana/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,151 @@
...
+import {
+  validateAuthorityMatchesWallet,
+  validatePoolType,
+  validatePublicKey,
+  validatePublicKeys,
+} from '../../validate.ts'
+
+/** Parameters for initializing a Solana token pool, optionally with an allowlist. */
+type DeployTokenPoolParams = {
+  /** Token mint address this pool manages. */
```
aelmanaa, line 31: Suggest making the `poolType`-only scoping explicit in the doc so the boundary is intentional-on-the-record, not an accidental omission.

`@remarks` Targets only the canonical CCIP pool programs selected by `poolType` (burn-mint, lock-release). Deploying/initializing a **custom** pool program is intentionally not supported here: this op initializes the pool through the SDK's bundled program IDL, and a custom program may declare a different `initialize` instruction or pool-state PDA layout. This is a deploy-op scoping choice, not a protocol limit. CCIP-Solana accepts custom pool programs, and the registry/lookup-table ops stay program-agnostic. We can revisit if there is concrete demand for custom-pool deployment.

aelmanaa (approved):

apedrob (approved):

PR #297, opened 2026-07-14
feat(cct-sdk): Deploy CrossChainToken + contract artifacts

apedrob:
What

- [DAPP-10491](https://smartcontract-it.atlassian.net/browse/DAPP-10491)
- Refactor the EVM transaction pipeline to return rich TransactionResults (response + receipt)
- Add deployToken and unsigned deployment support to the EVM CCT SDK
- Update tests to cover transaction results and contract deployment

## Why

- Support EVM token deployments in the CCT SDK while exposing richer transaction results required for deployment operations

[DAPP-10491]: https://smartcontract-it.atlassian.net/browse/DAPP-10491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/evm/token/bytecode.ts
apedrob, line 1: This will be removed by introducing `@chainlink/contracts-ccip` artifacts in the next follow-up PR

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,69 @@
...
+import { interfaces } from '../../../../evm/const.ts'
+import type { EVMChain } from '../../../../evm/index.ts'
+import type { UnsignedEVMTx } from '../../../../evm/types.ts'
+import { ChainFamily } from '../../../../networks.ts'
+import { CCTTxFailedError } from '../../../errors.ts'
+import type { DeployResult } from '../../../operation.ts'
+import { EVMOperation } from '../../operation.ts'
+import { submit } from '../../submit.ts'
+import { validateNonEmptyString, validateUint256, validateUint8 } from '../../validate.ts'
+import { BURN_MINT_ERC677_BYTECODE } from '../bytecode.ts'
```
aelmanaa: The deployment operation currently ships init-code for `BurnMintERC677Token` v1.5.1 (bytecode pinned to `release/contracts-ccip-1.5.1`). However, the current v2.0 CCT token is `CrossChainToken` v2.0.0, which features a different self-serve constructor. The supported CCT deploy tokens are `CrossChainToken` and `CrossChainPoolToken` — these two self-serve v2.0 tokens carry `getCCIPAdmin()`, role admin, and pre-mint, which register through the v2.0 flow. `BurnMintERC677Token` is the legacy CCIP token and is not a supported CCT deploy target. Please replace this with `CrossChainToken v2.0.0` (and `CrossChainPoolToken` for the pool-token variant) so it doesn't target a different token generation than the sibling registry/pool operations on the same facade.

aelmanaa (commented):
Here is the CCT deploy EVM token PR review.

ccip-sdk/src/cct/evm/token/operations/deploy-token.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  return {
+    provider: {} as never,
+    logger: { debug() {}, info() {}, warn() {}, error() {} },
+    nextNonce: async () => 0,
+    rollbackNonce: () => {},
+  } as unknown as EVMChain
+}
+
+/** Fake ethers Signer whose deployment receipt carries `contractAddress`. */
+function fakeSigner(opts: {
```
aelmanaa: The inline object type in `fakeSigner`'s parameter is split across multiple lines, causing a prettier violation that breaks `npm run check`. Please inline this definition, e.g.:
`function fakeSigner(opts: { contractAddress?: string | null; waitError?: Error }) {`

ccip-sdk/src/cct/operation.ts
```diff
@@ -7,14 +7,17 @@
 
 import type { ChainTransaction } from '../types.ts'
 
-/** Confirmed on-chain hash returned by a successful CCT write. */
-export type TransactionHash = Pick<ChainTransaction, 'hash'>
+/** Result of a successful CCT write: the confirmed on-chain tx hash. */
+export type TransactionResult = Pick<ChainTransaction, 'hash'>
+
+/** Result of a successful contract-deployment write: the tx hash plus the deployed address. */
+export type DeployResult = TransactionResult & { address: string }
```
aelmanaa: Using the bare name `address` in the result here collides with `address` in the context of the handoff to the very next call, `setPool`, where `tokenAddress` means the token and `address` means the TokenAdminRegistry/Router. Consider returning `tokenAddress` instead of `address` (e.g. `DeployTokenResult`) to make the vocabulary specific and avoid variable collision when destructuring.

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -31,8 +37,14 @@ export abstract class EVMOperation<P extends { sender?: string }> extends Operat
     return unsigned
   }
 
-  /** {@link generate}, then sign and submit via {@link submit}; returns once confirmed. */
-  async execute(chain: EVMChain, params: P & { wallet: unknown }): Promise<TransactionHash> {
-    return submit(chain, params.wallet, await this.generate(chain, params), this.name)
+  /** {@link generate}, then sign and submit; returns the confirmed tx hash. */
+  async execute(chain: EVMChain, params: P & { wallet: unknown }): Promise<TransactionResult> {
```
aelmanaa: Currently, because `EVMOperation` hardcodes `TransactionResult` as the `Result` generic parameter, any operation that needs to return more than a transaction hash (such as the deployed contract address in `DeployToken`) is forced to fully override the `execute` method, duplicating the submit-and-map wiring. A potential solution to avoid this boilerplate duplication would be to parameterize the `Result` type in the base class and introduce a protected hook for mapping the receipt (such as `resultFromReceipt`), keeping the execution flow unified in the base class.

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,69 @@
...
+      chain,
+      params.wallet,
+      await this.generate(chain, params),
+      this.name,
+    )
+    if (!receipt.contractAddress)
+      throw new CCTTxFailedError(this.name, 'deployment produced no contract address', {
+        context: { txHash: response.hash },
+      })
+    return { hash: response.hash, address: receipt.contractAddress }
```
aelmanaa: Since `BurnMintERC677Token` deploys with zero supply and doesn't grant mint/burn roles to anyone by default, the deployer must explicitly call `grantMintAndBurnRoles` before minting works. Let's document this post-deploy behavior on the facade `deployToken` comments to avoid access-control revert confusion for developers.

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,69 @@
...
+    params: DeployTokenParams & { wallet: unknown },
+  ): Promise<DeployResult> {
+    const { response, receipt } = await submit(
+      chain,
+      params.wallet,
+      await this.generate(chain, params),
+      this.name,
+    )
+    if (!receipt.contractAddress)
+      throw new CCTTxFailedError(this.name, 'deployment produced no contract address', {
```
aelmanaa, line 135: The generic `CCT_TX_FAILED` recovery hint ('Ensure the caller holds the required role...') doesn't apply to deployment failures, where the receipt simply reported no `contractAddress` (often an RPC quirk). We should make the failure reason self-contained so developers don't chase a nonexistent role issue.

ccip-sdk/src/cct/operation.ts
aelmanaa: The result currently omits the block-explorer verification handle (such as the ABI-encoded constructor arguments for Etherscan-style source verification), which is the data a later verify step needs.

While verification is EVM-specific (and Solana/Aptos token deploys have no custom bytecode to verify), a deploy tool emitting raw init-code without returning a verification handle makes it harder for downstream CLI or tooling steps to verify the contract later. If this is a deliberate design decision, it would be good to confirm and document it, as adding a verification handle field later would change the exported `DeployResult` type shape.

aelmanaa (commented):
Adding comment regarding contract verification handle.

ccip-sdk/src/cct/operation.ts
aelmanaa: Here are reference patterns for defining the token-specific vocabulary (`tokenAddress`) and the block-explorer `verification` handle (which is built at deploy time with no extra network calls):

```ts
export type DeployTokenResult = TransactionHash & {
 tokenAddress: string
 verification: DeployVerification
}

export interface DeployVerification { contract: string; encodedConstructorArgs: string }

export function buildDeployVerification(contract: string, deployData: string, bytecode: string) {
 // ctor args = the deploy calldata with the creation bytecode prefix removed
 return { contract, encodedConstructorArgs: `0x${deployData.slice(bytecode.length)}` }
}
```

Inside the operation (e.g. via `resultFromReceipt`), the handle can be populated directly from the transaction data:
```ts
return {
 ...hash,
 tokenAddress: receipt.contractAddress,
 verification: buildDeployVerification('CrossChainToken', data, CROSS_CHAIN_TOKEN_BYTECODE),
}
```

aelmanaa (commented):
Adding reference patterns and code examples for the suggested architecture and verification handle changes.

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
apedrob: Agreed we should move to CrossChainToken 2.0.0. For this PR I'm intentionally using the `BurnMintERC677Token` bytecode/ABI already existing in ccip-sdk; the 2.0 swap lands in the follow-up PR that introduces `contracts-ccip` dep (new bytecode + ABI + constructor). Also considering adding the factory erc20 (1.6) deploy op too, wdyt?

ccip-sdk/src/cct/operation.ts
apedrob: I'd prefer keeping `address` here. `DeployResult` is a generic deployment result that won't always be a token (pools, lockbox, hook), so it can be misleading

`setPool` won't collided given isn't a deployment, but can fix the ambiguity by renaming its param to `contractAddress` as we discussed before

ccip-sdk/src/cct/evm/operation.ts
apedrob: Talked this through with @mervin-link earlier [here](https://github.com/smartcontractkit/ccip-tools-ts/pull/291#discussion_r3579729034). The hook is essentially a cast, and it forces a signature onto every op to satisfy the base even when they don't need it. We could introduce a `DeployOperation` subclass to have it, but overriding execute falls out naturally across the EVM/Solana ops anyway. Suggest we push it down for now; happy to revisit if a second deploy-style op makes the pattern earn its keep

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
apedrob: Will add a note on `deployToken`

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
apedrob: You right, I'll pass a recovery referring to receipt/rpc

ccip-sdk/src/cct/operation.ts
apedrob: Deliberate for now, will document on `DeployResult`. Adding a verification field later is non-breaking, deferring until there is verify step to consume it

ccip-sdk/src/cct/operation.ts
```diff
@@ -7,14 +7,21 @@
...
-export type TransactionHash = Pick<ChainTransaction, 'hash'>
+/** Result of a successful CCT write: the confirmed on-chain tx hash. */
+export type TransactionResult = Pick<ChainTransaction, 'hash'>
+
+/**
+ * Result of a successful deployment write: the tx hash plus the deployed contract address (token, pool, etc.)
+ * Note: No block-explorer verification handle yet; it's recoverable from the init-code,
+ * so it can be added later without a breaking change.
+ */
+export type DeployResult = TransactionResult & { contractAddress: string }
```
mervin-link: Would it be better to keep this EVM-only and avoid generic return types? Some deploys return different fields, like Solana deployToken, which returns two extra fields with different names.

ccip-sdk/src/cct/evm/index.ts
```diff
@@ -91,7 +93,7 @@ export class EVMTokenManager extends TokenManager<typeof ChainFamily.EVM> {
    * })
    * ```
    */
-  setPool(opts: SetPoolParams & { wallet: unknown }): Promise<TransactionHash> {
+  setPool(opts: SetPoolParams & { wallet: unknown }): Promise<TransactionResult> {
```
mervin-link: Should we add an explicit param types and export it? It can be use by end users to type their own variables and will be easier to extend later.

ccip-sdk/src/cct/operation.ts
apedrob: Yes! I'll move it to the `EVMOperation`

ccip-sdk/src/cct/evm/index.ts
apedrob: I'll create the chain-agnostic `ExecuteParams` with does ` & wallet` + specific `EVMExecuteParams` (same pattern as we have in solana)

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
aelmanaa: Resolved by `DAPP-10544` (`feature/DAPP-10544-deploy-cross-chain-token`). The legacy `BurnMintERC677Token` bytecode is deleted and the op now deploys the supported CCT token types: `CrossChainToken` v2.0.0 (default) and `FactoryBurnMintERC20` v1.5.1/v1.6.2, dispatched on `version`. Constructor args verified byte-order-exact against canonical `BaseERC20.sol` / `CrossChainToken.sol` and both `FactoryBurnMintERC20` ABIs. Closing this thread.

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
aelmanaa: Resolved by `DAPP-10544`. The `execute` override now throws `CCTTxFailedError` with the specific message "deployment produced no contract address" and `context.txHash` — self-contained, no misleading role hint. The `CCTTxFailedError` doc was also updated to cover "mines without the expected effect (e.g. a deployment that produced no contract address)." Closing this thread.

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
aelmanaa: is this done?
or anyway let's revisit once you have the PR with CrossChainToken.sol

ccip-sdk/src/cct/operation.ts
aelmanaa: let s defer verification to another PR , I can help on that

aelmanaa: let's defer verification to a separate PR

apedrob: Created [DAPP-10572](https://smartcontract-it.atlassian.net/browse/DAPP-10572) to address it after

ccip-sdk/src/cct/evm/token/operations/deploy-token.ts
```diff
@@ -0,0 +1,140 @@
...
+  symbol: string
+  decimals: number
+  /** Max supply cap; `0n` means unlimited. */
+  maxSupply: bigint
+  /** Amount minted at deploy; defaults to `0n`. Must be `<= maxSupply` when capped. */
+  preMint?: bigint
+  /** Receives ownership; a valid address. */
+  owner: string
+  /** Recipient of `preMint`; required when `preMint > 0`, must be unset otherwise. */
+  preMintRecipient?: string
```
apedrob, line 40: Decision: Previously I set the default to `owner`; however, that could lead to unintended minting to the wrong address. I have decided to maintain the existing logic in the contract code

ccip-sdk/src/cct/evm/token/version.ts
```diff
@@ -0,0 +1,74 @@
...
+    iface: new Interface(CROSS_CHAIN_TOKEN_V2_0_0_ABI),
+    bytecode: CROSS_CHAIN_TOKEN_V2_0_0_BYTECODE,
+  },
+}
+
+/**
+ * Returns the cached deploy artifact for `version`.
+ * @throws {@link CCTContractVersionUnsupportedError} if `version` has no vendored deploy bytecode
+ */
+export function tokenArtifact(version: TokenVersion): TokenArtifact {
```
mervin-link, line 70: Do you think it's better to call this `getTokenArtifact()` since it's a function and not an object? Same with `getTokenAbi()`

ccip-sdk/src/cct/evm/index.ts
```diff
@@ -113,11 +116,61 @@ export class EVMTokenManager extends TokenManager<typeof ChainFamily.EVM> {
...
+   */
+  deployToken(opts: EVMExecuteParams<DeployTokenParams>): Promise<DeployResult> {
+    return this.#deployToken.execute(this.chain, opts)
+  }
 }
 
 export * from '../errors.ts'
 export type { SetPoolParams } from './token-admin-registry/operations/set-pool.ts'
-export type { TransactionHash } from '../operation.ts'
+export type { DeployTokenParams } from './token/operations/deploy-token.ts'
```
mervin-link, line 174: Nit: Can create a barrel on each operations folder and export the types and class.

mervin-link (approved):

aelmanaa (approved):
Approved

aelmanaa (approved):

PR #298, opened 2026-07-14
chore(cct-sdk): Gate subtree from public releases

apedrob:
## What

- Drop the ./cct/evm export and exclude src/cct from files + the release build, so cct/ stays on main but never ships. Adds build:dev for local npm link use (dist/cct emitted locally, never packed)

## Why

- CCT will be merged before audit; this keep it out of the public npm package until it's ready, without blocking development


ccip-sdk/tsconfig.build.json
```diff
@@ -11,6 +11,7 @@
     "node_modules",
     "**/*.test.*",
     "**/__tests__",
-    "**/__mocks__"
+    "**/__mocks__",
+    "./src/cct"
```
mervin-link, line 15: Do we still need this? Since we do have our own holding branch now.

apedrob: Given this is targeting holding branch, I believe we should have it to unblock its merge to main before auditing

mervin-link (approved):

PR #300, opened 2026-07-15
feat(cct-sdk): Add create token account (ATA) solana op

mervin-link:
## What

- [DAPP-10522](https://smartcontract-it.atlassian.net/browse/DAPP-10522)
- Add Solana create token account (ATA) operation to the CCT SDK
- Support creating ATAs for both wallet and PDA addresses

## Why

 - Enables users to create associated token accounts through the SDK as part of Solana token workflows

[DAPP-10522]: https://smartcontract-it.atlassian.net/browse/DAPP-10522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token/operations/create-token-account.ts
```diff
@@ -0,0 +1,112 @@
...
+} from '../../operation.ts'
+import { submit } from '../../submit.ts'
+import { validatePublicKey } from '../../validate.ts'
+
+/** Parameters for deriving and creating a Solana associated token account. */
+type CreateTokenAccountParams = {
+  /** SPL token mint address for the associated token account. */
+  tokenAddress: string
+  /** Wallet or PDA owner address for the associated token account. */
+  ownerAddress: string
```
aelmanaa, line 23: The headline use case for this op is the pool-signer PDA ATA — the step between `deployTokenPool` and `setPool` that fails with `AccountNotInitialized` if skipped. To use this op for that, the caller has to pass the pool signer PDA as `ownerAddress`, but the SDK gives them no public way to derive it: `deriveTokenPoolSignerPda` is internal (not re-exported from the facade), and `deployTokenPool` returns only `poolAddress` (the state PDA), not the signer PDA.

`TOKEN_POOL_PROGRAMS` / `TokenPoolType` are exported, so a caller can resolve the program address — but they'd still have to hand-roll `findProgramAddressSync(['ccip_tokenpool_signer', mint], poolProgram)` with a hardcoded seed to get the owner. That's exactly the kind of thing this op exists to abstract away, and we can't expect users to know the internal seed string.

This blocks merge for me — the op exists to close the deploy→setPool journey gap, but without a public way to derive the pool signer PDA it can't actually be pointed at the pool reserve. A user reading the facade `@remarks` ("pass the pool signer PDA as `ownerAddress`") is left with no way to obtain that value.

Three options, roughly best-to-worst DX:
- a pool-aware convenience (`createPoolTokenAccount({ poolType, tokenAddress })`) that derives the owner internally, mirroring the bs58 command's shape;
- export `deriveTokenPoolSignerPda` so the caller can derive it themselves;
- return the signer PDA from `deployTokenPool` so it threads straight into `ownerAddress`.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -123,6 +134,43 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
     return new DeployToken().execute(this.chain, opts)
   }
 
+  /**
+   * Builds an unsigned idempotent associated token account create instruction.
+   *
```
aelmanaa, line 139: Good that this calls out the pool PDA case — but it stops at "pass the pool signer PDA as `ownerAddress`" without telling the caller *how to obtain* that PDA, and without sequencing where this step sits. Right now a dev reading this still has to know to derive `findProgramAddressSync(['ccip_tokenpool_signer', mint], poolProgram)` out-of-band.

Worth expanding this `@remarks` to: (1) state the ordered journey — `deployTokenPool` → **createTokenAccount** → `setPool`; (2) note this op is idempotent / safe to re-run; (3) once the derivation is exposed (see the comment on `create-token-account.ts`), point at it. A `@see deployTokenPool` / `@see setPool` both ways would close the loop.

The sibling `createTokenAccount` method below has no `@remarks` at all — worth mirroring whatever lands here.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -123,6 +134,43 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
...
+  /**
+   * Creates an associated token account for a wallet or PDA owner.
+   *
+   * @example
+   * ```ts
+   * const cct = SolanaTokenManager.fromChain(chain)
+   * await cct.createTokenAccount({ wallet, tokenAddress: mint, ownerAddress: owner })
+   * ```
+   */
+  createTokenAccount(
```
aelmanaa, line 189: Missing `@throws` on both this and `generateUnsignedCreateTokenAccount` above. The op throws `CCTParamsInvalidError` for a missing mint or a mint owned by a non-token program (`create-token-account.ts:68`, `:75`), and `CCIPWalletInvalidError` for a non-wallet in the execute path. Worth listing them so they show up in autocomplete / hover.

ccip-sdk/src/cct/solana/token/operations/create-token-account.ts
```diff
@@ -0,0 +1,112 @@
...
+    const mint = new PublicKey(params.tokenAddress)
+    const owner = new PublicKey(params.ownerAddress)
+    const mintAccount = await chain.connection.getAccountInfo(mint)
+
+    if (!mintAccount) {
+      throw new CCTParamsInvalidError(this.name, 'tokenAddress', 'token mint not found')
+    }
+
+    if (
+      !mintAccount.owner.equals(TOKEN_PROGRAM_ID) &&
```
aelmanaa: This inline mint-fetch + owner-check duplicates what `resolveATA` in `solana/utils.ts` already does — and with weaker error types. `resolveATA` throws `CCIPTokenMintNotFoundError` / `CCIPTokenMintInvalidError` (specific), while this throws generic `CCTParamsInvalidError` for the same conditions. The ATA derivation on line 78 is also identical to what `resolveATA` returns.

Worth reusing `resolveATA` here — `buildUnsigned` collapses to roughly:

  const { ata, tokenProgram } = await resolveATA(chain.connection, mint, owner)
  // ... build createAssociatedTokenAccountIdempotentInstruction(payer, ata, owner, mint, tokenProgram)

One thing worth deciding at the same time: `resolveATA` does detection + ATA derivation in one call, but not every caller needs the ATA. `deriveCcipLookupTableAddresses` in `programs/alt.ts:78` already calls `resolveATA` and discards the ATA — it only wants `tokenProgram`. That's a sign the detection should be its own primitive:

  resolveTokenProgram(connection, mint): Promise<PublicKey>   // fetch + validate owner, return the program
  resolveATA(connection, mint, owner): Promise<ResolvedATA>   // calls resolveTokenProgram, then derives the ATA

Then this op uses `resolveATA` (needs both), `alt.ts` uses `resolveTokenProgram` (needs only the program, stops deriving an ATA it throws away), and future CCT ops that need the token program have the right primitive ready. Both live in `solana/utils.ts` — the CCT layer already imports from there.

Not a blocker for this PR on its own, but the duplication is here now and `alt.ts` already shows the shape mismatch. Worth landing as a small follow-up that extracts `resolveTokenProgram`, refactors `resolveATA` to call it, and retrofits both call sites.

ccip-sdk/src/cct/solana/token/operations/create-token-account.ts
mervin-link: I addressed this by exporting `deriveTokenPoolSignerPda` and returning `poolSignerAddress` from `deployTokenPool`, so callers can create the pool reserve ATA without reimplementing the PDA derivation.

I did not add `createPoolTokenAccount` yet because `createTokenAccount` already supports ATAs for any owner. Callers can use a user wallet address when a recipient needs an ATA for newly minted tokens. This keeps the API generic for now.

We can revisit adding `createPoolTokenAccount` later to improve DX. For now, this matches expert mode. We don't mint tokens there, so callers explicitly create any required recipient ATA before minting tokens.

ccip-sdk/src/cct/solana/token/operations/create-token-account.ts
mervin-link: Refactored `solana/utils.ts` to extract `resolveTokenMint` and `resolveTokenProgram`.

aelmanaa (approved):

apedrob (approved):

PR #301, opened 2026-07-15
feat(sdk): add pre-send destination-liquidity preflight that mandatorily simulates the destination pool's releaseOrMint for token transfers

aelmanaa:
This pull request introduces significant improvements to the gas estimation documentation and SDK, particularly around pre-send validation and error handling for CCIP transfers. The main focus is on clarifying and enhancing how `estimateReceiveExecution` not only estimates gas but also validates transfer executability on the destination chain, helping developers avoid stuck transfers. Additionally, new error types and lower-level simulation primitives are documented and implemented, and the guides are reorganized to highlight these features.

**Documentation and Guide Improvements:**

* Added a new guide, `pre-send-validation.mdx`, detailing how `estimateReceiveExecution` validates that a transfer can execute on the destination chain before sending, including error handling and lower-level simulation primitives.
* Updated `gas-estimation.mdx` to clarify that on EVM chains, `estimateReceiveExecution` simulates `ccipReceive` and returns gas units, while on Solana it estimates compute units; also documented that the method validates executability and throws errors if the transfer would be stuck. [[1]](diffhunk://#diff-eb81ca56ad4fd7d11d8af6b40947b28deb83ee69e85fd93ef9f90f5f810b14abL14-R25) [[2]](diffhunk://#diff-eb81ca56ad4fd7d11d8af6b40947b28deb83ee69e85fd93ef9f90f5f810b14abL114-R119)
* Improved tables and descriptions in `gas-estimation.mdx` for consistency and clarity, including error and method reference tables. [[1]](diffhunk://#diff-eb81ca56ad4fd7d11d8af6b40947b28deb83ee69e85fd93ef9f90f5f810b14abL74-R78) [[2]](diffhunk://#diff-eb81ca56ad4fd7d11d8af6b40947b28deb83ee69e85fd93ef9f90f5f810b14abL171-R176) [[3]](diffhunk://#diff-eb81ca56ad4fd7d11d8af6b40947b28deb83ee69e85fd93ef9f90f5f810b14abL182-R187)
* Added the new "Pre-Send Validation" guide to the SDK sidebar for better discoverability.

**SDK and Error Handling Enhancements:**

* Introduced new error codes and exports for destination pool preflight checks: `CCIPDestExecutionRevertError` and `CCIPDestSimulationUnavailableError`, improving error reporting when a transfer would fail on the destination. [[1]](diffhunk://#diff-24660f25cbb898cbec77132845abd18021806f8cbe14382a6da74fea50106c06R110-R113) [[2]](diffhunk://#diff-e9714197bce9c681f30d9d4425e06062e1d9b8c7c39bc31c37ceecb9cacbb236R180-R182)
* Extended the `Chain` class with a new `simulateLockOrBurn` method, allowing simulation of source pool logic to obtain accurate data for destination pool validation.
* Updated the validation logic in the SDK to allow more flexible message input and removed redundant liquidity checks from the chain-level preflight, as these are now handled by destination simulation. [[1]](diffhunk://#diff-807a4ca9474665a0a92dd38d0c7b802fa323550a99ef0c98a2ae754be2cd8838L1370) [[2]](diffhunk://#diff-807a4ca9474665a0a92dd38d0c7b802fa323550a99ef0c98a2ae754be2cd8838L1382-R1381) [[3]](diffhunk://#diff-807a4ca9474665a0a92dd38d0c7b802fa323550a99ef0c98a2ae754be2cd8838L1392-L1404)


ccip-sdk/src/evm/simulate.ts
```diff
@@ -0,0 +1,344 @@
...
+/**
+ * ERC165 interface id of `IPoolV2` (v2.0 pools, 2-arg `releaseOrMint`/`lockOrBurn`).
+ * Matches `type(IPoolV2).interfaceId` in chainlink-ccip.
+ */
+export const IPOOL_V2_INTERFACE_ID = '0x940a1542'
+/**
+ * ERC165 interface id of v1 pools: `Pool.CCIP_POOL_V1 = bytes4(keccak256("CCIP_POOL_V1"))`
+ * (1-arg `releaseOrMint`/`lockOrBurn`; v1.5.0 through v1.6.x, incl. the oUSDT pools).
+ */
+export const CCIP_POOL_V1_INTERFACE_ID = '0xaff2afbf'
```
andrevmatos: Prefer code as doc:
```suggestion
export const CCIP_POOL_V1_INTERFACE_ID = id('CCIP_POOL_V1').substring(0, 10) as '0xaff2afbf'
```

ccip-sdk/src/evm/dest-liquidity.fork.test.ts
andrevmatos, line 1: I don't think you need a fork.test for these; fork tests are mostly for read-write tests; your simulations here are read-only, a plain integration test would do

ccip-sdk/src/evm/simulate.ts
```diff
@@ -0,0 +1,344 @@
...
+      senderBytes.length > 0 && senderBytes.length < 32
+        ? zeroPadValue(senderBytes, 32)
+        : hexlify(senderBytes),
+    remoteChainSelector: input.remoteChainSelector,
+    receiver: input.receiver,
+    sourceDenominatedAmount: input.sourceDenominatedAmount,
+    localToken: input.localToken,
+    sourcePoolAddress: hexlify(getDataBytes(input.sourcePoolAddress)),
+    sourcePoolData: hexlify(getDataBytes(input.sourcePoolData ?? '0x')),
+    offchainTokenData: hexlify(getDataBytes(input.offchainTokenData ?? '0x')),
```
andrevmatos, line 220: 🔴 **Pre-send simulation false-blocks every proof/attestation-carrying pool (USDC CCTP v1, Lombard v1).**

`offchainTokenData` is always empty pre-send, but several pools consume it unconditionally:

- `LombardTokenPool.releaseOrMint` (v1 path) does `abi.decode(releaseOrMintIn.offchainTokenData, (bytes, bytes))` then `mailbox.deliverAndHandle(rawPayload, proof)` — empty bytes revert at decode.
- `USDCTokenPool` v1.5.1 requires the CCTP attestation from `offchainTokenData` for `receiveMessage`; `USDCTokenPoolProxy` on the `CCTP_V1_FLAG` path likewise.

These are exactly the lanes where the data only exists *after* the send (Circle attestation, bridge proof) — a "fundamentally inconclusive pre-send" case per this PR's own error model, yet today every such transfer is guaranteed to revert the simulation and block the send. Suggest: detect pools that require offchain token data (typeAndVersion probe, or classify the specific decode/attestation revert on an empty-data attempt) and report the check as **unavailable** pre-send; accept real `offchainTokenData` in the manual-exec path where it exists.

ccip-sdk/src/evm/simulate.ts
```diff
@@ -0,0 +1,344 @@
...
+  pool: string,
+  interfaceId: string,
+): Promise<boolean> {
+  try {
+    const result = await provider.call({
+      to: pool,
+      data: interfaces.TokenPool_v2_0.encodeFunctionData('supportsInterface', [interfaceId]),
+    })
+    return !!interfaces.TokenPool_v2_0.decodeFunctionResult('supportsInterface', result)[0]
+  } catch {
```
andrevmatos: 🔴 **`probeSupportsInterface` converts RPC/transport failures into ERC165 `false`.**

`catch { return false }` means a flaky destination RPC makes both probes return `false`, producing a hard, non-transient `CCIPContractTypeInvalidError` ("incompatible pool") for what was a transient transport failure — and `checkExecute` rethrows that as a definitive block. Only a *successful* `false` (or a successful undecodable response) should mean "unsupported"; rethrow transport errors so they surface as `CCIPDestSimulationUnavailableError`.

ccip-sdk/src/gas.ts
```diff
@@ -261,15 +261,49 @@ export async function estimateReceiveExecution({
...
+            onRamp,
+            destChainSelector: dest.network.chainSelector,
+            token: tokenAmount.token,
+            amount: tokenAmount.amount,
+            originalSender: message.sender,
+            receiver: message.receiver,
+            finality: message.finality,
+          })
+          forCheck = {
+            amount: tokenAmount.amount,
```
andrevmatos: 🔴 **Simulation uses the gross amount; v2 `lockOrBurn` returns a post-fee `destTokenAmount` that the OnRamp writes into the message.**

Confirmed in `OnRamp.sol`: for IPoolV2 pools, `(poolReturnData, destTokenAmount) = IPoolV2(...).lockOrBurn(...)` and the emitted `TokenTransferV1.amount = destTokenAmount` — *not* the input amount. Fee-charging pools (e.g. Lombard v2, which forwards the full amount to the verifier and returns the post-fee amount) make these diverge.

Here `forCheck.amount = tokenAmount.amount` (gross), and `simulateLockOrBurn` (`simulate.ts`) decodes only `LockOrBurnOutV1`, discarding the second return value. The destination check therefore runs against a different (larger) amount than the real message → false blocks on inbound rate limit/liquidity, plus a wrong `destinationAmount`. Return `destTokenAmount` from `simulateLockOrBurn` / `Chain.simulateLockOrBurn` and propagate it as `forCheck.amount`.

ccip-sdk/src/gas.ts
```diff
@@ -261,15 +261,49 @@ export async function estimateReceiveExecution({
...
+            sourceTokenAddress: tokenAmount.token,
+            sourcePoolAddress,
+            destTokenAddress: destTokenAmount.token,
+            extraData: destPoolData,
+          }
+        } catch (err) {
```
andrevmatos, line 361: 🔴 **Every source `lockOrBurn` revert is swallowed and replaced with fabricated decimals data.**

A revert from the source pool simulation is a definitive on-chain verdict under this PR's own error model (sender allowlist, outbound rate limit, finality gate, pool policy hook), yet the catch logs at debug level and silently falls back to the decimals default, letting the send proceed. At minimum branch on `getErrorData(err)`: revert-with-data → surface a typed source-side error; only a transport failure (or a pool shape known not to encode real state) should fall back. The actual send tx would still revert on-chain, so this is preflight quality rather than funds safety — but it defeats the purpose of running the check.

ccip-sdk/src/gas.ts
```diff
@@ -261,15 +261,49 @@ export async function estimateReceiveExecution({
...
-    (message.tokenAmounts ?? []).map(async (tokenAmount) =>
-      getDestTokenAmount({ source, dest, onRamp, tokenAmount }),
-    ),
+  const resolvedTokenAmounts = await Promise.all(
+    (message.tokenAmounts ?? []).map(async (tokenAmount) => {
+      const destTokenAmount = await getDestTokenAmount({ source, dest, onRamp, tokenAmount })
+      // with the source chain at hand, obtain the pool-reported destPoolData from the source
+      // pool's lockOrBurn, so the destination releaseOrMint simulation consumes the same
+      // sourcePoolData a real transfer would carry (paired with the source-denominated amount)
+      let forCheck: NonNullable<EstimateMessageInput['tokenAmounts']>[number] = destTokenAmount
```
andrevmatos, line 312: 🟠 **Post-send/manual-exec path drops authoritative `sourcePoolAddress` and `extraData`.**

When this is called post-send (CLI `manual-exec --estimate-gas-limit` passes `request.message`), token amounts arrive as `{sourcePoolAddress, destTokenAddress, extraData, amount}`. `forCheck` collapses to `getDestTokenAmount`'s `{token, amount}`, so `checkExecute` re-derives the source pool from `remotePools.at(-1)` — the *currently* configured pool, not necessarily the pool that actually sent the message (false pass after pool migration; false block after pool removal) — and fabricates decimals `sourcePoolData` even though the real `extraData` was in hand. In post-send mode, trust the emitted fields; never reconstruct or re-simulate.

ccip-sdk/src/evm/simulate.ts
```diff
@@ -0,0 +1,344 @@
...
+    localToken: input.localToken,
+  }
+
+  let poolInterface: PoolInterfaceVersion, fragment: string, args: unknown[]
+  if (await probeSupportsInterface(provider, pool, IPOOL_V2_INTERFACE_ID)) {
+    // IPoolV2.lockOrBurn(LockOrBurnInV1, bytes4 requestedFinalityConfig, bytes tokenArgs)
+    // returns (LockOrBurnOutV1, uint256 destTokenAmount); tokenArgs is empty for base pools
+    poolInterface = 'IPoolV2'
+    fragment = 'lockOrBurn((bytes,uint64,address,uint256,address),bytes4,bytes)'
+    args = [lockOrBurnIn, toBeHex(encodeFinality(finality ?? 0n), 4), '0x']
```
andrevmatos: 🟠 **v3 `tokenArgs` is hardcoded to `0x`.**

`OnRamp.sol` passes the user's `tokenArgs` through to `IPoolV2.lockOrBurn`. Pools and hooks that branch on it (AdvancedPoolHooks, Lombard) are simulated with a different call than the real message. Thread `tokenArgs` from the message's extraArgs into `simulateLockOrBurn`.

ccip-sdk/src/evm/index.ts
```diff
@@ -2384,6 +2396,241 @@ export class EVMChain extends Chain<typeof ChainFamily.EVM> {
...
+      try {
+        await simulateReleaseOrMint({
+          provider: this.provider,
+          pool: tokenPool,
+          offRamp,
+          finality: message.finality,
+          input: {
+            originalSender: message.sender,
+            remoteChainSelector: message.sourceChainSelector,
+            receiver: message.receiver,
```
andrevmatos: 🟠 **v3 `tokenReceiver` is not honored.**

`OnRamp.sol` resolves `lockOrBurnIn.receiver` from `tokenReceiver` when set, and the destination `releaseOrMint` releases to it. Both simulations use `message.receiver` instead. For v3 messages with a distinct `tokenReceiver`, the simulated release goes to a different address than the real one (and receiver-dependent pool hooks validate the wrong account). Thread `tokenReceiver` from extraArgs, falling back to `receiver`.

ccip-sdk/src/chain.ts
```diff
@@ -1389,19 +1388,6 @@ export abstract class Chain<F extends ChainFamily = ChainFamily> {
       if (!token || token.match(/^(0x)?0*$/i)) continue
       registry ??= await this.getTokenAdminRegistryFor(offRamp)
       const { tokenPool } = await this.getRegistryTokenConfig(registry, token)
-      const { typeAndVersion, lockBox } = await this.getTokenPoolConfig(tokenPool!)
```
andrevmatos, line 1392: 🔴 **Removing the generic LockRelease liquidity check regresses non-EVM destinations.**

The base `checkExecute` loses the `typeAndVersion.includes('LockRelease')` balance check. `EVMChain` replaces it (better) via simulation, but the override does not call `super`, and non-EVM dest chains now have only the inbound rate-limit read — no liquidity preflight at all (Solana implements `getTokenPoolConfig` with the pool program's typeAndVersion; the new docs confirm Solana runs no pool validation). Restore the generic check in the base class and layer the EVM simulation on top.

ccip-sdk/src/evm/index.ts
```diff
@@ -2384,6 +2396,241 @@ export class EVMChain extends Chain<typeof ChainFamily.EVM> {
...
+      } else {
+        let remotePools: readonly string[] | undefined
+        try {
+          remotePools = (await this.getTokenPoolRemote(tokenPool, message.sourceChainSelector))
+            .remotePools
+        } catch (err) {
+          const revertData = getErrorData(err)
+          // No revert data => an RPC/transport failure; executability is undetermined => transient
+          // so the caller retries.
+          if (!revertData)
```
andrevmatos, line 2759: 🟡 **Typed SDK errors are misclassified as transient transport failures.**

`getTokenPoolRemote` throws `CCIPTokenPoolChainConfigNotFoundError` (a typed SDK error with no revert data) when the lane config is missing — a definitive on-chain condition — but this branch treats anything without revert data as `CCIPDestSimulationUnavailableError` ("retry later"). Branch on SDK error types before inspecting revert bytes: typed config errors should hard-block.

ccip-sdk/src/evm/simulate.ts
```diff
@@ -0,0 +1,344 @@
...
+// generic dest-execution error as transient (retryable); it never selects an error type or the
+// block decision. Names are matched against the shared SDK ABI bundle, so every entry is a
+// standard chainlink-ccip pool / RateLimiter error. External errors not in the bundle (oUSDT
+// lockbox, xERC20 bridge limits) read as non-transient — the caller can parse `context.revert`
+// (the raw encoded revert) with `EVMChain.parse` to refine.
+const TRANSIENT_REVERT_NAMES = new Set([
+  'InsufficientLiquidity',
+  'InsufficientBalance',
+  'ERC20InsufficientBalance',
+  'TokenMaxCapacityExceeded',
```
andrevmatos: 🟡 **`TokenMaxCapacityExceeded` is not transient.**

It means `amount > bucket capacity` — the configured maximum never refills, so the transfer can never succeed at that amount. Only `TokenRateLimitReached` (and the aggregate-value `...RateLimitReached`) refill over time. Move it out of the transient set; conversely the refillable aggregate/value rate-limit errors could be added.

ccip-sdk/src/evm/const.ts
```diff
@@ -1,16 +1,25 @@
 import { parseAbi } from 'abitype'
 import { type EventFragment, AbiCoder, Interface } from 'ethers'
 
+import AdvancedPoolHooks_2_0_ABI from './abi/AdvancedPoolHooks_2_0.ts'
```
andrevmatos: 🟡 **Bundle size:** thirteen complete generated ABIs (~19k lines) are imported into this broadly-imported module solely to decode error selectors. Since `interfaces` is a single eager object, none of it tree-shakes. Consider generating minimal error-only fragments, or a lazy error registry. Fast follow is fine.

ccip-api-ref/docs-sdk/guides/pre-send-validation.mdx
```diff
@@ -0,0 +1,125 @@
...
+4. Returns the `ccipReceive` gas limit (unchanged from a gas-only estimate).
+
+**Whenever a message carries token transfers this destination simulation is not optional** — a transfer whose destination pool cannot release or mint would strand on the destination, so the check always runs. If the `releaseOrMint` simulation reverts, the message would not execute on the destination, so the send is **always blocked** — that verdict does not depend on whether the SDK recognizes the specific revert. Recognizing the revert only selects a more specific error; an unrecognized revert still blocks (as `CCIPDestExecutionRevertError`, carrying the decoded name or raw selector). The one case that is neither a pass nor a block is a transport/RPC failure that prevents the simulation from running at all — reported as the transient `CCIPDestSimulationUnavailableError` so you can retry. The returned gas number is unchanged.
```
andrevmatos: 🟠 **Docs overclaim coverage.** "Simulates the destination execution" overstates it: the pool-direct call cannot cover OffRamp source-lane enablement/allowed-OnRamp, CCV quorum and inbound implementations, CCV-managed minting (CCTPThroughCCV pools no-op the pool leg — the real mint is verifier-side), receiver `ccipReceive`, or state races. It also doesn't mention the guaranteed false-block on attestation/proof pools (USDC CCTP v1, Lombard v1). Please reframe as a point-in-time pool-leg check with an explicit "not covered" list.

andrevmatos (changes requested):
## Review: request changes

Destination `releaseOrMint` simulation is the right primitive for liquidity, lockbox, rate-limit, RMN, mint-authority, and pool-hook checks. But the message this PR reconstructs is not always the message the OnRamp will emit — it can both falsely approve and falsely block sends. Findings below are inline where the code is in the diff; three cross-cutting items are here in the body.

### 🔴 Default CLI sends abort on `CCIPDestSimulationUnavailableError` (`ccip-cli/src/commands/send.ts:382`, not in diff)

Gas estimation runs by default (`estimateGasLimit == null || > -100`), and the catch rethrows everything except `CCIPMethodUnsupportedError`. `CCIPDestSimulationUnavailableError` is explicitly the *inconclusive* case (no dest RPC / transport failure) — per the stated design goal ("error only when a definitive condition is met"), the default policy should be **warn and continue**, reserving failure for definitive on-chain verdicts or explicit `--estimate-gas-limit` / `--only-estimate`.

### 🟡 Pre-existing, worth fixing in this PR or as follow-up: receiver finality lookup (`ccip-sdk/src/evm/index.ts:2687`, not in diff)

The finality check queries the receiver contract directly and defaults *permissively* (`{finalityDepth: 1, finalitySafe: true}`) on any error, while the OffRamp defaults non-v2 receivers to wait-for-finality and composes receiver/pool/lane defaults. `OffRamp.getCCVsForMessage(bytes)` exists precisely for offchain use and is the canonical resolver — prefer it. (This block pre-exists the PR, so not a regression; flagged because this PR is the right pipeline to route it through.)

### Missing destination gates (the original ask's dest-only v2 conditions)

1. **OffRamp lane gates are never checked.** The pool-direct simulation bypasses the OffRamp, so nothing validates `OffRamp.getSourceChainConfig(sourceSelector)`: `isEnabled`, RMN verification state, and whether the sending OnRamp is in the OffRamp's allowed `onRamps` list. A message on a disabled lane or from a non-allowed ramp passes this preflight and strands. This is a cheap view call with the already-resolved `offRamp`, and a definitive on-chain condition → should block.
2. **Required CCVs / executors are not honored.** Pool `getRequiredCCVs`, receiver-required CCVs, and lane defaults are never compared against the user's `ccvs` extraArgs. `OffRamp.getCCVsForMessage` returns required/optional/threshold for offchain use — but it takes the *encoded* message, so it only becomes trustworthy after the message-construction fixes inline below (destTokenAmount, tokenArgs, tokenReceiver, preserved emitted fields).

### Suggested landing order

1. The "blocks valid sends" class: offchain-data pools false-block (`simulate.ts:159`), probe error swallow (`simulate.ts:141`), CLI policy (`send.ts:382`).
2. The "simulates the wrong message" class: `destTokenAmount` (`gas.ts:283`), preserved emitted fields (`gas.ts:270`), `tokenArgs`/`tokenReceiver` (`simulate.ts:322`, `index.ts:2531`).
3. Error-model fixes: source-revert swallow (`gas.ts:289`), non-EVM liquidity fallback (`chain.ts`), typed-error misclassification (`index.ts:2471`), transient-set fix (`simulate.ts:48`).
4. Missing gates: OffRamp source-config check, then CCV check via `getCCVsForMessage`; docs reframe (`pre-send-validation.mdx`) last.

### Required check ownership (proposed)

| Stage | Checks that belong there |
|---|---|
| Source chain | Router/lane enabled, RMN, sender allowlist, source pool registration/remote config, outbound rate/capacity, finality support, source pool required CCVs, fee/balance/approval |
| Destination chain | OffRamp source enabled + allowed OnRamp, receiver finality, receiver/pool/lane CCV requirements, pool compatibility, source-pool wiring, inbound rate/capacity, lockbox liquidity, mint authority, RMN, pool hooks |
| Post-send/manual exec | Exact emitted amount, source pool data, token receiver, offchain token data, verifier results, receiver execution |
| Inconclusive pre-send | Future attestations/proofs, executor liveness, concurrent state changes, future receiver state |

Validation note: this is a static trace against the diff (`61cd778..HEAD`) and the contracts in `chainlink-ccip` (`OnRamp.sol`, `OffRamp.sol`, `LombardTokenPool.sol`, `CCTPThroughCCVTokenPool.sol`). The current tests encode two of the assumptions flagged here (gross amount = message amount; decimals fallback after failed source simulation), so green CI does not cover these paths — please add cases for fee-charging pools, CCTP-v1/Lombard-v1 lanes, and post-send/manual-exec inputs.

andrevmatos (dismissed):
Good job!

andrevmatos (approved):

PR #302, opened 2026-07-16
feat(cct-sdk): Add create token multisig solana op

mervin-link:
## What

- [DAPP-10523](https://smartcontract-it.atlassian.net/browse/DAPP-10523)
- Add Solana create token multisig operation to the CCT SDK
- Add two defaults signers which are pool signer PDA and mint authority

## Why
- Enables users to create token multisig through the SDK as part of Solana token workflows

## Notes

- We add the pool’s PDA as a signer threshold times, so its programmatic signature satisfies the multisig approval requirement and lets the pool mint automatically.
- We also require at least threshold independent signers, so they can separately satisfy the multisig approval requirement.

[DAPP-10523]: https://smartcontract-it.atlassian.net/browse/DAPP-10523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
```diff
@@ -0,0 +1,216 @@
...
+/** Unsigned token multisig transaction plus the created multisig address. */
+export type GenerateCreateTokenMultisigResult = UnsignedSolanaTx & { multisigAddress: string }
+
+/** Parameters for executing Solana token multisig creation. */
+export type ExecuteCreateTokenMultisigParams = SolanaExecuteParams<CreateTokenMultisigParams>
+
+/** Result of executing Solana token multisig creation. */
+export type ExecuteCreateTokenMultisigResult = TransactionHash & { multisigAddress: string }
+
+function dedupePublicKeys(signers: PublicKey[]) {
```
aelmanaa, line 64: This dedup is the root of the authority-model problem below. The burn-mint pool's `transfer_mint_authority_to_multisig` checks the multisig's signer list and counts how many times the pool signer PDA appears (`signer_count`), then requires `signer_count >= threshold` so the pool can mint autonomously on CCIP delivery with no human co-signer. It also requires `signer_count <= n - threshold`, which implies `n >= 2 * threshold`.

Because `dedupePublicKeys` collapses the pool signer PDA to a single slot, `signer_count` is always 1. That means:
- At `threshold = 1`: valid (`signer_count(1) >= m(1)`). This is the shape the pool's own tests use (`pools_test.go`: `CreateMultisig(..., 1, [admin, poolSigner])`).
- At `threshold >= 2`: the pool rejects the multisig with `PoolSignerNotInMultisig`. If it were force-installed as mint authority, every inbound CCIP mint would fail because the pool alone can never reach threshold.

SPL's `InitializeMultisig` accepts duplicate signer keys — repeating the pool PDA `threshold` times is exactly how the pool intends the multisig to be built. The org's `ccip-solana-bs58-generator` forwards the caller's signer list verbatim to `createInitializeMultisigInstruction` (no dedup), so an operator can pass `[poolPDA, poolPDA, admin, gov]` with `threshold = 2` and get a valid pool-autonomous multisig.

Suggested fix: insert `deriveTokenPoolSignerPda(poolProgram, tokenMint)` exactly `threshold` times, dedup only the non-pool signers (mint authority + `additionalSigners`), and require `n >= 2 * threshold` so governance retains an independent path to threshold. If keeping the single-pool-slot shape, constrain `threshold` to 1 and update the examples/tests/docs accordingly — but the better fix is to build the pool-autonomous shape.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
```diff
@@ -0,0 +1,216 @@
...
+    const tokenMint = new PublicKey(opts.tokenAddress)
+    const poolProgram = resolveTokenPoolProgram(opts.poolType)
+    const mintAccount = await chain.connection.getAccountInfo(tokenMint)
+    validateMintAccount(this.name, mintAccount)
+
+    const tokenProgram = mintAccount.owner
+    const authority = getMintAuthority(this.name, tokenMint, mintAccount, tokenProgram)
+
+    const signers = dedupePublicKeys([
+      deriveTokenPoolSignerPda(poolProgram, tokenMint),
```
aelmanaa: This is where the pool signer PDA gets added exactly once. Combined with the dedup above, the multisig this op produces cannot satisfy the pool's `signer_count >= threshold` rule for any `threshold >= 2`.

For the pool-autonomous construction, this should emit the pool PDA `threshold` times, e.g.:
```ts
const poolSigner = deriveTokenPoolSignerPda(poolProgram, tokenMint)
const otherSigners = dedupePublicKeys([
  authority,
  ...(opts.additionalSigners ?? []).map((s) => new PublicKey(s)),
])
const signers = [
  ...Array.from({ length: opts.threshold }, () => poolSigner),
  ...otherSigners,
]
```
Then validate `n >= 2 * threshold` (enough outside signers that governance can also reach threshold) and `threshold >= 1`.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
```diff
@@ -0,0 +1,216 @@
...
+  const seen = new Set<string>()
+  return signers.filter((signer) => {
+    const address = signer.toBase58()
+    if (seen.has(address)) return false
+    seen.add(address)
+    return true
+  })
+}
+
+function validateMultisigSigners(operation: string, signers: PublicKey[], threshold: number) {
```
aelmanaa: This validator only checks `1 <= n <= 11` and `1 <= threshold <= n`. The on-chain `validate_multisig_config` in the burn-mint pool enforces a stricter set of rules that this doesn't mirror:
- `n >= 2` (at least two signers)
- `threshold >= 1`
- `n >= threshold`
- `signer_count >= threshold` (pool PDA appears enough times to satisfy threshold alone)
- `signer_count <= n - threshold` (pool doesn't occupy so many slots that outside signers can't reach threshold)

As a result, params that pass this validator can fail on-chain at `transfer_mint_authority_to_multisig`, or worse, produce a mint authority that silently can't mint. Suggested fix: mirror `validate_multisig_config` exactly — count pool-signer occurrences in the assembled signer list and assert all five rules, emitting `CCTParamsInvalidError`s that name the failing rule. This validator and the signer-set construction should be designed together so the construction guarantees `signer_count = threshold` and the validator proves the outside-signer headroom.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  /** Validates public keys, threshold, and optional seed before mint/account RPCs. */
+  protected validate(params: GenerateCreateTokenMultisigParams): void {
+    validatePublicKey(this.name, 'tokenAddress', params.tokenAddress)
+    validatePoolType(this.name, 'poolType', params.poolType)
+    validatePublicKey(this.name, 'payer', params.payer)
+    if (params.additionalSigners !== undefined) {
+      validatePublicKeys(this.name, 'additionalSigners', params.additionalSigners)
+    }
+    validateInteger(this.name, 'threshold', params.threshold)
```
aelmanaa: `threshold` is validated as an integer with no bounds here, so `threshold: 0` (or a very large int) passes `validate()` and is only caught later in `validateMultisigSigners`. The early validator gives false assurance. Suggested fix: pass `min = 1` here — `validateInteger(this.name, 'threshold', params.threshold, 1)` — and leave the upper bound (which depends on the signer count) to the signer validator.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
```diff
@@ -0,0 +1,216 @@
...
+import {
+  validateInteger,
+  validateNonEmptyString,
+  validatePoolType,
+  validatePublicKey,
+  validatePublicKeys,
+  validateTokenProgram,
+} from '../../validate.ts'
+
+export const SOLANA_MULTISIG_MAX_SIGNERS = 11
```
aelmanaa, line 32: `@solana/spl-token` already exports `MAX_SIGNERS` (which is 11). Hardcoding it here means this constant can drift if SPL ever changes the limit. Suggested fix: `import { MAX_SIGNERS } from '@solana/spl-token'` and re-export or alias it as `SOLANA_MULTISIG_MAX_SIGNERS`.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
```diff
@@ -0,0 +1,216 @@
...
+    const seedInput = concat([toUtf8Bytes(seedMaterial), tokenMint.toBuffer()])
+    const seedHash = sha256(seedInput)
+    const seed = seedHash.slice(2, 34)
+
+    const multisig = await PublicKey.createWithSeed(authority, seed, tokenProgram)
+    const lamports = await chain.connection.getMinimumBalanceForRentExemption(MULTISIG_SIZE)
+    const createIx = SystemProgram.createAccountWithSeed({
+      fromPubkey: payer,
+      newAccountPubkey: multisig,
+      basePubkey: authority,
```
aelmanaa, line 183: `createAccountWithSeed` uses `basePubkey = authority` (the mint authority). When `payer !== authority` — which is exactly the vault flow `generateUnsignedCreateTokenMultisig` exists for — Solana requires the base (mint authority) to co-sign the transaction in addition to the payer. Nothing in the JSDoc or the facade `@example` mentions this, so an integrator building the unsigned tx for a vault will produce a transaction that fails with "missing required signature". Suggested fix: add an `@remarks` to `generateUnsignedCreateTokenMultisig` (and the facade method) noting that the unsigned tx must be signed by both the payer and the mint authority (the seed base); in the vault flow the vault (mint authority) must sign.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    const generateParams: GenerateCreateTokenMultisigParams = {
+      ...rest,
+      payer: wallet.publicKey.toBase58(),
+    }
+    this.validate(generateParams)
+
+    const tokenMint = new PublicKey(generateParams.tokenAddress)
+    const mintAccount = await chain.connection.getAccountInfo(tokenMint)
+    validateMintAccount(this.name, mintAccount)
```
aelmanaa: `execute` fetches the mint account and reads the mint authority here for the wallet==authority guard, then `buildUnsigned` fetches the same mint account and unpacks the authority again — two RPC round-trips for the same mint per call. Suggested fix: fetch the mint account/authority once and thread it into `buildUnsigned` (e.g. an internal overload taking the resolved `mintAccount`/`authority`), or move the wallet-equality guard into `buildUnsigned`.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -171,6 +177,50 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
...
+   * The default signers are pool signer PDA and mint authority.
+   *
+   * @example
+   * ```ts
+   * const cct = SolanaTokenManager.fromChain(chain)
+   * const unsigned = await cct.generateUnsignedCreateTokenMultisig({
+   *   payer,
+   *   tokenAddress: mint,
+   *   poolType: 'burn-mint',
+   *   threshold: 2,
```
aelmanaa, line 219: This example uses `threshold: 2`. With the current construction (pool signer PDA added once, then deduped), the resulting multisig has `n = 2` (pool + authority, plus `admin` from `additionalSigners` → `n = 3`), `threshold = 2`, and `signer_count = 1`. The burn-mint pool requires `signer_count >= threshold`, so this multisig would be rejected at `transfer_mint_authority_to_multisig` with `PoolSignerNotInMultisig`. Until the construction is fixed to repeat the pool PDA `threshold` times, the example should use `threshold: 1` (the only value the current op can produce correctly), or be updated to match the fixed construction.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -171,6 +177,50 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
...
+   * Wallet pays fees and must match the mint authority.
+   *
+   * @example
+   * ```ts
+   * const cct = SolanaTokenManager.fromChain(chain)
+   * const { hash, multisigAddress } = await cct.createTokenMultisig({
+   *   wallet,
+   *   tokenAddress: mint,
+   *   poolType: 'burn-mint',
+   *   threshold: 2,
```
aelmanaa, line 248: Same issue as the `generateUnsignedCreateTokenMultisig` example: `threshold: 2` with no `additionalSigners` gives `n = 2`, `threshold = 2`, `signer_count = 1`, which the pool rejects (`PoolSignerNotInMultisig`). This is also the signed path where the wallet must be the mint authority, so there's no vault co-signing — the multisig is created but cannot be used as the pool's mint authority. Until the construction is fixed, this example should use `threshold: 1`.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    },
+  } as unknown as SolanaChain
+}
+
+function generate(opts = {}) {
+  return SolanaTokenManager.fromChain(stubChain()).generateUnsignedCreateTokenMultisig({
+    tokenAddress: MINT,
+    poolType: 'burn-mint',
+    threshold: 2,
+    payer: PAYER,
```
aelmanaa, line 62: The test default is `threshold: 2`, and the stubbed chain never asserts the pool's multisig invariant — so the tests pass while producing a multisig the burn-mint pool would reject. This is why the defect is silent at the unit-test level. Once the construction is fixed (pool PDA repeated `threshold` times), add an assertion that the pool signer PDA appears `threshold` times in `initIx.keys`, and add a test case for `threshold = 1` matching the canonical `pools_test.go` shape (`[admin, poolSigner]`). If the fix is deferred and `threshold` is pinned to 1, update this default to `threshold: 1`.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
mervin-link: I can't find `MAX_SIGNERS` in `@solana/spl-token`. I also checked the `ccip-solana-bs58-generator` and it's using a hardcoded `11`.

ccip-sdk/src/cct/solana/token-pool/operations/create-token-multisig.ts
mervin-link: Added a third `MintContext` param for `buildUnsigned`.

aelmanaa (approved):

apedrob (approved):

PR #303, opened 2026-07-16
feat(cct-sdk): Source chainlink/contracts-ccip artifacts

apedrob:
## What

- Source token + pool abis and bytecodes under `cct/evm/artifacts`, regenerable from `@chainlink/contracts-ccip` (build-time devDep)

## Why

- One versioned source of truth instead of hand-copied abis; foundation for token/pool deployments and operations (no runtime change)


ccip-sdk/src/cct/evm/artifacts/abi/V1_5_0/burn-mint-token-pool-and-proxy.ts
```diff
@@ -0,0 +1,1055 @@
+export default [
```
mervin-link, line 1: May I know how are these being generated? Do we have a script? Maybe it's better to write a script for this then add a type like:

ABI
```
import type { Abi } from 'viem'

export const BURN_MINT_TOKEN_POOL_AND_PROXY_1_5_0 = [] as const satisfies Abi
```

Bytecode
```
export const ADVANCED_POOL_HOOKS_2_0_0_BYTECODE = '0x123...'
```

or make it as `default`.

apedrob: Yes, there's already a script and it's an established repo-wide convention:`generate.cjs` at the root files inline 
`// generate: … // generate:end` blocks and generated via `npm run generate` / `generate:lint`.

For this file, you can check this generation block starting on line 2

PR #306, opened 2026-07-21
feat(cct-sdk): Add get token pool state config solana op

mervin-link:
## What

- [DAPP-10574](https://smartcontract-it.atlassian.net/browse/DAPP-10574)
- Add `getTokenPoolState` to the Solana CCT token manager for Burn/Mint and Lock/Release pools
- Return serialized pool configuration and pool-type-specific state fields

## Why

- Enable SDK consumers to inspect canonical Solana token pool configuration without manually fetching and decoding the state account

## Note

- Custom pool program address does not recognize `lock-release` type so `lock-release` fields will not be included.

[DAPP-10574]: https://smartcontract-it.atlassian.net/browse/DAPP-10574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,122 @@
...
+> {
+  /** Reads and serializes the token pool configuration account. */
+  async query(
+    chain: SolanaChain,
+    params: GetTokenPoolStateParams,
+  ): Promise<GetTokenPoolStateResult> {
+    validatePoolType('getTokenPoolState', 'poolType', params.poolType)
+    validatePublicKey('getTokenPoolState', 'tokenAddress', params.tokenAddress)
+
+    const programId = resolveTokenPoolProgram(params.poolType)
```
aelmanaa: A read decodes an account by layout, so it only needs the shared BaseConfig borsh schema — not the bundled IDL or a canonical program identity (that constraint is specific to deployTokenPool, which calls .initialize() through a shipped IDL). Right now poolType locks this to the two canonical programs, but CCIP-Solana does support custom pool programs (set_pool validates the ALT-derived PDAs structurally, not against an allow-list), and core getTokenPoolConfig is already program-agnostic (it reads account.owner to discover the program). So this read is weaker than the core code it parallels.

Suggest a PoolProgramRef union so both are supported:

  type PoolProgramRef =
    | { poolType: TokenPoolType; poolProgramAddress?: never }
    | { poolProgramAddress: string; poolType?: never }

poolType stays the recommended default (discoverable, typo-proof); poolProgramAddress is the escape hatch for custom pools. The caller supplies exactly one — the `?: never` enforces that at compile time. Both arms funnel through the existing resolvePoolProgram.

This op would then follow the ALT ops (createLookupTable/appendToLookupTable), not deployTokenPool — a read is program-agnostic like they are, and simpler since it needs no writableIndexes pairing. Not a blocker for this PR, but worth aligning before the PoolProgramRef follow-up lands.

ccip-sdk/src/cct/solana/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,122 @@
...
+    const mint = new PublicKey(params.tokenAddress)
+    const state = deriveTokenPoolConfigPda(programId, mint)
+    const account = await chain.connection.getAccountInfo(state)
+    if (!account) throw new CCIPTokenPoolStateNotFoundError(state.toBase58())
```
aelmanaa: The error passes state.toBase58() — a PDA the caller never supplied, so they can't map it back to the tokenAddress/poolType they called with. If they hit this, they're left guessing which input was wrong.

Worth passing the mint and poolType (or poolProgramAddress, once the union lands) as context and referencing the mint in the message, with a hint that the pool may not be initialized yet (→ deployTokenPool) or the poolType is wrong for this mint. Something like:

  throw new CCIPTokenPoolStateNotFoundError(state.toBase58(), {
    mint: tokenAddress,
    poolType: params.poolType,
  })

…with the message naming the mint, not just the PDA.

ccip-sdk/src/cct/solana/programs/token-pool.ts
```diff
@@ -36,6 +41,11 @@ export function createTokenPoolProgram(
   return new Program(TOKEN_POOL_IDL, poolProgram, simulationProvider(chain, payer))
 }
 
+/** Decodes a canonical token pool state account. */
+export function decodeTokenPoolState(data: Buffer): { version: number; config: TokenPoolConfig } {
+  return tokenPoolCoder.accounts.decode('state', data)
```
aelmanaa: accounts.decode throws a raw Error("Invalid account discriminator") when the fetched account isn't the expected type, and this function doesn't wrap it — so a decode failure (wrong account at the PDA, or a future layout change) surfaces as an untyped Error with an anchor-internal message. The not-found path above is already a typed CCIPError; the decode path isn't.

Worth try/catching here (or at the call site) and rethrowing a typed error with { tokenPool, mint, poolType } context, so callers get the same error taxonomy they do everywhere else in the SDK.

ccip-sdk/src/cct/solana/programs/token-pool.ts
```diff
@@ -13,6 +13,8 @@ const TOKEN_POOL_IDL = {
   types: BASE_TOKEN_POOL_IDL.types,
 }
 
+const tokenPoolCoder = new BorshCoder(TOKEN_POOL_IDL)
```
aelmanaa: Worth a one-line comment on the TOKEN_POOL_IDL merge above explaining why the types splice is needed. The burn-mint IDL's state account references BaseConfig but doesn't define it in its own types, so accounts.decode("state") breaks without splicing in BASE_TOKEN_POOL_IDL.types. A future "simplification" back to plain BURN_MINT_TOKEN_POOL_IDL would silently break the decode.

A comment like "// splice in BASE types so BaseConfig is defined; required for accounts.decode" would save the next person from undoing it.

aelmanaa (commented):
Heads-up on a maintainability angle that crosses core and CCT — not a blocker for this PR, but worth deciding before the next CCT read op lands.

The token-pool state account is decoded in two places that have to agree byte-for-byte: core (solana/index.ts — getTokenPoolRemotes uses tokenPoolCoder.accounts.decode) and CCT (programs/token-pool.ts — decodeTokenPoolState). Both build the same TOKEN_POOL_IDL = { ...BURN_MINT_IDL, types: BASE.types } merge and the same BorshCoder. The IDL data files themselves are already shared (both import solana/idl/1.6.0/{BASE,BURN_MINT}_TOKEN_POOL.ts), so the account layout is single-sourced and won't drift. But the merge recipe + coder construction are duplicated, and the version pin (1.6.0) is independent in each — those can drift.

The two obvious fixes are both wrong here: importing core's coder would force a runtime import of the 1929-line solana/index.ts into the CCT op layer (which today only type-imports core — that's the load-bearing separation), and pushing the read into core would fatten getTokenPoolConfig for every core user who never touches CCT.

A third option that honors both separation and DRY: extract the merge + coder into a small neutral leaf under solana/idl/ (e.g. token-pool-coder.ts) that depends only on @coral-xyz/anchor and the IDL data files — never on solana/index.ts or cct/**. Core and CCT both import it; each keeps its own result projection (core's 4-field view, CCT's 15-field view). The load-bearing types-splice comment lives once, in one place.

With more CCT read ops coming (registry state, chain config, rate limits), establishing the leaf now — at the first duplication — is cheaper than retrofitting it at the third. If it feels premature for this PR, the fallback is just adding the merge comment to both copies and aligning the version pin. Either way, worth a decision before the next read op duplicates the coder again.

ccip-sdk/src/cct/solana/token-pool/operations/get-token-pool-state.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    Buffer.from([1, 1]),
+    Buffer.from([2, 0, 0, 0]),
+    key(12).toBuffer(),
+    key(13).toBuffer(),
+    key(14).toBuffer(),
+  ])
+}
+
+describe('Solana token pool getTokenPoolState', () => {
+  it('returns pool-type-specific state fields', async () => {
```
aelmanaa: This is the only test, and it covers just the happy path — a mocked getAccountInfo returning a valid buffer, asserting the decoded fields for both pool types. Two paths that can actually fail aren't exercised:

- Not found — getAccountInfo returns null and we throw CCIPTokenPoolStateNotFoundError. No test hits it.
- Decode failure — an account whose data isn't a pool state. This one matters because decodeTokenPoolState currently throws anchor's raw Error("Invalid account discriminator") unwrapped (see the other comment on token-pool.ts), and there's no test asserting what type of error comes out — which is why that leak went unnoticed.

Could you add two cases: (1) getAccountInfo → null asserts CCIPTokenPoolStateNotFoundError, and (2) a bad-discriminator buffer asserts a typed CCIPError.

mervin-link:
> Heads-up on a maintainability angle that crosses core and CCT — not a blocker for this PR, but worth deciding before the next CCT read op lands.
> 
> The token-pool state account is decoded in two places that have to agree byte-for-byte: core (solana/index.ts — getTokenPoolRemotes uses tokenPoolCoder.accounts.decode) and CCT (programs/token-pool.ts — decodeTokenPoolState). Both build the same TOKEN_POOL_IDL = { ...BURN_MINT_IDL, types: BASE.types } merge and the same BorshCoder. The IDL data files themselves are already shared (both import solana/idl/1.6.0/{BASE,BURN_MINT}_TOKEN_POOL.ts), so the account layout is single-sourced and won't drift. But the merge recipe + coder construction are duplicated, and the version pin (1.6.0) is independent in each — those can drift.
> 
> The two obvious fixes are both wrong here: importing core's coder would force a runtime import of the 1929-line solana/index.ts into the CCT op layer (which today only type-imports core — that's the load-bearing separation), and pushing the read into core would fatten getTokenPoolConfig for every core user who never touches CCT.
> 
> A third option that honors both separation and DRY: extract the merge + coder into a small neutral leaf under solana/idl/ (e.g. token-pool-coder.ts) that depends only on @coral-xyz/anchor and the IDL data files — never on solana/index.ts or cct/**. Core and CCT both import it; each keeps its own result projection (core's 4-field view, CCT's 15-field view). The load-bearing types-splice comment lives once, in one place.
> 
> With more CCT read ops coming (registry state, chain config, rate limits), establishing the leaf now — at the first duplication — is cheaper than retrofitting it at the third. If it feels premature for this PR, the fallback is just adding the merge comment to both copies and aligning the version pin. Either way, worth a decision before the next read op duplicates the coder again.

I extracted the shared IDL merge and token-pool coder into `solana/idl/token-pool-coder.ts`. Solana CCT now imports it.

ccip-sdk/src/errors/recovery.ts
```diff
@@ -217,6 +217,8 @@ export const DEFAULT_RECOVERY_HINTS: Partial<Record<CCIPErrorCode, string>> = {
     'This contract version is not supported by the CCT SDK. Check the contract address and its typeAndVersion.',
   CCT_OPERATION_UNSUPPORTED:
     'This operation is not available at the contract version in error.context. Verify the contract version supports it.',
+  CCT_TOKEN_POOL_STATE_DECODE_FAILED:
+    'Ensure the account belongs to a compatible CCIP token pool program and uses the expected state layout.',
```
apedrob: Could we make this error code more generic and chain-agnostic?

ccip-sdk/src/cct/solana/validate.ts
```diff
@@ -32,6 +32,18 @@ export function validatePublicKey(operation: string, param: string, value: unkno
...
+ * Asserts `value` is a valid Solana public key string.
+ * @throws CCTParamsInvalidError if `value` is not a valid Solana public key string.
+ */
+export function validatePublicKey(
+  operation: string,
+  param: string,
+  value: unknown,
+): asserts value is string {
+  parsePublicKey(operation, param, value)
+}
```
apedrob, line 45: Should we use `parsePublicKey` directly?

ccip-sdk/src/cct/solana/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,139 @@
...
+      poolProgram: programId.toBase58(),
+    })
+    const result = {
+      stateAddress: state.toBase58(),
+      programId: programId.toBase58(),
+      version,
+    }
+    const baseConfig = serializeBaseConfig(config)
+
+    if (params.poolType === 'lock-release') {
```
apedrob, line 156: Does it make sense making `params` type more narrow, i.e. discriminated union on `poolType`

ccip-sdk/src/cct/solana/validate.ts
mervin-link: I kept `validatePublicKey` for validation-only paths. This keeps `validate()` methods readable, while call ops that need a `PublicKey` use `parsePublicKey` directly.

ccip-sdk/src/errors/recovery.ts
mervin-link: Renamed it to `CCT_DATA_DECODE_FAILED`.

aelmanaa (approved):

apedrob (approved):

PR #309, opened 2026-07-24
fix(cct-sdk): Refactor solana token-admin-registry ops

mervin-link:
## What

- Refactor Solana TokenAdminRegistry lookup-table operations to support canonical `poolType` or custom `poolProgramAddress`
- Centralize pool-program reference types and resolution
- Improve lookup-table and set-pool test organization and validation coverage (generate, validation, execute)
- Refresh the root lockfile for npm ci compatibility

## Why

- Make Solana CCT pool operations easier to use with canonical pool types while preserving custom-program support
- Ensure to keep test coverage aligned with the operation lifecycle

## Note

- I'll follow up to refactor the remaining test files.


ccip-sdk/src/cct/solana/token-admin-registry/operations/append-to-lookup-table.ts
```diff
@@ -12,21 +12,32 @@ import {
...
-  tokenAddress?: string
-  poolProgramAddress?: string
   additionalAddresses?: string[]
   /** ALT authority. Defaults to payer for unsigned generation and wallet public key for execute. */
   authority?: string
-}
+} & (
+  | { tokenAddress?: never; poolType?: never; poolProgramAddress?: never }
+  | ({ tokenAddress: string } & PoolProgramRef)
+)
```
apedrob: Readability: Extract these types above like:

```
type AppendAdditionalAddressesParams = {
  additionalAddresses: string[]
  tokenAddress?: never
  poolType?: never
  poolProgramAddress?: never
}

type AppendCanonicalAddressesParams = {
  tokenAddress: string
  additionalAddresses?: string[]
} & PoolProgramRef

/**
 * Parameters shared by Solana TokenAdminRegistry `appendToLookupTable` generation and execution.
 *
 * Provide `tokenAddress` with exactly one of `poolType` or `poolProgramAddress` to append the
 * canonical CCIP addresses. Additional addresses may also be included.
 *
 * Otherwise, provide `additionalAddresses` only.
 */
type AppendToLookupTableParams = {
  lookupTableAddress: string
  /** ALT authority. Defaults to payer for unsigned generation and wallet public key for execute. */
  authority?: string
} & (AppendAdditionalAddressesParams | AppendCanonicalAddressesParams)
```

ccip-sdk/src/cct/solana/token-admin-registry/operations/append-to-lookup-table.ts
```diff
@@ -52,21 +63,20 @@ export class AppendToLookupTable extends SolanaOperation<
...
-    if (params.poolProgramAddress) {
-      validatePublicKey(this.name, 'poolProgramAddress', params.poolProgramAddress)
-    }
-    for (const [i, address] of (params.additionalAddresses ?? []).entries()) {
-      validatePublicKey(this.name, `additionalAddresses[${i}]`, address)
-    }
 
-    if (Boolean(params.tokenAddress) !== Boolean(params.poolProgramAddress)) {
+    const hasPoolProgram = params.poolType !== undefined || params.poolProgramAddress !== undefined
+    if (Boolean(params.tokenAddress) !== hasPoolProgram) {
```
apedrob: Here we use value semantics (`!== undefined`), but resolvePoolProgram uses `Object.hasOwn`. Suggest making both use `Object.hasOwn` so the two guards agree

ccip-sdk/src/cct/solana/token-admin-registry/operations/append-to-lookup-table.ts
apedrob, line 72: Issue?: `poolProgramAddress` is no longer validated here in validate(); it's now only parsed later in `buildUnsigned` via `resolvePoolProgram`

ccip-sdk/src/cct/solana/validate.ts
```diff
@@ -91,6 +96,26 @@ export function validatePoolType(
   }
 }
 
+/** Resolves a canonical pool type or custom program address. */
+export function resolvePoolProgram(operation: string, params: PoolProgramRef): PublicKey {
```
apedrob, line 100: Missing validation test in `cct/solana/validate.test.ts`

ccip-sdk/scripts/pack-cct.mjs
apedrob, line 1: I'm not sure this is the right approach long-term. It exists only because we carve CCT out at build/files/exports, and the script reconstitutes all that at pack time, which is a fair bit of complexity to maintain. Depending on how the bundle is actually consumed, we might not need it at all (`build:dev` already emits `dist/cct/**` via the existing `./dist/*` export) or CCT could become its own package.json so its exports resolve honestly. Happy to keep this out of scope for this PR though, let's take the discussion outside

mervin-link: Current problem is that CI build is failing because we excluded the build files but we are adding it in exports in package.json.

ccip-sdk/src/cct/solana/token-admin-registry/operations/append-to-lookup-table.ts
```diff
@@ -52,21 +73,24 @@ export class AppendToLookupTable extends SolanaOperation<
     validatePublicKey(this.name, 'lookupTableAddress', params.lookupTableAddress)
     validatePublicKey(this.name, 'payer', params.payer)
     if (params.authority) validatePublicKey(this.name, 'authority', params.authority)
+
+    const hasPoolProgramAddress = Object.hasOwn(params, 'poolProgramAddress')
+    const hasPoolProgram = Object.hasOwn(params, 'poolType') || hasPoolProgramAddress
```
aelmanaa: On devnet, calling this op with `{ tokenAddress, poolProgramAddress: <valid canonical program>, poolType: undefined }` passes `validate()` (this OR-guard is satisfied) and then throws from `resolvePoolProgram` inside `buildUnsigned()`:

```
CCTParamsInvalidError: Invalid appendToLookupTable parameter "poolType": provide exactly one of poolType or poolProgramAddress
```

Two problems with that: the input is rejected even though its values are unambiguous, and the rejection escapes the validate phase. The input is also type-legal: `poolType?: never` on the union arms accepts an explicit `undefined` (the project uses `strict: true` but not `exactOptionalPropertyTypes`), which is exactly what idiomatic option-bag spreads produce:

```ts
// caller builds the ref conditionally; the unused key lands as undefined
const ref = cond ? { poolType: 'burn-mint' } : { poolProgramAddress: addr }
await cct.generateUnsignedAppendToLookupTable({ lookupTableAddress, tokenAddress, ...ref, payer })
```

The existing `resolves pool programs` test passes unchanged under value semantics, so nothing loosens. Same change needed in `resolvePoolProgram` itself (comment there).

```suggestion
    // Value semantics (not key presence): an explicit `poolType: undefined` must not count,
    // matching resolvePoolProgram (which buildUnsigned delegates to).
    const hasPoolProgramAddress = params.poolProgramAddress !== undefined
    const hasPoolProgram = params.poolType !== undefined || hasPoolProgramAddress
```

ccip-sdk/src/cct/solana/token-admin-registry/operations/create-lookup-table.ts
```diff
@@ -62,14 +66,13 @@ export class CreateLookupTable extends SolanaOperation<
...
   readonly name = 'createLookupTable'
 
-  /** Validates all public keys before any RPC. */
+  /** Validates params before `buildUnsigned()` performs any RPC. */
   protected validate(params: GenerateCreateLookupTableParams): void {
     validatePublicKey(this.name, 'payer', params.payer)
     if (params.authority) validatePublicKey(this.name, 'authority', params.authority)
     if (params.mode === 'createEmpty') return
 
     validatePublicKey(this.name, 'tokenAddress', params.tokenAddress)
```
aelmanaa, line 75: Tiny consistency point: after the latest update, `appendToLookupTable.validate()` checks the pool reference (together-guard + `validatePublicKey` on `poolProgramAddress`), but here in `createLookupTable` the `createAndExtend` path leaves the pool ref entirely to `resolvePoolProgram` in `buildUnsigned()` — the same asymmetry @apedrob flagged on the append op.

It still fails before any RPC (the `resolvePoolProgram` call at line 107 sits above the `getSlot('finalized')` call at line 114), so this is not a bug. But a malformed/missing pool ref now surfaces from a different phase in the two sibling ops:

```
createLookupTable:    validate() -> OK,  buildUnsigned() -> throws "provide exactly one of..."
appendToLookupTable:  validate() -> throws "tokenAddress and exactly one of ... must be provided together"
```

ccip-sdk/src/cct/solana/validate.test.ts
```diff
@@ -74,6 +76,28 @@ describe('cct/solana validate', () => {
...
+    const invalidRefs: unknown[] = [
+      {},
+      { poolType: 'burn-mint', poolProgramAddress: PublicKey.default.toBase58() },
+      { poolType: 'nope' },
+      { poolProgramAddress: 'nope' },
+    ]
+    for (const params of invalidRefs) {
+      assert.throws(() => resolvePoolProgram('op', params as PoolProgramRef), CCTParamsInvalidError)
+    }
+  })
```
aelmanaa, line 99: could we add the explicit-`undefined` direction too? It's the case the current `Object.hasOwn` implementation gets wrong (accepted by the types via `poolType?: never`, rejected at runtime), so it's the regression worth pinning whichever semantics you land on:

```suggestion
  })

  it('resolves a pool reference with the other key explicitly undefined', () => {
    // Value semantics: an explicitly-set `undefined` key must not count as provided.
    const custom = PublicKey.default.toBase58()
    assert.equal(
      resolvePoolProgram('op', { poolProgramAddress: custom, poolType: undefined }).toBase58(),
      custom,
    )
    assert.equal(
      resolvePoolProgram('op', { poolType: 'burn-mint', poolProgramAddress: undefined }).toBase58(),
      TOKEN_POOL_PROGRAMS['burn-mint'],
    )
  })
```

ccip-sdk/src/cct/solana/validate.ts
```diff
@@ -91,6 +96,26 @@ export function validatePoolType(
...
+  const hasPoolType = Object.hasOwn(params, 'poolType')
+  const hasPoolProgramAddress = Object.hasOwn(params, 'poolProgramAddress')
```
aelmanaa: This is the root of the issue I raised on `append-to-lookup-table.ts`: `Object.hasOwn` tests key presence, so `{ poolProgramAddress: <valid>, poolType: undefined }` — type-legal because the union arms declare `poolType?: never`, and the natural output of option-bag spreads — trips the `hasPoolType === hasPoolProgramAddress` check even though the values are unambiguous.

Reproduced live on devnet through both `createLookupTable` and `appendToLookupTable`: `validate()` passes, then this throws from `buildUnsigned()`:

```
CCTParamsInvalidError: Invalid createLookupTable parameter "poolType": provide exactly one of poolType or poolProgramAddress
```

Value semantics fix it without loosening anything:

```suggestion
  // Value semantics (not key presence): an explicit `poolType: undefined` must not count.
  const hasPoolType = params.poolType !== undefined
  const hasPoolProgramAddress = params.poolProgramAddress !== undefined
```

aelmanaa:
Ran this branch live on Solana devnet before reviewing, and the refactored ops work exactly as advertised end-to-end: `deployToken` → `deployTokenPool(poolType: 'burn-mint')` → `getTokenPoolState` returns the real config (mint `BWUrSHFZfzjvdSbWyKFUjHNYdwtUrNHAPMnubuMT2o2w`, pool state `5cjjtbSdiEJhLuMvaD3a1VtiHM7wFfg14wdry1tfTK8m`); the wrong-`poolType` path surfaces the typed `TOKEN_POOL_STATE_NOT_FOUND` error correctly; and both `createLookupTable(poolType)` (ALT `3Q4JFBN3Mpr3ShJTrrzYhJ4tdZ2k8Fy6r5Hybhfyw8Wm`) and the new `appendToLookupTable(poolType)` shape (ALT `YzhNLSwvRvZGxjgme3vdsCeTuNFcGkpCcEHzpttKVjW`) write the identical, correct 10-address canonical block on-chain. The test reorganization is a real step up too — the generate/validation/execute grouping and the named param unions read very well.

Two non-blocking notes:

1. **`files` vs `exports`** — +1 to @mervin-link's point: with `./cct/evm` / `./cct/solana` now public exports, `ccip-sdk/package.json#files` still carries `!dist/cct/**` and `!src/cct/**`, so the published tarball would exclude the very files the exports resolve to (and CI build fails for the same reason — `publint` reports `./dist/cct/evm/index.js` and `./dist/cct/solana/index.js` do not exist in the pack). If the intent is that CCT ships publicly now, dropping those two negations is the whole fix:

```diff
   "files": [
     ...
-    "!dist/cct/**",
-    "!src/cct/**"
   ]
```

2. **Follow-up polish batch (all outside this PR's diff, so not anchoring them here):**
   - An actionable recovery hint for `TOKEN_POOL_STATE_NOT_FOUND` (currently restates the error; suggest "verify poolType matches the deployed pool / pass poolProgramAddress for custom / confirm the pool is initialized for this mint on this cluster").
   - Typed `decode<{ version: number; config: TokenPoolConfig }>` plus wrapped non-`Error` causes and an `accountOwner` field in the decode-error context in `programs/token-pool.ts` (cheapest wrong-program diagnostic).
   - A structured `CCTDataDecodeError(account, …)` signature matching its sibling CCT errors.
   - Two `getTokenPoolState` TSDoc touches: missing `const cct = SolanaTokenManager.fromChain(chain)` line in the example, and a caveat that `poolProgramAddress` results omit lock-release-specific fields; plus a doc line on `GetTokenPoolStateResult.programId` being the *resolved* input reference.

ccip-sdk/src/cct/solana/token-admin-registry/operations/create-lookup-table.ts
mervin-link: Added a check for pool reference on both ops.

mervin-link:
> * Two `getTokenPoolState` TSDoc touches: missing `const cct = SolanaTokenManager.fromChain(chain)` line in the example, and a caveat that `poolProgramAddress` results omit lock-release-specific fields; plus a doc line on `GetTokenPoolStateResult.programId` being the _resolved_ input reference.

Fixed this. Will do a follow up for the others.

aelmanaa (approved):
**✅ Approve — M1 fixed and live-proven at `cbdb519`.**

Re-reviewed at `cbdb519` (was `d0b2786`). All 4 inline comments are resolved:

- **M1 (Object.hasOwn → value semantics) ✅** — `validate.ts:101-102` and `append-to-lookup-table.ts:77-78` both changed from `Object.hasOwn` to `!== undefined`. The `appendToLookupTable.validate()` now delegates to `resolvePoolProgram` (single source of truth). Live-proven on devnet: `{ poolProgramAddress, poolType: undefined }` passes `validate()` and `buildUnsigned()` cleanly, and executes on-chain (ALT filled to 10 canonical addresses).
- **N-clt (createLookupTable pool ref validation) ✅** — `create-lookup-table.ts:76`: `resolvePoolProgram(this.name, params)` added to `validate()`, matching the sibling append op.
- **M1-t (regression test) ✅** — `validate.test.ts`: new test `'resolves pool references with the other key explicitly undefined'` (both directions). Also `append-to-lookup-table.test.ts`: new test `'ignores an undefined unused pool reference'`.
- **TSDoc polish ✅** — `programId` doc added ("resolved pool program address"), lock-release-fields caveat added, `const cct = SolanaTokenManager.fromChain(chain)` added to examples.

**One non-blocking nit stands:** `files` vs `exports` in `package.json` — `!dist/cct/**` and `!src/cct/**` still exclude the files the public exports resolve to. Already raised by mervin-link; the fix is dropping the two negations.

**Gates:** `npm ci` works clean, `npm run check` green (prettier + eslint + tsc), 26/26 touched tests pass.

apedrob (approved):

PR #310, opened 2026-07-27
feat(cct-sdk): Add register token solana op

mervin-link:
## What

- [DAPP-10603](https://smartcontract-it.atlassian.net/browse/DAPP-10603)
- Add Solana token registration operation support for CCT
- Handles both the `ownerProposeAdministrator` and `ccipAdminProposeAdministrator` ('owner' or 'ccip-admin' method)

 ## Why

- Enable registering Solana tokens to CCIP token admin registry through the CCT SDK

[DAPP-10603]: https://smartcontract-it.atlassian.net/browse/DAPP-10603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apedrob (approved):

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -407,6 +413,61 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
     return this.#appendToLookupTable.execute(this.chain, opts)
   }
 
+  /**
+   * Builds an unsigned Solana token registration instruction.
```
aelmanaa, line 417: I live-tested this on devnet from this head: after `registerToken` the on-chain `TokenAdminRegistry` has `administrator = 11111…111` (unset) and only `pending_administrator` set — and an immediate `setPool` reverts `Unauthorized (7000)`. So this op **proposes** an administrator; registration completes only after the pending admin accepts (the `acceptAdmin` op from DAPP-10605). This matches the canonical program exactly: `owner_propose_administrator` writes `administrator = 0` and `pending_administrator = proposed` (`init_with_pending_administrator`, chainlink-ccip `chains/solana/contracts/programs/ccip-router/src/instructions/v1/token_admin_registry.rs:345-353`), `set_pool` gates on `authority == token_admin_registry.administrator` (`src/token_context.rs:154`, `Unauthorized` = 7000, `src/lib.rs:602`), and only `accept_admin_role_token_admin_registry` — callable solely by the `pending_administrator` (`token_context.rs:194`) — promotes pending → administrator (`token_admin_registry.rs:121-124`). The docstrings on both facade methods should say that and link the follow-up, otherwise the taught journey dead-ends at `setPool`:

```suggestion
   * Builds an unsigned Solana token registration instruction.
   *
   * This **proposes** the registry administrator (sets `pending_administrator` on the
   * TokenAdminRegistry PDA): the proposed admin must then run `acceptAdmin` (DAPP-10605)
   * before `setPool` will authorize. The mint authority is proposed as the administrator.
   * Choose `owner` when the mint authority signs, or `ccip-admin` when the Router CCIP
   * admin signs.
```

(replacing the current lines 417–420 summary) — and please add `@see {@link setPool}` here plus a matching `@remarks`/`@see acceptAdmin` on `registerToken` below, and a back-link from `setPool`'s docs (`setPool` currently `@see`s only `deployTokenPool`/`createTokenAccount`, so the doc graph skips registration+acceptance entirely).

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -407,6 +413,61 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
...
+   * ```
+   */
+  generateUnsignedRegisterToken(
+    opts: GenerateRegisterTokenParams,
+  ): Promise<GenerateRegisterTokenResult> {
+    return this.#registerToken.generate(this.chain, opts)
+  }
+
+  /**
+   * Registers a token using the executing wallet as the registration authority and fee payer.
```
aelmanaa: Same as above for the execute variant — "Registers a token…" overstates what lands on-chain (devnet-verified, and by construction in the program: `init_with_pending_administrator` sets `administrator` to the zero pubkey and only fills `pending_administrator` — ccip-router `src/instructions/v1/token_admin_registry.rs:345-353` — so nothing is "registered" until the accept instruction runs):

```suggestion
   * Proposes token registration using the executing wallet as the registration authority and
   * fee payer. The wallet becomes `pending_administrator`; run `acceptAdmin` (DAPP-10605) with
   * the same wallet to complete registration before calling `setPool`.
```

ccip-sdk/src/cct/solana/token-admin-registry/operations/register-token.ts
```diff
@@ -0,0 +1,202 @@
...
+  OWNER: 'owner',
+  CCIP_ADMIN: 'ccip-admin',
+} as const
+
+/** Authorization path used to register a token in the TokenAdminRegistry. */
+export type RegisterTokenMethod =
+  (typeof REGISTER_TOKEN_METHODS)[keyof typeof REGISTER_TOKEN_METHODS]
+
+type RegisterTokenParams = {
+  /** Token mint to register. Its current mint authority becomes the registry administrator. */
```
aelmanaa: Nit, same propose-vs-set semantics at the param level (on-chain, the proposed key lands only in `pending_administrator` — ccip-router `src/instructions/v1/token_admin_registry.rs:348` — and note the program's `owner_propose_administrator` actually takes the admin as a free `token_admin_registry_admin` argument, `src/lib.rs:370-376`; "its current mint authority is proposed" is this SDK op's choice, not a chain rule):

```suggestion
  /** Token mint to register. Its current mint authority is proposed as the registry administrator (pending until accepted). */
```

ccip-sdk/src/cct/solana/token-admin-registry/operations/register-token.ts
```diff
@@ -0,0 +1,202 @@
...
+  tokenAddress: string
+  /**
+   * CCIP contract to resolve the TokenAdminRegistry/Router from — the registry itself,
+   * a Router, OnRamp, OffRamp, or TokenPool address all work.
```
aelmanaa: This docstring (inherited from `setPool`) claims "the registry itself, a Router, OnRamp, OffRamp, or TokenPool address all work", but `SolanaChain.getTokenAdminRegistryFor` only resolves a **Router** (as-is) or an **OffRamp** (→ its router); anything else throws `CCIPContractNotRouterError`. Trim the claim (or widen the resolver once and keep the wording everywhere):

```suggestion
   * CCIP contract to resolve the TokenAdminRegistry/Router from — a Router or OffRamp
   * address works.
```

ccip-sdk/src/cct/solana/token-admin-registry/operations/register-token.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+  return program.methods.ownerProposeAdministrator(mintAuthority).accounts(accounts).instruction()
+}
+
+async function buildCcipAdminInstruction(
+  program: RouterProgram,
+  accounts: RegisterTokenAccounts,
+  mintAuthority: PublicKey,
+): Promise<TransactionInstruction> {
+  const routerConfig = await program.account.config.fetch(accounts.config)
```
aelmanaa: If the router `config` PDA can't be fetched here, anchor throws a raw non-`CCIPError` — the only ungrounded throw path in the op. Worth wrapping in a typed `CCIPError` subclass (or at least a `CCTParamsInvalidError` for `address`) so the taxonomy holds: a resolved-but-uninitialized router currently surfaces as an anchor internals message.

ccip-sdk/src/cct/solana/token-admin-registry/operations/register-token.ts
```diff
@@ -0,0 +1,202 @@
...
+    const routerAddress = await chain.getTokenAdminRegistryFor(opts.address)
+    const router = new PublicKey(routerAddress)
+    const tokenMint = new PublicKey(opts.tokenAddress)
+    const payer = new PublicKey(opts.payer)
+    const authority = new PublicKey(opts.authority ?? opts.payer)
+
+    const mintAccount = await resolveTokenMint(chain.connection, tokenMint)
+    const { mintAuthority } = unpackMint(tokenMint, mintAccount, mintAccount.owner)
+    if (!mintAuthority) {
+      throw new CCTParamsInvalidError(this.name, 'tokenAddress', 'token mint has no mint authority')
```
aelmanaa: This guard runs **before** the path switch, so it gates the `ccip-admin` path too — but the on-chain `RegisterTokenAdminRegistryByCCIPAdmin` context has **no** mint-authority constraint at all: its only auth check is `authority == config.owner` (chainlink-ccip `chains/solana/contracts/programs/ccip-router/src/token_context.rs:34`, `#[account(mut, address = config.owner @ CcipRouterError::Unauthorized)]`), the mint is a plain read-only `InterfaceAccount<Mint>` used only as the TAR PDA seed (`token_context.rs:24,31`), and the instruction takes the admin as a **free argument** `token_admin_registry_admin: Pubkey` (`src/lib.rs:340-346`) whose only on-chain constraint is non-zero (`init_with_pending_administrator`, `src/instructions/v1/token_admin_registry.rs:339-343`, `InvalidTokenAdminRegistryInputsZeroAddress`). The whole point of the ccip-admin path is registering tokens the owner path *can't*: fixed-supply / null-mint-authority tokens, or delegating admin to a DAO/vault. I reproduced this live on devnet from this head: a mint with its authority revoked (`ErfkvJ69TKQTcDdmqf6pdNTCwGaALBHpb4TGhiLGGsXU`) gets client-rejected with `token mint has no mint authority` on the `ccip-admin` path — exactly the token this path exists for. And `buildCcipAdminInstruction` (line 101) always proposes `mintAuthority` as the admin, so even when it works it can only ever nominate the mint authority (EVM's canonical `proposeAdministrator(token, administrator)` takes an explicit admin).

In fact the coupling is stricter-than-chain on the **owner path too**: `owner_propose_administrator` takes the same free `token_admin_registry_admin` argument (`src/lib.rs:370-376`) — the owner context only constrains the **signer** (`authority == mint.mint_authority.unwrap()`, `token_context.rs:82-86`), never the proposed admin. So a mint authority can legitimately nominate a different key (a DAO, a multisig) as admin on-chain; this op can't express that either. The org's own Solana operator tooling agrees: `ccip-solana-bs58-generator`'s `router owner-propose-administrator` takes a required explicit `--token-admin-registry-admin` (`src/commands/router/owner-propose-administrator.ts:15`) passed verbatim as the instruction arg (`src/programs/router/instructions.ts:42`) and never derives it from — or requires — a mint authority. This op is the only surface in the org that couples the proposed admin to the mint authority (and it would be the org's *first* working ccip-admin TS surface — bs58 ships the IDL entry but no command — so shipping it over-constrained sets the precedent).

Sketch: add an optional `administrator?: string` param honored on **both** paths (default: the mint authority when present), keep the mint-authority *signer* guard inside the `owner` branch only (that one mirrors the on-chain constraint), and drop the pre-switch null-mint-authority throw:

```ts
// params
/** Administrator to propose. Defaults to the mint authority when present; required (ccip-admin) when the mint has none. */
administrator?: string

// buildUnsigned — owner branch keeps the signer==mintAuthority guard (on-chain: token_context.rs:82-86);
// both branches:
const proposed = opts.administrator
  ? new PublicKey(opts.administrator)
  : mintAuthority // may be null → then administrator is required
if (!proposed)
  throw new CCTParamsInvalidError(this.name, 'administrator',
    'required when the mint has no mint authority')
// then per branch:
return program.methods.ownerProposeAdministrator(proposed).accounts(accounts).instruction()    // owner
return program.methods.ccipAdminProposeAdministrator(proposed).accounts(accounts).instruction() // ccip-admin
```

If the current constraint is intentional, it needs explicit documentation on `registrationMethod` — but I'd expect the param, since the on-chain instruction takes the administrator as a free argument on both variants.

ccip-sdk/src/cct/solana/token-admin-registry/operations/register-token.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+    const config = deriveRouterConfigPda(router)
+    const tokenAdminRegistry = deriveTokenAdminRegistryPda(router, tokenMint)
+    if (await chain.connection.getAccountInfo(tokenAdminRegistry)) {
+      throw new CCTParamsInvalidError(this.name, 'tokenAddress', 'token is already registered')
```
aelmanaa: This fires whenever the TAR PDA merely *exists* — which is true as soon as `registerToken` has proposed (pending, not accepted). Live on devnet from this head: re-running `registerToken` on a token whose registration is still pending returns "token is already registered", which points a dev whose accept failed in the wrong direction. The **existence check itself is right**: both register contexts use Anchor `init` on the TAR PDA (ccip-router `src/token_context.rs:23,71`), so re-sending the same instruction would fail on-chain anyway — this preflight isn't stricter than the chain. But the *message* is wrong about the state and the way out: on-chain, a pending-only entry is **not final** — the program ships dedicated `owner_override_pending_administrator` / `ccip_admin_override_pending_administrator` instructions (`src/lib.rs:384,354`) that overwrite `pending_administrator` for exactly this state (guarded by `administrator == 0`, `src/instructions/v1/token_admin_registry.rs:50-56,85-91`, else `InvalidTokenAdminRegistryProposedAdmin`). Now that `acceptAdmin` exists (DAPP-10605), the message can name the actual next action:

```suggestion
        throw new CCTParamsInvalidError(
          this.name,
          'tokenAddress',
          'a registry entry already exists for this token (possibly pending admin acceptance) — use acceptAdmin/setPool instead of registering again',
        )
```

(Ideally decode the account and branch pending vs accepted, but the message fix alone removes the trap. Separately worth a ticket, not this PR: the SDK has no surface at all for the on-chain override-pending instructions, so a mis-proposed pending admin is currently unrecoverable through the SDK.)

ccip-sdk/src/cct/solana/token-admin-registry/operations/register-token.test.ts
```diff
@@ -0,0 +1,149 @@
+import assert from 'node:assert/strict'
+import { createHash } from 'node:crypto'
+import { describe, it } from 'node:test'
+
+import { TOKEN_PROGRAM_ID } from '@solana/spl-token'
+import { Keypair, PublicKey } from '@solana/web3.js'
+
+import { CCTParamsInvalidError } from '../../../errors.ts'
```
aelmanaa: This import order trips `import-x/order` (the only warning in an otherwise clean `npm run check` at this head): `../../../errors.ts` must come after the `../../../../*` group. `eslint --fix` handles it:

```suggestion
import { ChainFamily } from '../../../../networks.ts'
import type { SolanaChain } from '../../../../solana/index.ts'
import { CCTParamsInvalidError } from '../../../errors.ts'
```

(replacing lines 8–10).

aelmanaa (changes requested):
**On-chain verification (devnet, this exact head):** I deployed a fresh mint (`2E9eR9NQeHQuTJ4p3VQMSnQJFzMUBRmq87dJX1j78JrA`), executed `registerToken` (owner path, tx `4x1M612…`), and decoded the resulting `TokenAdminRegistry` PDA (`B9rmPGHW…`, verified twice — SDK read + raw RPC): `pending_administrator == the proposed admin`, `administrator == zeros`. So the op's semantics are **propose**, and an immediate `setPool` reverts on-chain with `Unauthorized (7000)` (`authority == administrator` fails against the zeroed admin). All of this is exactly what the canonical program prescribes (chainlink-ccip `chains/solana/contracts/programs/ccip-router`): `init_with_pending_administrator` zeroes `administrator` and sets only `pending_administrator` (`src/instructions/v1/token_admin_registry.rs:345-353`), and `SetPoolTokenAdminRegistry` constrains `authority == token_admin_registry.administrator` (`src/token_context.rs:154`). That's not a defect in this PR's wiring — the instruction, PDAs, and both auth guards are exact against the canonical program, and the propose lands correctly :

1. **Docs must say "propose" and link the accept step.** The `acceptAdmin` op from **DAPP-10605** completes the journey (register → accept → setPool) — I'm *not* asking for a new op or a rename here, just `@remarks`/`@see` cross-references on both facade methods (and a back-link from `setPool`) so the required sequence is discoverable from the docstrings. An optional follow-up discussion: whether `registerToken` vs `proposeAdministrator` naming should be revisited once both PRs are in — fine as a separate conversation.
2. **The one blocking change: the `ccip-admin` path (inline comment at `register-token.ts:139`).** It client-rejects null-mint-authority tokens (live-reproduced on devnet) and can only ever nominate the mint authority as admin — but registering fixed-supply/delegated-admin tokens is that path's reason to exist, and the on-chain instruction takes the administrator as a free `token_admin_registry_admin` argument on **both** variants (`src/lib.rs:340-346, 370-376`); the ccip-admin context never reads the mint beyond using it as the PDA seed (`src/token_context.rs:24,31,34`). The SDK must not be more restrictive than the chain here. An optional `administrator?` param fixes both.
3. Nits inline: "already registered" misfires on the pending state (live-reproduced), the `address` docstring overclaims what resolves, the config-fetch raw throw, one import-order lint warning, and two cheap discriminator asserts to lock the tests.

aelmanaa (approved):
**✅ Approve — all 9 comments addressed at `9e94aff`.**

Re-reviewed at `9e94aff` (was `ac50e35`). Every finding from the prior request-changes is fixed:

- **F2 (ccip-admin over-constrained) ✅** — `administrator?: string` param added, defaults to mint authority when present. Mint-authority guard moved inside `buildOwnerInstruction` only. `buildCcipAdminInstruction` takes `administrator` directly — no mint-authority requirement. Null-mint + no-administrator → typed `CCTParamsInvalidError('administrator', 'is required when the mint has no mint authority')`. `registrationMethod` now optional (defaults `owner`). Live-proven on devnet: ccip-admin path with explicit administrator on a null-mint-authority token now builds correctly.
- **F1/F4 (docs journey) ✅** — facade docs say "This proposes the registry administrator", `@see {@link generateUnsignedAcceptAdmin}` + `@see {@link generateUnsignedSetPool}`. `setPool` docs back-link `@see {@link registerAdmin}` and say "The token must first be registered and its proposed administrator accepted".
- **F3 (address overclaim) ✅** — trimmed to "a Router or OffRamp address works".
- **F5 (already-registered message) ✅** — "a registry entry already exists for this token (possibly pending admin acceptance) — use acceptAdmin/setPool instead of registering again".
- **F6 (config-fetch raw throw) ✅** — wrapped in try/catch → `CCTParamsInvalidError('address', 'Router config could not be fetched', { cause })`.
- **Import-order lint ✅** — `errors.ts` now after `../../../../` group; eslint clean.
- **Discriminator asserts ✅** — both `af51a0f6ce841216` (owner) and `da258b6b8ee433db` (ccip-admin) pinned as hex literals.
- **Rename ✅** — `registerToken` → `registerAdmin` end-to-end (class, facade, file, tests).

**Two cosmetic residuals (non-blocking):** (1) class docstring at `register-admin.ts:129` still says "Registers a token…" (non-rendered class level); (2) `{@link acceptAdmin}` / `{@link generateUnsignedAcceptAdmin}` are dangling until PR #317 lands — both resolve post-merge in the agreed order (#310 → #313 → #317).

**CI note:** `npm ci` fails on the lockfile sync issue (`@emnapi/core@1.11.2` missing) — same repo-wide infra issue, not from this PR's code.

PR #307, opened 2026-07-22
feat(cct-sdk): Add deploy evm token pool + version resolution

apedrob:
## What

- Add `deployTokenPool` to `EVMTokenManager` (build + execute): deploys a TokenPool at 2.0.0, type selects the contract (`BurnMintTokenPool`, `BurnFromMintTokenPool`, `BurnWithFromMintTokenPool`, `LockReleaseTokenPool`)
   - `generateUnsignedDeployTokenPool` for offline/multisig; `deployTokenPool` signs + submits, returns { hash, contractAddress }
- Add `resolveTokenPool`: read onchain typeAndVersion, narrow to a known pool type + version
- `transferOwnership` resolves the pool's real type/version to pick the ABI and floor-match the encoder (was a hardcoded interface)
- Slim version file to the version/resolution axis; move deploy artifacts (creation-bytecode map + DeployableTokenPoolType) next to the deploy op

## Why

- CCT needs first-class pool deployment to complete the 2.0.0 pool deployment
- Resolving type/version on-chain keeps encoder/ABI selection correct as pool versions evolve

## Testing

- CI for `deployTokenPool` (golden-vector init-code per type incl. lockBox, param validation, execute paths)
- `transferOwnership` probe + version-resolution tests; full CCT suite green, typecheck + lint clean

## Notes
- LockRelease scope: deploys with a caller-supplied lockBox (pre-deployed ERC20LockBox for the same token). deployLockBox + authorizing the pool on it are follow-ups

PR #311, opened 2026-07-27
feat(cct-sdk): Deploy Lockbox

apedrob:
## What

- Add deployLockbox (ERC20LockBox 2.0.0) to EVMTokenManager

## Why
- `LockReleaseTokenPool` requires a pre-deployed `ERC20LockBox` for the same token (its constructor calls `lockbox.isTokenSupported(token)`). Until now the SDK could deploy the pool but not the lockbox, so LockRelease wasn't exercisable SDK-only.

## Testing 
- Golden-vector init-code + validation (zero/invalid token) + execute paths; typecheck, lint, and tests green

## Notes
- Follow-up mentioned in https://github.com/smartcontractkit/ccip-tools-ts/pull/307#issuecomment-5090779114


ccip-sdk/src/cct/evm/lockbox/operations/deploy-lockbox.ts
```diff
@@ -0,0 +1,70 @@
...
+  deploymentTx,
+} from '../../operation.ts'
+import { submit } from '../../submit.ts'
+import { validateAddress } from '../../validate.ts'
+import { LOCKBOX_BYTECODE, LOCKBOX_INTERFACE } from '../interface.ts'
+
+/** Parameters for {@link DeployLockbox} — deploys `ERC20LockBox` (v2.0.0). */
+export interface DeployLockboxParams {
+  /** Address of the token the lockbox escrows; the v2.0.0 constructor reverts on the zero address. */
+  token: string
```
mervin-link, line 28: Nit: Currently, I'm using `tokenAddress` in Solana ops. Should we do the same for evm? Including the deploy token pool. Just want to sync on this one.

ccip-sdk/src/cct/evm/lockbox/operations/deploy-lockbox.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/** Deploys an `ERC20LockBox`; `execute` resolves to `{ hash, contractAddress }`. */
+export class DeployLockbox extends EVMOperation<DeployLockboxParams> {
+  readonly name = 'deployLockbox'
+
+  /** Validates the constructor params before building init-code. */
+  protected validate(params: DeployLockboxParams): void {
+    validateAddress(this.name, 'token', params.token)
+    if (params.token === ZeroAddress)
```
mervin-link, line 40: Can be extracted to `validate.ts` since we also used it in deploy token pool.

mervin-link (approved):
Small comments and not a blocker.

ccip-sdk/src/cct/evm/lockbox/operations/deploy-lockbox.ts
apedrob: Yes, it makes sense. Will change it

ccip-sdk/src/cct/evm/index.ts
```diff
@@ -180,52 +234,46 @@ export class EVMTokenManager extends TokenManager<typeof ChainFamily.EVM> {
...
    *   wallet,
    * })
    * ```
    */
-  deployToken(opts: EVMExecuteParams<DeployTokenParams>): Promise<DeployResult> {
-    return this.#deployToken.execute(this.chain, opts)
+  deployLockbox(opts: EVMExecuteParams<DeployLockboxParams>): Promise<DeployResult> {
+    return this.#deployLockbox.execute(this.chain, opts)
   }
+
```
aelmanaa: **[blocker — one-line]** This branch's head doesn't pass its own gate: `npm run check -w ccip-sdk` exits 1, and the *only* failure is prettier on this stray blank line (`276:1 Delete ⏎`; tsc and the rest of eslint are green — I verified in an isolated build of `2414b29`). Same finding as F1 of my branch review; it survives into the PR unchanged.

```suggestion
```

ccip-sdk/src/cct/evm/artifacts/abi/V2_0_0/erc20-lockbox.ts
```diff
@@ -0,0 +1,253 @@
...
+    name: 'UnauthorizedCaller',
+    inputs: [{ name: 'caller', type: 'address', internalType: 'address' }],
+  },
+  {
+    type: 'error',
+    name: 'UnsupportedToken',
+    inputs: [{ name: 'token', type: 'address', internalType: 'address' }],
+  },
+  { type: 'error', name: 'ZeroAddressNotAllowed', inputs: [] },
+] as const
```
aelmanaa, line 254: **[major — CI/provenance]** These two new artifact files are the only ones in `artifacts/` without the `// generate:end` marker, and that silently disables the drift gate for exactly the contract this PR adds — in two different ways, both of which I verified live at `2414b29`:

1. **This abi file** is picked up by the root `generate` grep, but `generate.cjs` then prints `generate error on "ccip-sdk/src/cct/evm/artifacts/abi/V2_0_0/erc20-lockbox.ts": Error: no "// generate:end" found` **and still exits 0** (`generate.cjs:38` throws, the catch only logs; only a content *diff* sets `process.exitCode = 1`). So `npm run generate:lint` is green while skipping it.
2. **The bytecode file** (`bytecode/V2_0_0/erc20-lockbox.ts`) is worse off: its `// generate:` is mid-line (`export default // generate:`), so the file-list grep (`^\s*//\s*generate\b`) never even lists it — compare `bytecode/V2_0_0/cross-chain-token.ts`, which is only *included in the gate via its trailing `// generate:end` line*.

Adding the two markers fixes both, and I verified it's safe: with them in place, `generate:lint` reports both files **`up-to-date`** (i.e. the generator reproduces your vendored content byte-for-byte from `@chainlink/contracts-ccip@2.0.0` — nice provenance side-effect) and exits 0. Here:

```suggestion
  // generate:end
] as const
```

…and in `bytecode/V2_0_0/erc20-lockbox.ts`, append `// generate:end` as the final line (after the hex string), matching `cross-chain-token.ts:4`. Optional hardening while you're in there (pre-existing, applies to all artifacts): make `generate.cjs`'s per-file `catch` set `process.exitCode = 1` so a missing/typo'd marker can never pass CI silently again.

ccip-sdk/src/cct/evm/lockbox/operations/deploy-lockbox.ts
```diff
@@ -0,0 +1,70 @@
+/**
+ * deployLockbox — deploys an `ERC20LockBox` (v2.0.0) via raw init-code. The tx has no
+ * `to`; `execute` returns the deployed contract address. A lockbox escrows a single token
+ * for `LockReleaseTokenPool`s; deploy it before the pool, then authorize the pool on it via
+ * {@link AuthorizeLockboxCallers}. Mirrors `token-pool/operations/deploy-token-pool.ts`.
```
aelmanaa: **[minor — docs, sequencing]** This `{@link AuthorizeLockboxCallers}` — plus the four `{@link authorizeLockboxCallers}` in `cct/evm/index.ts` (lines 186, 214, 242, 260) — reference DAPP-10788's op, which doesn't exist anywhere at this head (I grepped: doc-mentions only). Runtime-harmless, but they're broken links in the generated API docs until 10788 lands. Totally fine if 10788 merges right behind this PR (it's already reviewed); if there's any chance of a gap, de-link to plain text, e.g. ``…then authorize the pool on it via `authorizeLockboxCallers` (DAPP-10788).`` Teaching the 6-step sequence here is the right call either way — this is only about link integrity.

aelmanaa (approved):
**✅ Approve-with-fixes — and this PR closes the loop my #307 approval was conditioned on. Live-proven.**

I ran the **complete 2-step LockRelease journey SDK-only on Sepolia at this head (`2414b29`)** — the half that #307 couldn't do because `deployLockbox` didn't exist yet:

1. **This PR's `deployLockbox({token, wallet})`** → lockbox [`0x774D34A51074F11367209d42aba501EE59BB10F6`](https://sepolia.etherscan.io/address/0x774D34A51074F11367209d42aba501EE59BB10F6) (tx [`0x5525239e849b6bf755602a523a98c2101bab91e8b8247f5baae61e9118a47aa6`](https://sepolia.etherscan.io/tx/0x5525239e849b6bf755602a523a98c2101bab91e8b8247f5baae61e9118a47aa6)) — on-chain `typeAndVersion() == "ERC20LockBox 2.0.0"`, `isTokenSupported(token) == true`, `owner()` == the deployer, read back via the SDK **and** raw `eth_call` independently.
2. **Its `contractAddress` fed straight into the base branch's `deployTokenPool({type:'LockReleaseTokenPool', lockbox, …})`** → pool [`0xcEe7953E2514CBd9345fAA8b89664CAFE8ad605C`](https://sepolia.etherscan.io/address/0xcEe7953E2514CBd9345fAA8b89664CAFE8ad605C) (tx [`0x6f302a715936bf3ac85f9fa0814515be1694f9d5b197c306ee51b651bbc51a74`](https://sepolia.etherscan.io/tx/0x6f302a715936bf3ac85f9fa0814515be1694f9d5b197c306ee51b651bbc51a74)) — **on-chain `getLockBox()` == the step-1 address**, `typeAndVersion() == "LockReleaseTokenPool 2.0.0"`, and the ctor's max `forceApprove(pool → lockbox)` visible on-chain. That's M4 of the deploy-pool review closed end-to-end.

Both rejection guards fired **client-side with no tx** (zero and malformed `token` — nonce verified unchanged), exactly mirroring canonical `ERC20LockBox.sol`'s ctor `ZeroAddressNotAllowed` — neither stricter nor looser. Provenance re-verified at this head: vendored bytecode **byte-identical** and ABI **28/28 canonical-identical** to npm `@chainlink/contracts-ccip@2.0.0`; with the `generate:end` markers added the generator itself reproduces both files (`up-to-date`).

Also good: because this PR targets `feature/DAPP-10545-deploy-evm-pool` directly, the `lockBox` → `lockbox` rename of #307's surface is **inside this PR's reviewable diff** — that resolves the coherence condition I left on #307 (one nit: mention the rename in this PR's description). 27/27 delta tests pass; merging onto the base's current head is conflict-free with an identical EVM-only delta.

**Before merge**: the two inline fixes (stray blank line `cct/evm/index.ts:276` — the head's only `npm run check` failure — and the missing `// generate:end` markers). **Remaining follow-up**: DAPP-10788 `authorizeLockboxCallers` (already reviewed approve-with-fixes) — a fresh lockbox has an empty authorized-caller set, so the pool's lock/release reverts `UnauthorizedCaller` until it lands; please restack 10788 onto this head `2414b29` (its current base predates the Solana `TransactionHash` rename and it doesn't compile standalone — I verified the restack merge is conflict-free), which also picks up the `:276` fix once made.

PR #313, opened 2026-07-27
feat(cct-sdk): Add transfer admin solana op

mervin-link:
## What

- [DAPP-10604](https://smartcontract-it.atlassian.net/browse/DAPP-10604)
- Add Solana CCT `transferAdmin` and `generateUnsignedTransferAdmin` TokenAdminRegistry operations
- Validate current admin authority before building `transferAdminRoleTokenAdminRegistry`
 
## Why

- Enable SDK consumers to propose a new Solana TokenAdminRegistry administrator

[DAPP-10603]: https://smartcontract-it.atlassian.net/browse/DAPP-10603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DAPP-10604]: https://smartcontract-it.atlassian.net/browse/DAPP-10604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/index.ts
```diff
@@ -413,6 +419,58 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
...
+   * ```ts
+   * const unsigned = await cct.generateUnsignedProposeAdmin({
+   *   tokenAddress: mint,
+   *   address: router,
+   *   newAdmin,
+   *   payer: currentAdmin,
+   * })
+   * ```
+   */
+  generateUnsignedProposeAdmin(
```
aelmanaa: **This op should be named `transferAdminRole`, not `proposeAdmin` — and the rename has to happen before merge, because afterwards it's a public-surface break.**

Every naming authority we have calls this operation a *transfer*, and reserves *propose* for the **registration** step:

- The Solana Router entrypoint this op invokes is `transfer_admin_role_token_admin_registry` (chainlink-ccip `ccip-router/src/lib.rs:416`); the router's *propose* entrypoints (`owner_propose_administrator` / `ccip_admin_propose_administrator`, `lib.rs:339-357`) are the registration ops our `registerToken` already wraps.
- I executed this op live on devnet (tx [`5sQaftoHuU3cWekpZUri6kqhZVuYsLBYZZTKLjyTFfgpmQYqwxXffSmJ9VRaT9he2xGjWjATY2YYgbpbtDpgCH8m`](https://explorer.solana.com/tx/5sQaftoHuU3cWekpZUri6kqhZVuYsLBYZZTKLjyTFfgpmQYqwxXffSmJ9VRaT9he2xGjWjATY2YYgbpbtDpgCH8m?cluster=devnet)): the instruction discriminator it emits is `b262cbb5cb6b6a0e` = sha256(`global:transfer_admin_role_token_admin_registry`)[0..8], and the chain's own program log for our transaction reads `Instruction: TransferAdminRoleTokenAdminRegistry`.
- EVM canonical `TokenAdminRegistry.sol` names this exact op `transferAdminRole` (`:141`) — while EVM's `proposeAdministrator` (`:182`) is the *registration* op, i.e. the sibling of Solana `registerToken`, not of this op. Naming this `proposeAdmin` pre-claims the wrong EVM sibling and will collide the day an EVM `proposeAdministrator` op lands in `cct/evm`.
- Org precedent: `ccip-solana-bs58-generator` already exposes this instruction as the `transfer-admin-role` command and names registration `owner-propose-administrator`.

There's also a concrete user trap: after `registerToken`, the registry shows a `pendingAdministrator` — a user reaching for `proposeAdmin` in autocomplete to *finish registration* instead gets `must be the current token administrator` (I hit exactly this live, on a registry whose administrator is the all-zeros default). The right next step there is the accept op (DAPP-10605), never this one.

Suggested rename across the delta: op class `TransferAdminRole`, `name = 'transferAdminRole'`, facade `generateUnsignedTransferAdminRole`/`transferAdminRole`, and the `Generate*/Execute*` param/result type aliases to match. `newAdmin` is already the canonical param name — keep it.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -413,6 +419,58 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
     return this.#appendToLookupTable.execute(this.chain, opts)
   }
 
+  /**
+   * Builds an unsigned Solana instruction that proposes a new token administrator.
+   *
+   * The supplied authority must be the current token administrator. The proposed administrator
```
aelmanaa:
```
   * The supplied authority must be the current token administrator. The proposed administrator
   * must accept the role separately.
```

"Must accept the role separately" — separately *how*? There's no `@see` and no named next step, and the same gap exists on the `proposeAdmin` execute docstring below. This op is also unusable right after `registerToken` (on-chain the fresh registry's `administrator` is the zero default until the pending admin accepts — chainlink-ccip `ccip-router/src/token_context.rs:264-265` init, `:132` constraint; I had to perform the accept instruction out-of-SDK on devnet before this op would build). When DAPP-10605's accept op lands, please wire the journey both ways: `@remarks` here naming the accept op as the required follow-up (and distinguishing this *transfer* from `registerToken`'s *registration*), plus `@see` links registerToken ↔ accept ↔ setPool and propose/transfer ↔ accept. This was the documentation half of the register-token review's F1 fix and it's still missing.

ccip-sdk/src/cct/solana/token-admin-registry/operations/propose-admin.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    connection: {},
+    getTokenAdminRegistryFor: async (address: string) => {
+      onAddress?.(address)
+      return ROUTER
+    },
+    getRegistryTokenConfig: async () => ({ administrator }),
+  } as unknown as SolanaChain
+}
+
+function generate(opts = {}) {
```
aelmanaa:
```suggestion
function generate(opts: Partial<GenerateProposeAdminParams> = {}) {
```

`opts = {}` types the overrides as `{}`, so a typo'd key in a test (`{ authoirty: … }`) compiles silently. `Partial<GenerateProposeAdminParams>` keeps the call sites identical and makes overrides checked (needs the type-only import added alongside).

ccip-sdk/src/cct/solana/token-admin-registry/operations/propose-admin.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  describe('generate', () => {
+    it('builds an unsigned propose admin instruction', async () => {
+      const unsigned = await generate()
+      const [instruction] = unsigned.instructions
+
+      assert.ok(instruction)
+      assert.equal(unsigned.family, ChainFamily.Solana)
+      assert.equal(unsigned.mainIndex, 0)
+      assert.equal(instruction.programId.toBase58(), ROUTER)
+      assert.deepEqual(instruction.data.subarray(-32), new PublicKey(NEW_ADMIN).toBuffer())
```
aelmanaa: The instruction assertion checks only the trailing 32 bytes (`newAdmin`). Asserting the 8-byte discriminator too pins the op to the actual Router entrypoint — `sha256('global:transfer_admin_role_token_admin_registry')[0..8]` = `b262cbb5cb6b6a0e` (chainlink-ccip `ccip-router/src/lib.rs:416`; devnet-verified byte-for-byte) — and `register-token.test.ts` already establishes the in-repo pattern with `createHash('sha256')`. Worth adding the account-metas assertion (config PDA, TAR PDA writable, mint, authority signer+writable per `token_context.rs:117-134`) in the same breath.

ccip-sdk/src/cct/solana/token-admin-registry/operations/propose-admin.test.ts
```diff
@@ -0,0 +1,103 @@
...
+      await assert.rejects(
+        () =>
+          SolanaTokenManager.fromChain(stubChain()).proposeAdmin({
+            tokenAddress: TOKEN,
+            address: ADDRESS,
+            newAdmin: NEW_ADMIN,
+            authority: CURRENT_ADMIN,
+            wallet: WALLET,
+          }),
+        (err: unknown) => err instanceof CCTParamsInvalidError && err.context.param === 'authority',
```
aelmanaa: This asserts only `err.context.param === 'authority'`, which is indistinguishable from the `buildUnsigned` preflight rejection two tests up. In this fixture `authority` equals the stub's `administrator`, so the `buildUnsigned` preflight would pass — if the execute-path wallet≠authority guard were deleted, the op would proceed to `submit()` and fail with a different error, not the expected `CCTParamsInvalidError`. Assert `err.context.reason` (the "requires authority to be the executing wallet…" text) to pin the branch actually under test.

ccip-sdk/src/cct/solana/token-admin-registry/operations/transfer-admin.ts
```diff
@@ -0,0 +1,126 @@
...
+  deriveRouterConfigPda,
+  deriveTokenAdminRegistryPda,
+} from '../../programs/router.ts'
+import { submit } from '../../submit.ts'
+import { validateAuthorityMatchesWallet, validatePublicKey } from '../../validate.ts'
+
+/** Parameters shared by Solana TokenAdminRegistry `proposeAdmin` generation and execution. */
+type ProposeAdminParams = {
+  tokenAddress: string
+  /**
```
aelmanaa, line 25:
```suggestion
  /**
   * CCIP contract to resolve the TokenAdminRegistry/Router from — a Router or OffRamp
   * address works.
   */
```

`SolanaChain.getTokenAdminRegistryFor` (`solana/index.ts:1185-1194`) resolves only a Router (as-is) or an OffRamp (→ its router); "the registry itself … or TokenPool" throws `CCIPContractNotRouterError`. Same overclaim was flagged on `register-token.ts`/`set-pool.ts` — trim here too (or widen the resolver once, for all three).

ccip-sdk/src/cct/solana/token-admin-registry/operations/transfer-admin.ts
```diff
@@ -0,0 +1,126 @@
...
+    opts: GenerateProposeAdminParams,
+  ): Promise<UnsignedSolanaTx> {
+    const tokenMint = new PublicKey(opts.tokenAddress)
+    const payer = new PublicKey(opts.payer)
+    const authority = new PublicKey(opts.authority ?? opts.payer)
+    const newAdmin = new PublicKey(opts.newAdmin)
+    const router = new PublicKey(await chain.getTokenAdminRegistryFor(opts.address))
+    const tokenConfig = await chain.getRegistryTokenConfig(router.toBase58(), tokenMint.toBase58())
+
+    if (!new PublicKey(tokenConfig.administrator).equals(authority)) {
```
aelmanaa, line 73:
```suggestion
    if (!new PublicKey(tokenConfig.administrator).equals(authority)) {
      const pending = tokenConfig.pendingAdministrator
      throw new CCTParamsInvalidError(
        this.name,
        'authority',
        PublicKey.default.toBase58() === tokenConfig.administrator && pending
          ? `registration for this token is still pending acceptance by ${pending}; the pending administrator must accept the admin role first — this operation only transfers an accepted role`
          : `must be the current token administrator (${tokenConfig.administrator})`,
      )
    }
```

The message is state-blind: it never says who the administrator *is*, and the most common way to hit it is the freshly-registered state (administrator = zero default, caller is the *pending* admin — chainlink-ccip `token_context.rs:264-265`/`:132`), where "must be the current token administrator" points the user at the wrong op entirely. I hit exactly this live on devnet. Branching the pending-registration case turns a dead-end error into the correct next step. (Exact wording/shape yours — the two states are the point.)

aelmanaa (changes requested):
I ran this op live on Solana devnet:

- Executed `proposeAdmin` end-to-end: tx [`5sQaftoHuU3cWekpZUri6kqhZVuYsLBYZZTKLjyTFfgpmQYqwxXffSmJ9VRaT9he2xGjWjATY2YYgbpbtDpgCH8m`](https://explorer.solana.com/tx/5sQaftoHuU3cWekpZUri6kqhZVuYsLBYZZTKLjyTFfgpmQYqwxXffSmJ9VRaT9he2xGjWjATY2YYgbpbtDpgCH8m?cluster=devnet). Read the registry PDA back two independent ways (SDK decode + raw `getAccountInfo` borsh-offset decode): `pending_administrator` == the proposed `newAdmin`, `administrator` unchanged. Accounts and data match the canonical `ModifyTokenAdminRegistry` context byte-for-byte.
- The client preflight mirrors the on-chain constraint (`authority == administrator @ Unauthorized`, `token_context.rs:132`) exactly — neither stricter nor looser — and both negative paths reject **client-side with typed errors before any transaction** (wrong-state authority; authority ≠ executing wallet). Nice.
- Type discipline, error taxonomy, layer boundaries, and the facade/barrel pattern all match the sibling ops; 4/4 new tests pass at head.

Two things I'd ask for before merge, both raised inline: the **rename to `transferAdminRole`** (the chain's own program log for this op's transaction is `Instruction: TransferAdminRoleTokenAdminRegistry`; "propose" is the registration verb everywhere else in CCIP — including our own bs58 generator), and the **journey docs** hook-up once DAPP-10605's accept op exists. On that second point, an operational note: to test this op at all I had to perform the accept instruction *outside* the SDK — after `registerToken` the registry's administrator is the zero default, so `proposeAdmin`/`transferAdminRole` has nothing it can act on until the accept op lands. The two PRs are really one journey; landing 10604 → 10605 in order (10605 rebased onto this head) and cross-linking their docs would close the register-token review's F1 completely.

(Not this PR's diff, but noting for the stack: `register-token.test.ts:8` on the base branch has an `import-x/order` warning — `errors.ts` import above the `../../../../` group; a `lint:fix` on whichever branch lands first clears it.)

ccip-sdk/src/cct/solana/index.ts
mervin-link: Renamed to `transferAdmin`.

aelmanaa (approved):
**✅ Approve — the rename landed correctly and completely; all prior blockers and nits resolved.**

Re-reviewed at `4771aef` (was `dd03030`). Every finding from the prior pass is fixed:

- **P1 (rename) ✅** — `transferAdmin` end-to-end: class `TransferAdmin`, facade `generateUnsignedTransferAdmin`/`transferAdmin`, op file `transfer-admin.ts`, tests. No residual `proposeAdmin` identifier in SDK code. Still wraps the unchanged `transferAdminRoleTokenAdminRegistry` instruction. Fresh devnet run under the new name is wire-identical (same discriminator `b262cbb5cb6b6a0e`, same on-chain log `Instruction: TransferAdminRoleTokenAdminRegistry`).
- **Comments 2–7 ✅** — journey docs wired (`@remarks`/`@see` to `acceptAdmin` + register↔accept↔setPool), test typing (`Partial<GenerateTransferAdminParams>`), discriminator + account-metas asserts, execute-path `reason` assert, address docstring trimmed, state-blind message branched (pending-registration case).
- **Import-order lint ✅** — `errors.ts` now inside the internal group; eslint clean.

**Live re-verification (devnet, 2026-07-29):** fresh mint → `registerAdmin` (owner) → raw accept → `transferAdmin` executed — on-chain `pending_administrator == newAdmin`, `administrator` unchanged, read back via SDK + raw borsh decode. Wire byte-identical to the pre-rename `dd03030` run.

**One non-blocking nit:** the `{@link acceptAdmin}` / `{@link generateUnsignedAcceptAdmin}` links are forward references — the accept op ships on DAPP-10605 (PR #317), so they resolve only once #313 → #317 land in the agreed order.

**CI note:** `npm ci` fails on the lockfile sync issue (`@emnapi/core@1.11.2` missing) — same repo-wide infra issue as #310/#317, not from this PR's code. Local `npm run check` is green with a clean install (prettier + eslint + tsc + 5/5 + 9/9 tests).

apedrob (approved):

PR #312, opened 2026-07-27
feat(cct-sdk): Authorize Lockbox Callers

apedrob:
## What

- Add `authorizeLockboxCallers` to EVMTokenManager; it adds/removes authorized callers on an ERC20LockBox

## Why
- After a LockReleaseTokenPool is deployed it can't lock/release until it's an authorized caller on its lockbox (deposit/withdraw are gated by the allowlist, not ownership)

## Testing
- Golden-vector calldata + validation + execute; typecheck, lint and CI tests green

PR #316, opened 2026-07-28
feat(cct-sdk): Add deploy verification + unify ops

apedrob:
## What
- Add block-explorer verification handle to EVM deploy results
- Unify 3 deploy ops under one base + a consistent per-family contract layer

## Why
- Etherscan verify needs the ABI-encoded ctor args
- Deploy ops had duplicated and inconsistent interface/bytecode/name sourcing

## Testing
- typecheck + prettier + eslint clean
- CI tests green
- Ran local playground on Base Sepolia; deployed + Basescan-verified a pool (`0x71ba2bfDBD3564D3C8434aA98EA70b07A5bf0fD6`) using the SDK verification output


aelmanaa (commented):
test

aelmanaa (commented):
test

aelmanaa (commented):
test

aelmanaa (commented):
test

ccip-sdk/src/cct/evm/lockbox/operations/authorize-callers.ts
```diff
@@ -46,12 +43,8 @@ export class AuthorizeLockboxCallers extends EVMOperation<AuthorizeLockboxCaller
...
       )
     }
-    const validateCaller = (field: string, c: string, i: number): void => {
-      validateAddress(this.name, `${field}[${i}]`, c)
-      if (c === ZeroAddress) {
-        throw new CCTParamsInvalidError(this.name, `${field}[${i}]`, 'must not be the zero address')
-      }
-    }
+    const validateCaller = (field: string, c: string, i: number): void =>
+      validateNonZeroAddress(this.name, `${field}[${i}]`, c)
```
aelmanaa, line 52: **[major — one-word diff]** This PR creates `validateNonZeroAddress` (`1bee3c2`) and applies it to `deployLockbox.token`, `deployTokenPool.lockbox`, and every `addedCallers[i]`/`removedCallers[i]` *in this very function* — but the `lockbox` **target** one line up is still plain `validateAddress`. That's the exact gap I live-demonstrated on #312: `lockbox: ZeroAddress` sails through, the SDK **sends the tx to `0x0` and reports success** — status 1, 0 logs, nothing authorized ([tx `0x21404580…`](https://sepolia.etherscan.io/tx/0x21404580d7f0ba22a1d2db42b1449e8dfca29ce4aceab4ab71d2e1a4d04048b9)), and the user's next pool op reverts `UnauthorizedCaller` with no clue why. Since this PR already rewrites this function, please fold it in:

```suggestion
    validateNonZeroAddress(this.name, 'lockbox', lockbox)
```

plus one rejection test next to the caller zero-check ones. (Unlike the caller checks — which mirror canonical `AuthorizedCallers.sol`'s `ZeroAddressNotAllowed` on adds — the chain can't protect this target: there's no contract at `0x0` at all.)

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -1,38 +1,81 @@
...
   return { family: ChainFamily.EVM, transactions: [{ data: bytecode + ctorArgs.slice(2) }] }
 }
 
+/** Assembles an unsigned call to an existing contract: `to` + ABI-encoded calldata. */
+export function callTx(to: string, data: string): UnsignedEVMTx {
+  return { family: ChainFamily.EVM, transactions: [{ to, data }] }
+}
+
+/** Block-explorer verification handle for a deployed contract: its name and ABI-encoded ctor args. */
+export interface DeployVerification {
```
aelmanaa: **[minor — docs]** The new handle never surfaces in any doc a consumer reads: the three op docstrings still say ``execute` resolves to `{ hash, contractAddress }`` (`deploy-lockbox.ts:24`, `deploy-token-pool.ts:89`, `deploy-token.ts:58`), and every facade `@example` destructures `const { hash, contractAddress } = …` with no mention of `verification` or what it's *for*. Suggest: update the three docstrings, and extend one facade example (e.g. `deployTokenPool`) to show the intended journey — `const { contractAddress, verification } = …` → paste `verification.encodedConstructorArgs` into the explorer's "Constructor Arguments" field (Etherscan expects it without the `0x`, worth saying). I live-verified the handle is exactly what Basescan wants (your `0x71ba2bfD…` pool shows 320 hex chars of ctor args matching the handle format), so this is purely a discoverability gap on a feature that works.

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -66,3 +109,46 @@ export abstract class EVMOperation<P extends { sender?: string }> extends Operat
...
+    return deployTx(a.bytecode, this.encode(a.iface, params))
+  }
+
+  /**
+   * {@link generate}, then sign and submit; resolves to the tx hash and the newly deployed
+   * contract address (read from the mined receipt), plus a {@link DeployVerification} handle
+   * recovered from the init-code.
+   * @throws {@link CCTTxFailedError} if the tx mined without producing a contract address
+   */
+  override async execute(chain: EVMChain, params: EVMExecuteParams<P>): Promise<DeployResult> {
```
aelmanaa, line 143: **[minor — suggestion, not blocking]** Naming note first: `DeployVerification` reads like "the SDK verified the deploy", but `execute` performs no post-deploy reads — it's an **explorer-verification handle** (great!), recovered offline from the init-code. Two thoughts:

1. Now that all three deploy ops share this one `execute`, a single `typeAndVersion()` `eth_call` against `receipt.contractAddress` here would give every deploy op a built-in on-chain self-check (the exact read-back my live reviews do manually after each deploy) — one RPC, closes the "deploy succeeded but deployed the wrong thing / nothing" class for the whole family. Happy to see it as a follow-up.
2. The handle alone can't complete an explorer verification — that also needs the source/compiler side (standard-json from `@chainlink/contracts-ccip@2.0.0`, matching solc settings). One `@remarks` line on `DeployVerification` saying where the sources live would make the handle actionable without spelunking. (Typing nit: `encodedConstructorArgs` could be `` `0x${string}` `` to match `bytecode`.)

ccip-sdk/src/cct/evm/submit.ts
```diff
@@ -51,7 +51,9 @@ export async function submit(
   let response: TransactionResponse
   let nonceConsumed = false
   try {
-    let tx: TransactionRequest = { ...unsigned.transactions[0]! }
+    const [first] = unsigned.transactions
+    if (!first) throw new CCTTxFailedError(operation, 'no transaction to submit')
```
aelmanaa: **[nit]** Nice catch replacing the `transactions[0]!` assertion — but the guard throws *inside* the `try`, so its `CCTTxFailedError('no transaction to submit')` is immediately caught by the surrounding `catch` and re-wrapped in a second `CCTTxFailedError` (message survives; the type is now its own `cause`). Hoisting the two lines above the `try` keeps it single-wrapped and skips the pointless `rollbackNonce` bookkeeping path.

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -1,38 +1,81 @@
...
+ * Recovers the verification handle from a deployment's init-code with no extra RPC.
+ * {@link deployTx} builds `data = bytecode + ctorArgs.slice(2)`, so slicing off
+ * `bytecode.length` chars recovers the (0x-prefixed) ABI-encoded constructor args.
+ */
+export function buildDeployVerification(
+  contract: string,
+  deployData: string,
+  bytecode: string,
+): DeployVerification {
+  return { contract, encodedConstructorArgs: `0x${deployData.slice(bytecode.length)}` }
```
aelmanaa: **[major — scope/completeness]** First, the structural point: **once you have the handle, the SDK offers no way to actually verify.** There is no submission surface at all — no Etherscan/Sourcify/Blockscout client, no `verifyContract`, no `verify/` module (`git grep` for any of these across `ccip-sdk/src` + `ccip-cli/src` at this head returns nothing; the `verify/` directory has 0 files). `DeployResult.verification` answers *"what are the constructor args?"*, not *"is my contract verified?"* — the developer is left to source std-json + compiler version + FQN elsewhere and POST to each explorer by hand.

I live-tested exactly how far the handle gets you, using only what the SDK provides (bare contract name + `encodedConstructorArgs`; the SDK vendors ABI/bytecode but no sources, no compiler version, no settings, no qualified `path/File.sol:Name`):

- **Etherscan v2** `verifysourcecode` → `NOTOK: Missing sourceCode` (the request can't even be formed);
- **Blockscout** standard-input API → `Bad request` (`files[]` mandatory);
- **Sourcify** v2 → `must have required property 'stdJsonInput'`.

So today the handle only serves the *manual* Etherscan form — and even there the user must source the std-json, compiler version and qualified name elsewhere. (Both my Sepolia deploys from this head showed as "verified" on Etherscan purely via **SimilarMatch** before any submission — Blockscout and Sourcify give nothing for free.) The missing inputs are entirely available: `@chainlink/contracts-ccip@2.0.0` on npm ships all 118 `.sol` sources + `remappings.txt` + `foundry.toml` — I built a standard-json from the npm package contents alone and it fully verified my `LockReleaseTokenPool` on Sourcify (`creationMatch: match`, which then auto-propagated to Etherscan/Routescan/Blockscout), and a pre-built per-contract standard-input took my `ERC20LockBox` to `Pass - Verified` on Etherscan and verified on Blockscout.

**You don't have to design this from scratch — it's already built** on the org branch [`smartcontractkit/feat/propose-admin-role`](https://github.com/smartcontractkit/ccip-tools-ts/tree/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify) (PR #204, @ `1cecb26`), and its `ERC20LockBox` bytecode is byte-identical to this PR's, so the same fixtures verify #316's deploys exactly. Concretely, either:

1. grow the handle into the full input set the three APIs need — every piece below is already implemented on that branch and can be lifted directly:
   - **standard-json-input bundling** (sources + settings that produced the bytecode) — the vendored per-contract fixtures + a manifest keyed by contract name: [`fixtures/manifest.json`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/fixtures/manifest.json), a sample [`fixtures/ERC20LockBox.standard-input.json`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/fixtures/ERC20LockBox.standard-input.json), loaded by [`getVerificationArtifact` (`registry.ts`)](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/registry.ts);
   - **long `compilerVersion` resolver** (short `0.8.26` → `v0.8.26+commit.8a97fa7a`, offline pin map + solc-list fallback) — [`solc-version.ts`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/solc-version.ts);
   - **fully-qualified `contract`** (`contracts/pools/ERC20LockBox.sol:ERC20LockBox`) — the `contractName` field in [`fixtures/manifest.json`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/fixtures/manifest.json);
   - **constructor-args encoder** (ABI-encode, no `0x`, no selector — Etherscan's exact form) — [`constructor-args.ts`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/constructor-args.ts);
   - **Etherscan v2 submission client** (submit → GUID poll → already-verified short-circuit → slow-explorer `getsourcecode` confirm) — [`etherscan.ts`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/etherscan.ts);
   - **Sourcify submission** (v2 JSON API, `verificationId` poll, creation-tx match) — [`sourcify.ts`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/sourcify.ts);
   - **Blockscout support + per-chain verifier routing** (67-chain map to etherscan-v2/blockscout/sourcify) — [`registry.ts`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/registry.ts) over [`fixtures/verifiers.json`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/fixtures/verifiers.json);
   - **high-level orchestrator** that ties it together (`verifyContract` / `verifyDeployedContract`) — [`verify.ts`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/verify.ts);
   - **CLI `verify` command + `--verify` on deploy** (verify an already-deployed contract, deriving ctor args from on-chain creation code, incl. factory **CREATE2 recovery via `debug_traceTransaction`**) — [`ccip-cli/.../verify.ts`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-cli/src/commands/verify.ts) + [`verify-utils.ts`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-cli/src/commands/verify-utils.ts).

   The `{ standardJsonInput, compilerVersion, contract(FQN) }` tuple is the complete input set for all three APIs; the clients above then submit it. **Or**
2. add one `@remarks` on `DeployVerification` stating the scope explicitly: "constructor-args companion for explorer verification; sources/compiler settings must come from `@chainlink/contracts-ccip@2.0.0` (npm ships them)" — so consumers don't discover the gap the way I did.

Severity is "major" against a multi-explorer support goal; if the intended scope is Etherscan-manual-only, option 2 downgrades this to docs.

ccip-sdk/src/cct/evm/index.ts
```diff
@@ -332,5 +332,10 @@ export type {
 } from './token-pool/operations/deploy-token-pool.ts'
 export type { DeployLockboxParams } from './lockbox/operations/deploy-lockbox.ts'
 export type { AuthorizeLockboxCallersParams } from './lockbox/operations/authorize-callers.ts'
-export type { DeployResult, EVMExecuteParams } from './operation.ts'
+export type {
```
aelmanaa, line 338: **[major — asymmetry]** `execute` returns the full `DeployResult` with the handle (`operation.ts:138-153`: `{ hash, contractAddress, verification: { contract, encodedConstructorArgs } }`), but the `generateUnsigned*` deploy facades return a **bare** `Promise<UnsignedEVMTx>` with **no** verification handle: `generateUnsignedDeployToken` (`index.ts:152`), `generateUnsignedDeployTokenPool` (`index.ts:206`), `generateUnsignedDeployLockbox` (`index.ts:257`). `generate` (`operation.ts:93-99`) only ever returns the unsigned tx.

That drops verification on the one workflow `generateUnsigned*` exists for: **generate unsigned deploy → route to a Safe/Squads multisig → deploy → learn the address → verify.** A consumer on that path has no way to obtain the handle — `buildDeployVerification` (`operation.ts:42-49`) is address-independent and fully derivable *without* deploying (it just slices the init-code by `bytecode.length` and pairs it with the artifact name), but `bytecode` lives in `protected artifact()`, so a generate-path caller can't reconstruct it cleanly. So the handle is available for free on the path that *already* returns the address (`execute`) and unavailable on the path that *can't* (`generate`) — backwards.

Fix is pure, no chain/address needed — expose the handle on the generate side. Either a facade accessor:

```ts
// alongside generateUnsignedDeployTokenPool(...)
deployVerification(opts: DeployTokenPoolParams): DeployVerification {
  // delegates to the op below
}
```

or, cleaner, a public method on `EVMDeployOperation` that reuses what `execute` already does:

```ts
// EVMDeployOperation
deployVerification(params: P): DeployVerification {
  const { contract, bytecode } = this.artifact(params)
  return buildDeployVerification(contract, this.buildUnsigned(/* chain-free init-code */).data, bytecode)
}
```

`DeployResult` already proves the value is correct (my live Sepolia run matched it against on-chain `tx.data`); this just makes `generate` symmetric with `execute` so the multisig flow isn't a second-class citizen. (Pairs with ⑥: this exposes the ctor-args handle everywhere; ⑥ is about completing the *rest* of the explorer-submission inputs.)

ccip-sdk/src/cct/evm/token-pool/contracts.ts
```diff
@@ -150,6 +156,38 @@ export function getTokenPoolInterface(type: TokenPoolType, version: TokenPoolVer
   return TOKEN_POOL_INTERFACES[getTokenPoolFamily(type)][version]
 }
 
+/**
+ * Creation bytecode per deployable pool type (2.0.0 only — pre-2.0.0 bytecode is not vendored).
+ * The keys define the deployable set ({@link DeployableTokenPoolType}). The burn-* variants share
+ * the `BurnMint` constructor ABI but are distinct contracts with distinct bytecode.
+ */
+const TOKEN_POOL_BYTECODE = {
```
aelmanaa, line 164: **[major — coverage gap, shared with the reference branch]** This PR's deployable set (`TOKEN_POOL_BYTECODE` keys at `contracts.ts:164-172`, plus `CrossChainToken` and `ERC20LockBox`) is **six** contracts: `CrossChainToken`, `ERC20LockBox`, `BurnMintTokenPool`, **`BurnFromMintTokenPool`**, **`BurnWithFromMintTokenPool`**, `LockReleaseTokenPool`. When the source/compiler side lands (⑥), a verification submission needs a per-contract `standard-input.json`. The reference branch [`smartcontractkit/feat/propose-admin-role`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify/fixtures/manifest.json) (PR #204) ships fixtures for only 6 contracts — `CrossChainToken`, `ERC20LockBox`, `LockReleaseTokenPool`, `BurnMintTokenPool`, `CrossChainPoolToken`, `AdvancedPoolHooks` — so **`BurnFromMintTokenPool` and `BurnWithFromMintTokenPool` have no std-json fixture on *either* side.** Your own note at `contracts.ts:162` is the reason this bites: the burn-* variants "share the `BurnMint` constructor ABI but are **distinct contracts with distinct bytecode**." So `BurnMintTokenPool.standard-input.json` will **not** verify a `BurnFromMintTokenPool` deploy — the recompiled creation bytecode won't match. Net effect: even lifting the whole branch, **2 of your 6 deployables stay unverifiable** until fixtures are generated for them (compile `BurnFromMintTokenPool`/`BurnWithFromMintTokenPool` from `@chainlink/contracts-ccip@2.0.0` with the same solc settings and add two `*.standard-input.json` + manifest entries). Your `0x71ba2bfD…` Basescan proof was a `BurnFromMintTokenPool` verified via Etherscan **SimilarMatch**, not a fixture-driven submission — which is exactly why the missing fixture went unnoticed.

**Settings-fidelity rider:** whatever fixtures land must pin the exact optimizer settings — `@chainlink/contracts-ccip@2.0.0`'s `foundry.toml` release profile says `runs: 80000` while the branch fixtures and the Etherscan similar-match record say `50000` (both reproduce byte-identical bytecode in my tests, so the value isn't self-evident from the artifact). The branch pins `runs:50000, viaIR:true, paris, bytecodeHash:none` inside each std-json; #316 records no settings at all. Pin them explicitly rather than leave verifiers guessing — a wrong guess produces a non-matching recompile and a confusing "bytecode mismatch" failure.

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -1,38 +1,81 @@
...
+export interface DeployVerification {
+  contract: string
+  encodedConstructorArgs: string
+}
+
+/**
+ * Recovers the verification handle from a deployment's init-code with no extra RPC.
+ * {@link deployTx} builds `data = bytecode + ctorArgs.slice(2)`, so slicing off
+ * `bytecode.length` chars recovers the (0x-prefixed) ABI-encoded constructor args.
+ */
```
aelmanaa, line 48: **[major — design direction]** Flagging this now because I understand the **factory deploy path is planned**, and it's the deciding factor for how verification should be shaped. `buildDeployVerification` (`operation.ts:42-49`) recovers the constructor args by slicing the **outer transaction's `data`** by `bytecode.length`. That's exact for a *direct* `CREATE` (which is all #316 does today — I proved it live), so there's no bug as it stands. But for a **factory / CREATE2** deploy the outer `tx.data` is the *factory call*, and the contract's real init-code lives in an **inner** CREATE frame — so the same slice would return a wrong `encodedConstructorArgs` **with no error**. The deploy-time-handle model therefore can't be *extended* to factory verification; it has to be replaced there. The address-based verifier already handles it by recovering the inner init-code via `debug_traceTransaction` ([`verify-utils.ts` on `feat/propose-admin-role`](https://github.com/smartcontractkit/ccip-tools-ts/blob/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-cli/src/commands/verify-utils.ts)). So this is the strongest single reason to adopt the address-based verifier as the primary path *now* (see the Design-recommendation in the general comment) rather than ship the handle and discover it can't carry factory deploys later. Not a blocker for merging #316's direct-deploy scope — but please don't build further verification on the slice-the-outer-tx model, since the next deploy op invalidates it.

aelmanaa (approved):
**Approve with fixes — the unification is real and non-regressive, and the verification handle is live-proven.** I re-ran the whole thing on Sepolia at this head: `deployLockbox` through the new `EVMDeployOperation` deployed [`0xC645befF…55B1`](https://sepolia.etherscan.io/address/0xC645befF632b9933B476F5ABEfbF1c773e5255B1) ([tx `0xa98116b2…`](https://sepolia.etherscan.io/tx/0xa98116b25583a041365a2cbbb680776e7d843733a0b8d52ebe6b83e02633de6c)), and the returned `verification.encodedConstructorArgs` matched **both** an independently computed `abi.encode([token])` **and** the on-chain `tx.data` tail after the vendored bytecode — the slice-recovery math holds against real chain data, and the zero-token rejection still fires client-side with the exact pre-refactor message (nonce unchanged). I also confirmed the PR-body claim: your Basescan pool `0x71ba2bfD…` is verified (`BurnFromMintTokenPool`), i.e. the handle drives a real explorer verification.

What I like: the three copy-pasted deploy `execute`s collapse into one audited path; the `contracts.ts` layer is now symmetric across token/token-pool/lockbox; every public surface approved on #307/#311/#312 keeps its exact shape, validators and messages (I checked param-by-param and live); and the per-type verification tests assert against the hand-encoded golden vectors rather than the SDK encoder.

Three asks before merge, in order: **(①)** the `authorizeLockboxCallers` `lockbox` zero-check — this PR built the exact helper and applied it everywhere *except* the one target I live-proved produces a status-1 tx to `0x0`; it's now a one-word diff. **(⑤)** merge the updated base so the head stops failing `npm run check` on inherited Solana staleness. **(②)** make the new `verification` field visible in the docs/examples — it currently ships invisible. One framing note: this is *explorer*-verification support (valuable!), not deployability verification — `execute` does no post-deploy reads; with the shared deploy base you now have, a single `typeAndVersion` read-back would give every deploy op the on-chain self-check my reviews keep doing manually (③, follow-up welcome).

Landing order note: this stacks 10545 → 10738 → 10788 → 10572 (#307 → #311 → #312 → #316); #311's `// generate:end` artifact-gate fix and #312's items travel with their own PRs — nothing here supersedes them, and #312's zero-`lockbox` comment is *restated* (not mooted) by this PR.

**Addendum (multi-explorer follow-up):** I subsequently exercised the handle against all three EVM explorer verification APIs (Etherscan v2, Blockscout, Sourcify) — with only SDK-provided data, none of the three requests can be formed (details + live transcripts in ⑥). The handle is a fine constructor-args companion, but "deploy verification" currently rides on Etherscan SimilarMatch; completing it needs the standard-json/compiler side, which the `@chainlink/contracts-ccip` npm package already ships (live-proven end-to-end on Sepolia against all three explorers). A **complete reference implementation** of the missing pieces (Etherscan v2 / Sourcify / Blockscout clients, per-chain verifier routing, solc long-version resolver, bundled std-json fixtures + manifest, and a CLI `verify` command with CREATE2 recovery) already exists on the org branch [`smartcontractkit/feat/propose-admin-role` @ `1cecb26`](https://github.com/smartcontractkit/ccip-tools-ts/tree/1cecb266d924bb52c6c6030e0bdc2b4b6f5e430e/ccip-sdk/src/verify) (PR #204) and can be lifted directly — see ⑥ for the file-by-file map and the two ways to close this. And ⑦: the ctor-args handle it *does* build is dropped on the `generateUnsigned*` path (the async/multisig deploy flow), which is a one-accessor fix independent of ⑥.

**Coverage caveat (⑧):** I ran a full completeness audit of the deployable set. Two of this PR's six deployables — `BurnFromMintTokenPool` and `BurnWithFromMintTokenPool` — have **no std-json fixture on either this PR or the reference branch**, and since each burn-* variant has distinct bytecode (`contracts.ts:162`), `BurnMintTokenPool`'s fixture can't verify them. So even adopting the branch, those two stay unverifiable until fixtures are generated (details + settings-pin note in ⑧). The audit otherwise confirmed ⑥/⑦ are the only design gaps and that the init-code slice is **correct** for every #316 deploy (all are direct CREATE — there's no factory/CREATE2 path in this PR's scope, so no silent-slice risk).

**On direction (DX):** verification is inherently a post-deploy, address-first, retryable operation, so my recommendation is to make the **address-based verifier** the primary path (derive ctor args from the deployed address / creation tx, incl. CREATE2, submit full std-json across all four providers — already built on `feat/propose-admin-role`, PR #204) and keep this PR's inline handle as an **optional zero-RPC fast-path input** into it when you deploy directly. That closes the two verification asks together and avoids two half-verification stories. Honest cost of the address model: a `debug_traceTransaction` dependency for factory/CREATE2 (gated on some RPCs) — worth weighing, not disqualifying.

ccip-sdk/src/cct/evm/operation.ts
apedrob: Renamed

ccip-sdk/src/cct/evm/operation.ts
apedrob: Took option 2. I renamed it to `ExplorerVerificationInput` since `execute` never does a
post-deploy check, and added an `@remarks` saying it's just the constructor args, with
sources and compiler settings coming from `@chainlink/contracts-ccip`. The op docstrings
and facades now mention `verification`, so by your own framing this is docs now.

On the rest, I've also created my own verification that I shared with you, so we agree
it's wanted. Shouldn't this be handled at the CLI level though? `@chainlink/contracts-ccip`
is only a root devDependency, not a dependency of `ccip-sdk`, which vendors the generated
ABI/bytecode instead. Putting std-json in the SDK means vendoring all `.sol` files or
adding heavy runtime dep. And even if we want to use the Etherscan API, an API key is
needed, so credentials and polling fit better in the CLI. Your reference implementation
already puts the command there.

Can you create a Jira ticket with the requirements for what the SDK needs to expose to
feed a CLI verify command, so we can plan it?

ccip-sdk/src/cct/evm/index.ts
apedrob: Documented for now: the three `generateUnsigned*` docstrings say the path
returns the unsigned tx only, with no `verification`

apedrob: Also  I dropped `buildDeployVerification` and take the args straight from `encode()` to be more DRY

ccip-sdk/src/cct/evm/token-pool/contracts.ts
apedrob: There are no std-json fixtures in this repo, so nothing to change here. It's good input for
the verify ticket though

ccip-sdk/src/cct/evm/operation.ts
apedrob: Agreed on not building further on the slice, so I removed it. The constructor args now come straight from `encode()` instead of being sliced back out of the init-code, so there's no dependency on the outer tx layout and the factory case can't silently return wrong args. Whether the factory path goes address-based can be decided in the ticket.

mervin-link (approved):

PR #317, opened 2026-07-28
feat(cct-sdk): Add accept admin solana op

mervin-link:
## What

- [DAPP-10605](https://smartcontract-it.atlassian.net/browse/DAPP-10605)
- Add Solana CCT `acceptAdmin` and `generateUnsignedAcceptAdmin` TokenAdminRegistry operations
- Validate current pending authority before building `acceptAdminRoleTokenAdminRegistry` 

## Why

- Enable pending TokenAdminRegistry administrators to accept the role through the SDK

[DAPP-10605]: https://smartcontract-it.atlassian.net/browse/DAPP-10605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/token-admin-registry/operations/accept-admin.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (
+      !tokenConfig.pendingAdministrator ||
+      !new PublicKey(tokenConfig.pendingAdministrator).equals(authority)
+    ) {
+      throw new CCTParamsInvalidError(
+        this.name,
+        'authority',
+        'must be the pending token administrator',
+      )
+    }
```
aelmanaa, line 82:
```suggestion
    if (!tokenConfig.pendingAdministrator) {
      throw new CCTParamsInvalidError(
        this.name,
        'authority',
        `no administrator is pending for this token (current administrator: ${tokenConfig.administrator}) — nothing to accept`,
      )
    }
    if (!new PublicKey(tokenConfig.pendingAdministrator).equals(authority)) {
      throw new CCTParamsInvalidError(
        this.name,
        'authority',
        `must be the pending token administrator (${tokenConfig.pendingAdministrator})`,
      )
    }
```

This one message covers two different states, and one of them is the state where the message is actively wrong: "no pending admin" is exactly what an **idempotent retry after a successful accept** hits. I ran that live on devnet from this head (merged into the 10604 base): right after `acceptAdmin` succeeded (tx [`2Uendc85994sVP3K5wgy2DJpGhSj9epRFgfHan4ndjRVYhgofhj4jFssvXs2orDn6hJKfbkjQqcPYiSwAVT6bQiX`](https://explorer.solana.com/tx/2Uendc85994sVP3K5wgy2DJpGhSj9epRFgfHan4ndjRVYhgofhj4jFssvXs2orDn6hJKfbkjQqcPYiSwAVT6bQiX?cluster=devnet)), a second call told the wallet that *is now the administrator* it "must be the pending token administrator" — byte-identical to the rejection a genuinely-wrong wallet gets (I reproduced both: retry on the fresh mint vs. a registry whose pending is a different key). On-chain both states collapse into `Unauthorized` (`token_context.rs:194` gates only on `pending_administrator`), but client-side we've already fetched the registry and *know* which state we're in — that's the whole value of preflighting. Branching also gives the retry case its natural reading: "nothing to accept" ≈ already done. (Exact wording yours; the two states are the point. Please also assert the two `context.reason`s distinctly in the tests — see the test comment.)

ccip-sdk/src/cct/solana/token-admin-registry/operations/accept-admin.ts
```diff
@@ -0,0 +1,125 @@
...
+  deriveRouterConfigPda,
+  deriveTokenAdminRegistryPda,
+} from '../../programs/router.ts'
+import { submit } from '../../submit.ts'
+import { validateAuthorityMatchesWallet, validatePublicKey } from '../../validate.ts'
+
+/** Parameters shared by Solana TokenAdminRegistry `acceptAdmin` generation and execution. */
+type AcceptAdminParams = {
+  tokenAddress: string
+  /**
```
aelmanaa, line 25:
```suggestion
  /**
   * CCIP contract to resolve the TokenAdminRegistry/Router from — a Router or OffRamp
   * address works.
   */
```

`SolanaChain.getTokenAdminRegistryFor` (`solana/index.ts:1185-1194`) resolves only a Router (as-is) or an OffRamp (→ its router); "the registry itself … or TokenPool" throws `CCIPContractNotRouterError`. Same overclaim already flagged on `register-token.ts` / `set-pool.ts` / `propose-admin.ts` — trim here too (or widen the resolver once, for all four).

ccip-sdk/src/cct/solana/token-admin-registry/operations/accept-admin.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    connection: {},
+    getTokenAdminRegistryFor: async (address: string) => {
+      onAddress?.(address)
+      return ROUTER
+    },
+    getRegistryTokenConfig: async () => ({ administrator: PAYER, pendingAdministrator }),
+  } as unknown as SolanaChain
+}
+
+function generate(opts = {}) {
```
aelmanaa:
```suggestion
function generate(opts: Partial<GenerateAcceptAdminParams> = {}) {
```

`opts = {}` types the overrides as `{}`, so a typo'd key compiles silently. `Partial<GenerateAcceptAdminParams>` keeps call sites identical and makes overrides checked (add the type-only import alongside). Same nit as on the propose-admin test.

ccip-sdk/src/cct/solana/token-admin-registry/operations/accept-admin.test.ts
```diff
@@ -0,0 +1,110 @@
...
+describe('AcceptAdmin (cct/solana)', () => {
+  describe('generate', () => {
+    it('builds an unsigned accept admin instruction', async () => {
+      const unsigned = await generate()
+      const [instruction] = unsigned.instructions
+
+      assert.ok(instruction)
+      assert.equal(unsigned.family, ChainFamily.Solana)
+      assert.equal(unsigned.mainIndex, 0)
+      assert.equal(instruction.programId.toBase58(), ROUTER)
```
aelmanaa, line 57: The instruction assertion checks only `programId`. Asserting the 8-byte discriminator pins the op to the actual Router entrypoint — `sha256('global:accept_admin_role_token_admin_registry')[0..8]` = `6af010ad89d5a3f6` (chainlink-ccip `ccip-router/src/lib.rs:400`; I verified byte-for-byte on devnet — the landed ix data is exactly those 8 bytes, no args) — and `register-token.test.ts` already establishes the in-repo `createHash('sha256')` pattern. Worth asserting the account metas in the same breath: config PDA (RO), TAR PDA (writable), mint (RO), authority (signer+writable), per the canonical context (`token_context.rs:179-196`).

ccip-sdk/src/cct/solana/token-admin-registry/operations/accept-admin.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  describe('validation', () => {
+    it('rejects an authority that is not the pending administrator', async () => {
+      await assert.rejects(
+        () => generate({ authority: PAYER }),
+        (err: unknown) => err instanceof CCTParamsInvalidError && err.context.param === 'authority',
+      )
+    })
+
+    it('rejects when no administrator is pending', async () => {
```
aelmanaa, line 110:
```suggestion
    it('rejects when no administrator is pending', async () => {
      const noPendingChain = {
        logger: { debug() {}, info() {}, warn() {}, error() {} },
        connection: {},
        getTokenAdminRegistryFor: async () => ROUTER,
        getRegistryTokenConfig: async () => ({ administrator: PAYER }),
      } as unknown as SolanaChain
      await assert.rejects(
        () =>
          SolanaTokenManager.fromChain(noPendingChain).generateUnsignedAcceptAdmin({
            tokenAddress: TOKEN,
            address: ADDRESS,
            payer: PAYER,
          }),
        (err: unknown) => err instanceof CCTParamsInvalidError && err.context.param === 'authority',
      )
    })
```

`stubChain(undefined)` doesn't produce a no-pending registry: JS default parameters fire on explicit `undefined`, so the stub returns `pendingAdministrator = PENDING_ADMIN` and the guard rejects via the **authority-mismatch** branch — same error class and param, so the test passes for the wrong reason. I proved it empirically at this head: splitting the guard's two branches into distinct sentinel messages, this test rejects with the *mismatch* sentinel — and still passes. It would keep passing even if the `!pendingAdministrator` branch were deleted. The suggested stub mirrors what production actually returns for a no-pending registry: `getRegistryTokenConfig` (`solana/index.ts:1544-1551`) **omits** the `pendingAdministrator` key when the on-chain field is the zero default (the handler zeroes it on accept, `token_admin_registry.rs:124`). Once the messages are branched (see the accept-admin.ts comment), asserting `err.context.reason` here locks the branch in for good.

ccip-sdk/src/cct/solana/token-admin-registry/operations/accept-admin.test.ts
```diff
@@ -0,0 +1,110 @@
...
+            address: ADDRESS,
+            payer: PAYER,
+          }),
+        (err: unknown) => err instanceof CCTParamsInvalidError && err.context.param === 'authority',
+      )
+    })
+  })
+
+  describe('execute', () => {
+    it('requires the pending admin to be the executing wallet', async () => {
```
aelmanaa, line 135: This asserts only `err.context.param === 'authority'`, indistinguishable from the `buildUnsigned` preflight rejection above — if the execute-path wallet≠authority guard were deleted, this test would still pass via the preflight (which runs first inside `execute`). Assert `err.context.reason` (the "requires authority to be the executing wallet…" text) to pin the branch actually under test. Same nit as on the propose-admin test.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -355,6 +361,48 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
...
+   * @example
+   * ```ts
+   * const unsigned = await cct.generateUnsignedAcceptAdmin({
+   *   tokenAddress: mint,
+   *   address: router,
+   *   payer: pendingAdmin,
+   * })
+   * ```
+   */
+  generateUnsignedAcceptAdmin(opts: GenerateAcceptAdminParams): Promise<GenerateAcceptAdminResult> {
```
aelmanaa, line 457: This op is the missing rung the register-token review identified — I ran the full journey live from this PR's code merged into its base: `registerToken` → **`acceptAdmin`** (tx [`2Uendc85994sVP3K5wgy2DJpGhSj9epRFgfHan4ndjRVYhgofhj4jFssvXs2orDn6hJKfbkjQqcPYiSwAVT6bQiX`](https://explorer.solana.com/tx/2Uendc85994sVP3K5wgy2DJpGhSj9epRFgfHan4ndjRVYhgofhj4jFssvXs2orDn6hJKfbkjQqcPYiSwAVT6bQiX?cluster=devnet)) → `setPool` **succeeds** (tx [`2UqWZfapjayYkRSfAQyg9uu7ChUdgkjhdwVkH3fEBN7hewqcawEM1Acmwpk7hTS1dUsCEaVG7yTyJqXtAPyAVS3U`](https://explorer.solana.com/tx/2UqWZfapjayYkRSfAQyg9uu7ChUdgkjhdwVkH3fEBN7hewqcawEM1Acmwpk7hTS1dUsCEaVG7yTyJqXtAPyAVS3U?cluster=devnet)) — where the #310 review's identical `setPool` call died on-chain with `Unauthorized (7000)` because no accept op existed. But a reader can't discover that ordering from the docs yet: nothing here says acceptAdmin is the *required* step between `registerToken` and `setPool`, and there's no `@see` in either direction. Please add `@remarks` + bidirectional `@see` (registerToken ↔ acceptAdmin ↔ setPool, and proposeAdmin/transferAdmin ↔ acceptAdmin) — that's the documentation half of the register-token F1 fix. Two smaller doc asks in the same pass: note that `authority` defaults to the wallet/payer and that Squads/vault flows should use `generateUnsignedAcceptAdmin` (the `setPool` docs already carry this guidance), and consider `@remarks` for the idempotent-retry behavior once the A3 message lands.

aelmanaa (approved):
I ran this op live on Solana devnet from an isolated build:

- deployToken [`3NSLt4D1…c77KW→`](https://explorer.solana.com/tx/3NSLt4D19icHDkdF5CWZeddXspXKF6bd8pBxdMjQEEMr1YVDNkpuSJ12VrNPGA2qP7qp5W67rBcnz2BwoySzFwh2?cluster=devnet) deployTokenPool → createLookupTable → registerToken(owner) [`2EaqH4xV…rCrs`](https://explorer.solana.com/tx/2EaqH4xVvD5QvELWo3197uict3KRwLbXPqVbgajUhZudiid4htaaa3cQE3t5NHqjbvDuZjWcRzjdrh4NREZdrCrs?cluster=devnet) (registry: administrator = zero default, pending = wallet) → **acceptAdmin** [`2Uendc85…bQiX`](https://explorer.solana.com/tx/2Uendc85994sVP3K5wgy2DJpGhSj9epRFgfHan4ndjRVYhgofhj4jFssvXs2orDn6hJKfbkjQqcPYiSwAVT6bQiX?cluster=devnet) → **setPool succeeds** [`2UqWZfap…VS3U`](https://explorer.solana.com/tx/2UqWZfapjayYkRSfAQyg9uu7ChUdgkjhdwVkH3fEBN7hewqcawEM1Acmwpk7hTS1dUsCEaVG7yTyJqXtAPyAVS3U?cluster=devnet) — the exact call that reverted `Unauthorized (7000)` in the register-token review before this op existed.
- **Wire-exact:** the landed instruction data is exactly `6af010ad89d5a3f6` = sha256(`global:accept_admin_role_token_admin_registry`)[0..8], no args; program log `Instruction: AcceptAdminRoleTokenAdminRegistry`; accounts (config RO, registry writable, mint RO, authority signer+writable) match the canonical `AcceptAdminRoleTokenAdminRegistry` context (`token_context.rs:179-196`) and what `ccip-solana-bs58-generator`'s `accept-admin-role` builds. Read the registry PDA back two independent ways (SDK decode + raw borsh-offset decode): administrator == wallet, pending zeroed — the handler's exact promotion semantics (`token_admin_registry.rs:115-132`).
- **Both accept paths proven:** accept-after-register (admin = zeros) *and* accept-after-transfer (admin non-zero, via a self-`proposeAdmin` then a second accept, tx [`oEExtfeJ…Vb8n`](https://explorer.solana.com/tx/oEExtfeJjtFuKaV8d6foKdzcDRzxpHaZKGhNJfYjhaKr1oZELNjhCNd2LAUrYBSCetgWAP6x6ezTViKbEGRVb8n?cluster=devnet)) — the preflight gates on `pending == authority` only, exactly like the chain (never stricter, never looser), and the registry's `lookup_table` survives the accept untouched, as the handler promises.
- **Isolation semantics are right and typed, client-side before any tx:** unregistered mint → `CCIPTokenNotConfiguredError`; no pending / wrong wallet → `CCTParamsInvalidError`. Notably better than our bs58 generator, which has no preflight at all for this op.
- **Naming: 👍.** `acceptAdmin` is the one op of the trio that already matches the agreed cross-family scheme (`registerAdmin` / `transferAdmin` / `acceptAdmin`) — nothing to rename here.

Two asks before merge, both raised inline with suggestions: **A1** (the "no pending admin" test rejects via the wrong branch — proven with a sentinel-message experiment at this head) and **A3** (one message for two different rejection states; the idempotent-retry case reads wrong — hit it live). On stacking: since #313 will rename `proposeAdmin` → `transferAdmin` on the base branch, and this PR's added lines never reference the propose op, the rename only shuffles adjacent import/barrel lines — landing order #313-rename → #317 keeps both merges trivial. 5/5 new tests pass at the merged form on my machine.

ccip-sdk/src/cct/solana/token-admin-registry/operations/accept-admin.test.ts
```diff
@@ -0,0 +1,152 @@
...
+          SolanaTokenManager.fromChain(noPendingChain).generateUnsignedAcceptAdmin({
+            tokenAddress: TOKEN,
+            address: ADDRESS,
+            payer: PAYER,
+          }),
+        (err: unknown) =>
+          err instanceof CCTParamsInvalidError &&
+          err.context.param === 'authority' &&
+          typeof err.context.reason === 'string' &&
+          err.context.reason.includes('no pending administrator'),
```
aelmanaa: **[blocker — test is RED]** The stub fix is correct — `noPendingChain` omits `pendingAdministrator`, so the test now reaches the `!tokenConfig.pendingAdministrator` branch (the structural A1 concern is addressed). But the assertion substring doesn't match the message text:

- **Message** (`accept-admin.ts:73`): `"no administrator is pending for this token (current administrator: …) — nothing to accept"`
- **Assertion** (this line): `err.context.reason.includes('no pending administrator')`

`"no administrator is pending"` does not contain `"no pending administrator"` — the word order is swapped. `node --test accept-admin.test.ts` fails 1/5 on this case. `npm run check` stays green because it runs lint+typecheck only, not tests.

```suggestion
          err.context.reason.includes('no administrator is pending'),
```

(Or `'nothing to accept'` — either substring matches the actual message.)

aelmanaa (changes requested):
Re-reviewed at `0429741` (was `0a53a3c`). 6 of 7 prior comments are fixed — thank you. One trivial blocker remains: the "no pending admin" test assertion substring doesn't match the message text, so `accept-admin.test.ts` fails 1/5.

**Fixed at `0429741`:**
- A3 (conflated message) ✅ — two distinct messages at `accept-admin.ts:69-82`, both live-proven on devnet
- Address docstring ✅ — trimmed to "Router or OffRamp"
- Test typing ✅ — `Partial<GenerateAcceptAdminParams>`
- Discriminator + account-metas assert ✅ — `6af010ad89d5a3f6` + full key flags
- Execute-path `reason` assert ✅ — `includes('requires authority to be the executing wallet')`
- Journey docs ✅ — `@remarks`/`@see` to registerAdmin/transferAdmin/setPool on both facade methods

**Remaining (inline):** the A1 stub fix is structurally correct (now reaches the `!pendingAdministrator` branch), but the assertion checks `reason.includes('no pending administrator')` while the message reads `"no administrator is pending for this token…"` — word order is swapped, substring doesn't match, test is RED. One-word fix: align the assertion to `'no administrator is pending'` or `'nothing to accept'`.

ccip-sdk/src/cct/solana/token-admin-registry/operations/accept-admin.test.ts
aelmanaa: @mervin-link  nit

mervin-link: @aelmanaa Fixed.

aelmanaa (approved):

apedrob (approved):

PR #319, opened 2026-07-30
feat(cct-sdk): Add get supported tokens solana op

mervin-link:
## What

- [DAPP-10812](https://smartcontract-it.atlassian.net/browse/DAPP-10812)
- Expose the existing `SolanaChain.getSupportedTokens` implementation through the Solana CCT facade, with CCT input validation and Router resolution.
- Improve Solana token-pool decode diagnostics and missing-state recovery guidance.

## Why

- Allow CCT SDK consumers to list token mints configured in a Solana Router’s TokenAdminRegistry
- Make token-pool configuration failures easier to diagnose

## Notes

- All error changes are follow ups from this [comment](https://github.com/smartcontractkit/ccip-tools-ts/pull/309#issuecomment-5105200935) (2. Follow-up polish batch).

[DAPP-10812]: https://smartcontract-it.atlassian.net/browse/DAPP-10812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/solana/index.ts
```diff
@@ -726,6 +729,22 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
     return this.#getTokenAdminRegistry.query(this.chain, opts)
   }
 
+  /**
+   * Lists SPL token mints configured in a Router's TokenAdminRegistry.
```
aelmanaa: **[Low — cross-family parity, follow-up]** This adds `SolanaTokenManager.getSupportedTokens` but `EVMTokenManager` has no counterpart at the TokenManager layer. It's the same asymmetry the 10787 ⑩ residual noted for `getTokenAdminRegistry`, and it's *lower-stakes here* because the shared read already exists on **both** families at the `Chain` layer (`chain.ts:1994`) — a user can call `chain.getSupportedTokens` cross-family today, and the name+`string[]` shape you chose match it exactly (nice). I'd treat a symmetric `EVMTokenManager` convenience as an optional follow-up, only if/when EVM grows its own address→registry resolution facade — not a blocker.

ccip-sdk/src/cct/solana/token-admin-registry/operations/get-supported-tokens.ts
```diff
@@ -0,0 +1,23 @@
+import type { SolanaChain } from '../../../../solana/index.ts'
+import { SolanaQuery } from '../../query.ts'
+import { validatePublicKey } from '../../validate.ts'
+
+/** Parameters for listing tokens configured in a Solana TokenAdminRegistry. */
+export type GetSupportedTokensParams = {
```
aelmanaa, line 6: **[Low — call-shape drift]** The facade takes `{ address }` (an opts object) while core/EVM are positional `getSupportedTokens(address, opts?: { page? })` (`chain.ts:1994`). That's fine — it matches the CCT convention of one opts object per op — but note two small consequences worth a line of doc: (a) a dev porting a call from `chain.getSupportedTokens(router)` to `cct.getSupportedTokens({ address: router })` rewrites the call shape; (b) there's no `page` here. Solana returns all TARs in one `getProgramAccounts` (655 live) so pagination genuinely doesn't apply — worth a one-liner in the tsdoc saying "returns all configured mints in a single scan (no pagination)" so the omission reads as deliberate.

ccip-sdk/src/cct/solana/token-admin-registry/operations/get-supported-tokens.ts
```diff
@@ -0,0 +1,23 @@
...
+  address: string
+}
+
+/** Lists the SPL token mints configured in a TokenAdminRegistry. */
+export class GetSupportedTokens extends SolanaQuery<GetSupportedTokensParams, string[]> {
+  /** Resolves the Router and lists its configured token mints. */
+  async query(chain: SolanaChain, params: GetSupportedTokensParams): Promise<string[]> {
+    validatePublicKey(this.constructor.name, 'address', params.address)
+
+    const router = await chain.getTokenAdminRegistryFor(params.address)
```
aelmanaa, line 20: **[Low — perf note, mostly inherited]** The delegated core `getSupportedTokens` (`solana/index.ts:1721`) is an **unbounded** `getProgramAccounts` (memcmp on the discriminator only — good, not a full scan — but pulls full account data for every TAR and re-derives a PDA per account: 655 `findProgramAddressSync` calls live). This is **pre-existing in core, not introduced by this PR**, and the op correctly reuses it rather than forking. Flagging only so it's on record: if this ever needs to scale, the fix belongs in core (add a `dataSlice` to fetch just through mint@169, and consider whether the per-account PDA re-derivation can be relaxed given the memcmp already pins the discriminator). No change requested in this PR.

ccip-sdk/src/cct/solana/token-admin-registry/operations/get-supported-tokens.test.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { SolanaChain } from '../../../../solana/index.ts'
+import { CCTParamsInvalidError } from '../../../errors.ts'
+
+const ADDRESS = Keypair.generate().publicKey.toBase58()
+const ROUTER = Keypair.generate().publicKey.toBase58()
+const TOKENS = [Keypair.generate().publicKey.toBase58()]
+
+describe('GetSupportedTokens (cct/solana)', () => {
+  describe('query', () => {
+    it('resolves the Router and lists its configured token mints', async () => {
```
aelmanaa: **[Low — test coverage]** The happy-path test stubs `getTokenAdminRegistryFor` to return `ROUTER` unconditionally, so it proves the delegation wiring (resolved address + router passthrough) but not the **OffRamp-resolution** branch that is the op's main DX value-add over calling core directly (`getTokenAdminRegistryFor` resolves an OffRamp → its router, `solana/index.ts:1299`). Consider one more case asserting an OffRamp address flows through to the resolved router. Minor — the resolution itself is covered in core's tests.

ccip-sdk/src/cct/errors.ts
```diff
@@ -212,10 +212,11 @@ export class CCTOperationUnsupportedError extends CCIPError {
 export class CCTDataDecodeError extends CCIPError {
   override readonly name = 'CCTDataDecodeError'
   /** Creates a CCT data decode error. */
-  constructor(message = 'Unable to decode CCT data', options?: CCIPErrorOptions) {
-    super(CCIPErrorCode.CCT_DATA_DECODE_FAILED, message, {
+  constructor(account: string, options?: CCIPErrorOptions) {
```
aelmanaa, line 215: **[Nit — collateral breaking constructor change, acceptable]** `CCTDataDecodeError` changed from `(message?, options?)` to `(account: string, options?)`. This new shape is strictly better (structured `context.account`, `CCIPError.from` normalizing non-Error causes, both tested). Single caller updated (`programs/token-pool.ts:65`). No action needed — noting for the record since it's outside the getSupportedTokens feature proper.
One of the advantages here is that you can reuse this error and each time directly parse the account address from error.context.account (instead of parsing from a non structured string)

aelmanaa (approved):
I verified it live on devnet with a triple-source proof: the CCT op, core, and a from-scratch `getProgramAccounts` enumeration all return the identical **655-mint** set against the real `ccip-router 1.6.2` router, and I confirmed the enumeration's PDA invariant against a live router-owned v2 TAR. `npm run check -w ccip-sdk` is green and the touched tests (86) pass. My comments are all Low/Nit: the `EVMTokenManager` facade asymmetry (①, a follow-up, and milder than the 10787 analog because the shared read is already on `Chain`), a doc-the-no-pagination note (②), a perf note on the inherited unbounded scan in core (③, not this PR's fix), and an OffRamp-path test (④). None block. **APPROVE (mergeable).**

apedrob (approved):

PR #321, opened 2026-08-03
fix(cct-sdk): Update and refactor solana lifecycle and tests

mervin-link:
## What

- Include CCT sources and builds in the `ccip-sdk` package and remove the dev build
- Add Solana `prepare` lifecycle: `prepare (validate → parse) → build unsigned tx → submit`
- Migrate `acceptAdmin` to parse validated public-key strings into PublicKeys
- Add `validateOptionalPublicKey` and replace truthy optional-key validation checks
- Group Solana operation tests by `generate`, `validation`, and `execute`

## Why
 
- Enable incremental Solana input normalization without changing existing operation APIs or refactoring all operations at once
- Make operation test coverage easier to navigate


ccip-sdk/src/cct/solana/validate.ts
```diff
@@ -49,6 +49,15 @@ export function validatePublicKey(
   parsePublicKey(operation, param, value)
 }
 
+/** Asserts an optional `value` is a valid Solana public key string when provided. */
```
aelmanaa: The one-liner "Asserts an optional `value` is a valid Solana public key string when provided" is ambiguous about exactly the thing this refactor changes — under the old truthiness code `''` and `null` counted as "not provided"; now only `undefined` is absent. It's also the only pubkey validator missing the `@throws` tag its siblings (`validatePublicKey`, `validatePublicKeys`, `validateNonEmptyString`) all carry, so it renders thinner in TypeDoc.

Suggest:

```
/**
 * Asserts `value` is a valid Solana public key string, or is absent.
 * Only `undefined` counts as absent; `null` and `''` are treated as
 * provided and will be rejected.
 * @throws {@link CCTParamsInvalidError} if a non-`undefined` `value` is not a valid public key string.
 */
```

ccip-sdk/src/cct/solana/validate.test.ts
```diff
@@ -26,6 +27,14 @@ describe('Validate (cct/solana)', () => {
     assert.doesNotThrow(() => validatePublicKey('op', 'payer', PublicKey.default.toBase58()))
   })
 
+  it('accepts omitted and rejects invalid optional public keys', () => {
+    assert.doesNotThrow(() => validateOptionalPublicKey('op', 'authority', undefined))
```
aelmanaa, line 31: Coverage is thin for the behavior this refactor guards. The test at line 31-33 covers `undefined` passes and `''` throws, but two cases are missing:

- `null` throws — the value most likely to arrive from JSON/env, and it exercises the `typeof value !== 'string'` branch that `''` doesn't (empty string is a string, it fails inside `new PublicKey('')`).
- a valid key passes — without this, a regression that rejects all values would pass silently.

Ideally also one op-level test asserting `authority: ''` now throws `CCTParamsInvalidError` end-to-end, not just at the validator level.

ccip-sdk/src/cct/solana/operation.ts
```diff
@@ -32,14 +32,28 @@ function withPayer<P extends object>(
...
+  Parsed = SolanaGenerateParams<P>,
 > extends Operation<SolanaChain, SolanaGenerateParams<P>, Tx, TransactionResult> {
-  /** Build instructions after params have been validated. */
-  protected abstract buildUnsigned(chain: SolanaChain, params: SolanaGenerateParams<P>): Promise<Tx>
+  /**
+   * Normalize params without mutating the caller's input.
+   * Override to apply defaults, convert values, or validate individual fields.
+   */
+  protected parse(params: SolanaGenerateParams<P>): Parsed {
+    return params as Parsed
```
aelmanaa, line 51: The default `parse` is concrete with `return params as Parsed`. If a subclass narrows `Parsed` (e.g. `ParsedAcceptAdminParams` with `PublicKey` fields) but forgets to `override parse`, it inherits this identity cast and returns a wrong runtime shape that TypeScript can't catch — `SolanaGenerateParams<P>` is not `ParsedAcceptAdminParams`, but the `as Parsed` silences it.

Worth a doc-comment here: "override `parse` whenever you narrow `Parsed`" — so the footgun is named at the source.

ccip-sdk/src/cct/solana/validate.ts
```diff
@@ -49,6 +49,15 @@ export function validatePublicKey(
...
+export function validateOptionalPublicKey(
```
aelmanaa, line 57: The error message ends with `got ${String(value)}` — for an empty string that produces `got ` with nothing after it, looking like a cut-off log line. This is now a reachable path for the most common "unset" mistake (passing `''` instead of `undefined`).

Worth quoting the value so the message is unambiguous: `got "${String(value)}"` → `got ""`.

ccip-sdk/src/cct/solana/operation.ts
```diff
@@ -32,14 +32,28 @@ function withPayer<P extends object>(
 export abstract class SolanaOperation<
   P extends object,
   Tx extends UnsignedSolanaTx = UnsignedSolanaTx,
+  Parsed = SolanaGenerateParams<P>,
```
aelmanaa, line 42: The class-level TSDoc doesn't mention the new `parse` / `Parsed` machinery — the validate-vs-parse rule (cross-field constraints → `validate`; per-field normalize/default/convert → `parse`) is taught only in an inline comment on `accept-admin`. Now that only `accept-admin` adopts `parse` while 7 ops still do everything in `validate`, a dev reading the base class has no canonical pointer for which method to use.

Worth adding a short block on the class doc explaining: `validate` for cross-field constraints, `parse` for per-field normalization/defaults/conversion, and that `parse` should be overridden whenever `Parsed` is narrowed.

apedrob (approved):

ccip-sdk/src/cct/solana/operation.ts
```diff
@@ -28,18 +28,40 @@ function withPayer<P extends object>(
   return { ...rest, payer } as SolanaGenerateParams<P>
 }
 
-/** Solana CCT write base. Subclasses supply {@link validate} and {@link buildUnsigned}. */
+/**
```
aelmanaa, line 32: **[§B nit — partial-migration tracking]** The `parse`/`Parsed` lifecycle is now adopted by `accept-admin`, `configure-allowlist`, and `remove-from-allowlist`, while the remaining parse-only ops still let `validate()` do everything (or run empty). The class doc now teaches the rule well (good — resolves the earlier doc nit), but there's still no in-code tracking of the in-progress migration. Add a short `TODO`/issue link so the mixed pattern is a known, intentional state rather than drift.

ccip-sdk/src/cct/solana/operation.test.ts
```diff
@@ -27,9 +27,51 @@ class TestOperation extends SolanaOperation<{ value: string }> {
...
+    this.lifecycle.push(`build:${params.value}`)
+    this.captured = params
+    return Promise.resolve({ family: ChainFamily.Solana, instructions: [] })
+  }
+}
+
 const chain = { logger: console, connection: {} } as unknown as SolanaChain
 
 describe('SolanaOperation', () => {
+  it('validates, parses, then builds without mutating input', async () => {
```
aelmanaa, line 64: **[§B nit — short-circuit coverage]** Lifecycle order is asserted on the success path (`['validate','parse','build']`), but there's no test that a **throwing `validate` short-circuits** `parse`/`buildUnsigned`. Add one `assert.rejects` where `validate` throws and assert `parse`/`build` never ran — it guards the ordering guarantee the base now documents.

ccip-sdk/src/cct/solana/operation.ts
```diff
@@ -28,18 +28,40 @@ function withPayer<P extends object>(
...
+ * Solana CCT write base. Subclasses supply {@link validate} and {@link buildUnsigned}.
+ *
+ * Use {@link validate} for cross-field constraints. Override {@link parse} for per-field
+ * validation, defaults, or conversion; it must be overridden whenever `Parsed` differs from
+ * `SolanaGenerateParams<P>`.
+ */
 export abstract class SolanaOperation<
```
aelmanaa, line 39: **[§B nit — base no-op `validate`]** Parse-only ops must still write an empty `validate() {}` stub. Consider a default no-op `validate` on the base so those ops can omit the boilerplate (removes the empty stub in `configure`/`remove`/`accept`-style ops).

aelmanaa (approved):
**§A — `validateOptionalPublicKey` + test refactor.** Centralizes 8 duplicated `if (params.x) validatePublicKey(...)` guards into one `validateOptionalPublicKey`, swapping truthiness (`if (x)`) for `!== undefined`. The only new runtime divergence is the empty string: old code silently rewrote a caller's `''`/`null` authority into "use payer" (masking a bug); new code rejects it. `''`/`null` were never valid base58 pubkeys → **pure hardening, no program-accepted input newly refused**. Aligns to an existing in-file standard (`resolvePoolProgram`, `validateInteger` already use `!== undefined`); `validate.ts` isn't re-exported, so no CLI↔SDK boundary moves. The `deploy-token.ts` `freezeAuthority?: string | null` keeping its explicit `!== null` guard is **correct** — `null` is its documented "disable freezing" sentinel.

**§B — `parse`/`prepare` lifecycle.** `SolanaOperation<P,Tx,Parsed=SolanaGenerateParams<P>>` gains `parse(params): Parsed` (no-mutation normalize/default/convert), `prepare()` = `validate()`+`parse()`, and `buildUnsigned` consumes typed `Parsed`. The `Parsed` default makes the 9 non-adopters textually identical to before; adopters (accept-admin, and now configure/remove-allowlist) get typed `PublicKey`s with one fewer re-parse. Contained (base not re-exported), cross-family `Operation<>` untouched.

**Live devnet proofs (carry forward — the `add68cb` delta is validation/doc/test only, no instruction/account change):**

- §A `createLookupTable` on devnet, wallet [`2SGSoyjD…C3V3`](https://explorer.solana.com/address/2SGSoyjD1QLEpL8SsA2Zhc51jRCQPiZ6GjJWZGkvC3V3?cluster=devnet): (a) `authority` explicit → [tx `5Xcj2tj…Hwjvp`](https://explorer.solana.com/tx/5Xcj2tjJGqRLr2A1wyL3TFFPgeZ9fS2FKCcWjASJcrgV1Lxo8vvEEFBFJnWfCTDtGgynZjRGAWYN1NgWMRdHwjvp?cluster=devnet); (b) `authority` omitted (→payer) → [tx `8NCoUz8…utRt9`](https://explorer.solana.com/tx/8NCoUz8YoApvcNsYLVWsWnnCiZwZRioPRNh7FEBnJzxJ8zydx5n7FytGe3f4VcztigDV34DYZSHydW5HPxutRt9?cluster=devnet); both ALTs read back `authority = 2SGSoyjD…C3V3`. (c) negative `authority: ''` → rejected pre-flight `CCTParamsInvalidError`.
- §B real `acceptAdmin` LANDED through the new `prepare` path — [tx `5gLXmjt…5ZUzJu`](https://explorer.solana.com/tx/5gLXmjtaxHkbrQjaCx1PqTJaLnXWkubZxYghXvfKmPFajMeHFrNCqTzg76gUuVMeN87q7C79iL54Ut98u25ZUzJu?cluster=devnet), on-chain ix data `6af010ad89d5a3f6`, accounts `{config, token_admin_registry, mint, authority}` **byte-identical** to the generated instruction; TAR read-back `administrator = 2SGSoyjD…C3V3`, `pendingAdministrator = 1111…1111`. Generate-path `authority` explicit vs omitted → identical instruction (no-mutation confirmed). Grounded to `ccip-router/src/token_context.rs:179-196` (`authority` = `Signer` with `address = pending_administrator @ Unauthorized`).

**Verified at `add68cb`:** `npm run check` exit 0 (lint+typecheck+generate), **27/27** targeted tests pass (`validate` + `deploy-token-pool` + `operation`), including the new `null`/`''`-rejected and op-level `authority:''`-throws cases.

**Resolved since the last pass (`e27c959 → add68cb`):** §A TSDoc (validateOptionalPublicKey now documents "only `undefined` is absent; `null`/`''` rejected" + `@throws`), §A coverage (`null`-throws + valid-passes + op-level `authority:''`), §A quoted error message, §B class/`parse` TSDoc (validate=cross-field / override `parse` when `Parsed` differs), and the **B4** `append-to-lookup-table` regression fix. Thanks for these.

**Not blockers / correct-by-design:** the truthiness→`!== undefined` shift is pure hardening (`''`/`null` never valid pubkeys); `freezeAuthority`'s `null` sentinel correctly keeps its explicit guard; `parse()` no-mutation is on-chain-proven byte-identical; the EVM side having no optional-address validator is a **pre-existing** gap, not this branch's responsibility. The B4 `append-to-lookup-table.ts:94` fix (`?.` restored) is correct.

ccip-sdk/src/cct/solana/operation.ts
mervin-link: I considered this, but it would require `override` across every existing validator under `noImplicitOverride`. I’ll add a default no-op validation on the base once every op uses prepare.

PR #322, opened 2026-08-03
feat(cct-sdk): Add EVM getTokenPoolState op + query read class

apedrob:
## What

- Adds `getTokenPoolState`: EVM read op returning a token pool's admin state (supports versions 1.5.0-2.0.0).
- Adds a cross-family `Query` base (`cct/query.ts`) wiring `validate → read`, with `EVMQuery` and
  `SolanaQuery`  (read counterpart of `cct/operation.ts`)

## Why

CCT SDK needs to surface the admin/config of deployed token pools

## Testing

- CI / Manual testing


ccip-sdk/src/cct/solana/index.ts
```diff
@@ -722,14 +727,23 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
...
+  ): Promise<LockReleaseGetTokenPoolStateResult>
+  /**
+   * Reads a Burn/Mint or custom token pool's state account; its config carries no liquidity
+   * fields. Pass `poolProgramAddress` instead of `poolType` for a custom pool program.
+   */
+  getTokenPoolState(
+    opts: (BurnMintPoolProgramRef | CustomPoolProgramRef) & { tokenAddress: string },
+  ): Promise<BaseGetTokenPoolStateResult>
+  /** Reads a pool state account whose program type is not known statically. */
+  getTokenPoolState(opts: GetTokenPoolStateParams): Promise<GetTokenPoolStateResult> {
```
mervin-link, line 845:
```suggestion
  getTokenPoolState(opts: GetTokenPoolStateParams): Promise<GetTokenPoolStateResult>
  getTokenPoolState(opts: GetTokenPoolStateParams): Promise<GetTokenPoolStateResult> {
```

Should we add a public overload for `GetTokenPoolStateParams` as well? TypeScript does not expose the implementation signature to callers.

ccip-sdk/src/cct/query.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Abstract CCT read operation. Subclasses supply {@link Query.validate} and {@link Query.read};
+ * the base wires {@link Query.query}, mirroring how `Operation.generate` gates `buildUnsigned`.
+ * No wallet, no calldata, no submit.
+ */
+export abstract class Query<Chain, Params extends object, Result> {
+  /** camelCase id; matches the token-manager facade method and error context. */
+  abstract readonly name: string
+
+  /** Reject invalid params before any chain RPC. */
+  protected abstract validate(params: Params): void
```
mervin-link: What do you think of introducing `prepare` here? It will call both `validate` and `parse`.

ccip-sdk/src/cct/solana/index.ts
apedrob: Uh, good catch, added!

ccip-sdk/src/cct/query.ts
apedrob: Adopted, with one difference: reads get a single prepare rather than prepare calling validate + parse

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,270 @@
...
+ * liquidity concerns, not admin ones; read those via `cct.chain.getTokenPoolConfig()`.
+ */
+export type LegacyTokenPoolState = TokenPoolStateCore & {
+  version: Exclude<TokenPoolVersion, typeof TokenPoolVersion.V2_0_0>
+  type: TokenPoolType
+}
+
+/** Admin state v2.0.0 adds to {@link TokenPoolStateCore}: the fee role and the finality window. */
+type TokenPoolStateCoreV2_0_0 = TokenPoolStateCore & {
+  version: typeof TokenPoolVersion.V2_0_0
```
aelmanaa, line 69: `version` is the same field name with a different type across families: EVM is a semver string (`'2.0.0'`, the result discriminant at `:69`/`:80`/`:85`); Solana is a number (account-layout version, `get-token-pool-state.ts:46`). A dev who writes `state.version === '2.0.0'` on EVM and reuses the check on Solana gets a silent always-false — no type error (the field exists on both), no runtime error, just wrong behavior.

Worth renaming the Solana one (e.g. `accountVersion`) or, at minimum, documenting both loudly. Non-breaking on EVM (new/unshipped); on Solana keep it additive — retyping/renaming `version` is breaking on the already-merged read.

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,270 @@
...
+ * @throws {@link CCTContractTypeInvalidError} for a siloed pool — it escrows per remote chain
+ * (`getLockBox(uint64)`), so no single `lockBox` describes it
+ */
+async function readLockReleaseTokenPoolV2_0_0(
+  chain: EVMChain,
+  poolAddress: string,
+  type: LockReleaseTokenPoolType,
+): Promise<LockReleaseTokenPoolStateV2_0_0> {
+  if (type !== 'LockReleaseTokenPool')
+    throw new CCTContractTypeInvalidError(poolAddress, 'LockReleaseTokenPool', type)
```
aelmanaa: A `SiloedLockReleaseTokenPool` is a recognized, supported type, but it's rejected with `CCTContractTypeInvalidError(poolAddress, 'LockReleaseTokenPool', type)` → renders as "expected a LockReleaseTokenPool, got SiloedLockReleaseTokenPool", telling the dev their address/type is wrong. The real reason — siloed pools escrow per-remote-chain via `getLockBox(uint64)`, so the no-arg getter this op uses doesn't exist — is in the `@throws` (`:213`) but a caught error doesn't carry it.

Worth throwing an error whose message states the actual reason + next step (e.g. "siloed pools escrow per remote chain — read per-lane lockboxes via getLockBox(remoteChainSelector)").

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,270 @@
...
+  /** Validates the pool address; nothing to convert for {@link read}. */
+  protected prepare(params: GetTokenPoolStateParams): GetTokenPoolStateParams {
+    validateAddress(this.name, 'poolAddress', params.poolAddress)
+    return params
+  }
+
+  /**
+   * Resolves the pool's type + version, then reads it through the getters that version has.
+   * @remarks Dispatch is explicit, not floor-matched like the write ops' encoders: a read's shape
+   * is bound to the ABI it decodes through, so adding a pool version has to fail to compile here
```
aelmanaa: "adding a pool version has to fail to compile here" is overstated — a new non-2.0.0 version is absorbed by `LegacyTokenPoolState['version'] = Exclude<TokenPoolVersion, …>` and flows silently through `readLegacyTokenPool` (`:264`) with no compile error. Either soften the wording, or add a `never`-exhaustive switch on `version` to make it actually true.

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,270 @@
+/**
+ * getTokenPoolState — reads a token pool's admin state (v1.5.0–v2.0.0): the owner and admin roles
```
aelmanaa, line 2: Class doc says "v1.5.0–v2.0.0", but real 1.5.0 pools report `*AndProxy` type names not in `TOKEN_POOL_TYPES` and are rejected — so 1.5.0 support is effectively non-proxy only. Worth qualifying the range (e.g. "v1.5.0 non-proxy – v2.0.0") so it doesn't contradict the AndProxy rejection.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -731,14 +736,30 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
    * ```ts
```
aelmanaa, line 815: The `getTokenPoolState` facade `@see` point at `deployTokenPool`/`setPool` (copy-paste from sibling ops) — point them at the EVM `getTokenPoolState` (cross-family) and at each other (generate↔execute). The Solana `@example` is also thinner than EVM's — worth bringing to parity (show a `config` field / the LR-only fields).

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.test.ts
```diff
@@ -0,0 +1,308 @@
+import assert from 'node:assert/strict'
```
aelmanaa, line 1: Missing two test cases:
- a `CCTContractVersionUnsupportedError` case — a supported type reporting an unknown version. Documented as `@throws` but untested.
- a non-pool-address case — a valid-but-non-pool address currently surfaces a raw ethers `CALL_EXCEPTION`, not a typed `CCT*` error. Worth wrapping the `typeAndVersion` failure into a typed error.

ccip-sdk/src/cct/solana/token-pool/operations/get-token-pool-state.ts
```diff
@@ -57,24 +59,16 @@ export type LockReleaseGetTokenPoolStateResult = GetTokenPoolStateResultBase & {
   }
```
aelmanaa, line 59: Result shape differs by family and it's undocumented: EVM is flat (`state.owner`), Solana nests under `config` (`state.config.owner`), and three shared concepts are renamed (`token`/`tokenDecimals`/`rmnProxy` vs `config.mint`/`config.decimals`/`config.rmnRemote`). The renames track genuine chain terminology so they're defensible — but nothing warns a cross-family dev.

Worth a `@see` between the two facade methods and a one-line "result shape differs by family (EVM flat; Solana nests under `config`; `version` is a string here / number there)". Good: the shared subset — `owner`/`rateLimitAdmin`/`router` — is identically named, no `owner`-vs-`admin` trap.

aelmanaa (approved):
Reviewed `33e48dd` across type-safety, architecture, and DX — grounded against the chainlink-ccip Solidity pool contracts and proven live on both native chains. The shared `Query` read base and the EVM `getTokenPoolState` op are well-built; approving with nits. Also re-reviewed the `Address PR comments` delta (the `Query` base `validate→read` → `prepare→read` refactor): clean, non-regressive, DRY win (removes the Solana double-parse); all 4 read subclasses migrated.

**Strengths**

- **DRY** — `cct/query.ts` (`prepare → read`) is a clean cross-family read base both families bind to; `getTypedContract` centralizes the single ethers→ethers-abitype cast without hiding call-site types. Good template for future reads.

- **Solidity-exact** — every returned field maps to a real getter, verified line-by-line: `getToken()`/`owner()`/`getRmnProxy()`/`getTokenDecimals()`/`getSupportedChains()`, `getDynamicConfig()`→`[router, rateLimitAdmin, feeAdmin]`, `getAllowedFinalityConfig()` (decode matches `FinalityCodec.sol`), `getLockBox()` (LR), `typeAndVersion()`. No field misdecoded/mistyped/misnamed. Siloed rejection is legit — `SiloedLockReleaseTokenPool.getLockBox(uint64)` takes an arg, so the no-arg getter genuinely doesn't exist.

**Live-proven on both native chains (EVM→Sepolia, Solana→devnet)**

- **EVM (Sepolia), SDK-decode vs raw `eth_call`, all fields match:** burn-mint pool [`0x4CcbDd6CF18800360161E4D2A519A2047176bDF0`](https://sepolia.etherscan.io/address/0x4CcbDd6CF18800360161E4D2A519A2047176bDF0) (CCIP-BnM) and lock-release [`0x4f3DF8bF87EfC7ef2076571e0d2e4F762F1d06c0`](https://sepolia.etherscan.io/address/0x4f3DF8bF87EfC7ef2076571e0d2e4F762F1d06c0) (LINK). No standalone v2.0.0 pool is registered on Sepolia/Base Sepolia (only proxies, correctly rejected) → v2.0.0 reader covered by the 47-test suite.

- **Solana (devnet), refactored `SolanaQuery.read()` path vs raw `getAccountInfo` decode, all fields match:** burn-mint pool state [`5oqErPLJqkRuE573WopwRpnJszwBjC6Xo5mBx5Y98NVH`](https://explorer.solana.com/address/5oqErPLJqkRuE573WopwRpnJszwBjC6Xo5mBx5Y98NVH?cluster=devnet) and lock-release [`7Xxhu5RxWxb8FhnS5y1yXRgM3fr7zge1GU9FeLTsPBNB`](https://explorer.solana.com/address/7Xxhu5RxWxb8FhnS5y1yXRgM3fr7zge1GU9FeLTsPBNB?cluster=devnet); the new `validate→read` gating rejects a bad address pre-RPC.

**Per-protocol-version validity vs chainlink-ccip EVM (verified per version)** — the op is valid across every supported version; no getter is assumed on a version that lacks it:

- **v1.5.0 / v1.5.1 / v1.6.1** (single shared `readLegacyTokenPool`): all six getters it calls are standalone functions in each version's vendored ABI. The v1.5.0 decimals-from-token fallback is correct (no `getTokenDecimals` pre-2.0). Live-confirmed the fold-in happened at 2.0: `getDynamicConfig()` reverts on a real 1.6.1 pool.

- **v2.0.0**: `getDynamicConfig()→[router,rateLimitAdmin,feeAdmin]` (TokenPool.sol:213), `getAllowedFinalityConfig()` (bytes4, decode matches `FinalityCodec.sol` bit layout), `getLockBox()` (LR, LockReleaseTokenPool.sol:43) — grounded to source. These v2-only getters are never called on legacy.

- **Rejections correct:** `*AndProxy` (not in `TOKEN_POOL_TYPES`) and Siloed v2.0 (`getLockBox(uint64)` takes an arg → the bare getter genuinely doesn't exist) — both live-confirmed to throw `CCTContractTypeInvalidError`.

**Cross-family polymorphism (verified):** the operation interface is polymorphic — `Query<Chain,Params,Result,Parsed>` is chain-generic, `EVMQuery`/`SolanaQuery` just bind `Chain` and forward (no override), the `prepare→read` lifecycle is identical, and `getTokenPoolState` is the same facade method name on both TokenManagers. It breaks only at the data shape (M1 `version` string-vs-number, M2 flat-vs-`config`) — by design. One caveat: the base `TokenManager` does not declare `getTokenPoolState` abstract, so cross-family facade parity is convention, not compiler-enforced.

The 3 majors + nits in the inline comments are all DX/docs/test — cheap to fix while CCT is still unpublished. No blockers.

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
apedrob: It won't compile with a string on Solana types. Collision is real tho, I'll doc both instead of renaming; covering it in the cross-family note from your comment

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
apedrob: Agreed, reads like a wrong address. Adding an optional reason to `CCTContractTypeInvalidError` so it says siloed pools escrow per remote chain, read per-lane via `getLockBox(remoteChainSelector)`. On the error so the other type gates reuse it

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
apedrob: True, `Exclude<TokenPoolVersion, '2.0.0'>` swallows a new version. Making it true instead of softening: `switch (version)` + `const _exhaustive: never = version`

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.test.ts
apedrob: Adding the op-level `CCTContractVersionUnsupportedError` case

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
apedrob: Non-proxy exists at 1.5.0 and is accepted; `BurnMintTokenPool 1.5.0` / `LockReleaseTokenPool 1.5.0` in `contracts-ccip/v1.5.0`. `*AndProxy` is excluded by type, not by version. Keeping the range; it's already called out on the facade (`evm/index.ts:349`) and tested

ccip-sdk/src/cct/solana/index.ts
apedrob: No `@see` in that block (`solana/index.ts:727-764`), maybe from a sibling op. Example parity holds tho, bringing the Solana one up to EVM's with the cross-family note

ccip-sdk/src/cct/solana/token-pool/operations/get-token-pool-state.ts
apedrob: Agreed. `@remarks` on both facades: EVM flat vs Solana nested under `config`, `token`/`tokenDecimals`/`rmnProxy` vs `config.mint`/`config.decimals`/`config.rmnRemote`, `version` semver here vs account-layout number there

mervin-link (approved):

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,286 @@
...
```
aelmanaa, line 2: nit — version range: Class doc says "v1.5.0–v2.0.0", but `*AndProxy` types are rejected. Suggest qualifying: "v1.5.0 non-proxy – v2.0.0" so it doesn't contradict the AndProxy rejection.

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,286 @@
...
+ * @throws {@link CCTContractTypeInvalidError} for a siloed pool — it escrows per remote chain
+ * (`getLockBox(uint64)`), so no single `lockBox` describes it
+ */
+async function readLockReleaseTokenPoolV2_0_0(
+  chain: EVMChain,
+  poolAddress: string,
+  type: LockReleaseTokenPoolType,
+): Promise<LockReleaseTokenPoolStateV2_0_0> {
+  if (type !== 'LockReleaseTokenPool')
+    throw new CCTContractTypeInvalidError(
```
aelmanaa, line 222: [M3 — siloed-pool error message (RESOLVED)] The `4599c45` commit added an optional `reason` param to `CCTContractTypeInvalidError`; siloed rejection now says "siloed pools escrow per remote chain; read per-lane lockboxes via `getLockBox(remoteChainSelector)`". New test case asserts the reason is present.

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
```diff
@@ -0,0 +1,286 @@
...
+
+  /** Validates the pool address; nothing to convert for {@link read}. */
+  protected prepare(params: GetTokenPoolStateParams): GetTokenPoolStateParams {
+    validateAddress(this.name, 'poolAddress', params.poolAddress)
+    return params
+  }
+
+  /**
+   * Resolves the pool's type + version, then reads it through the getters that version has.
+   * @remarks Dispatch is an exhaustive `switch`, not floor-matched like the write ops' encoders: a
```
aelmanaa, line 253: nit — exhaustive switch: The `@remarks` wording now correctly states "a new `TokenPoolVersion` fails to compile here until it is pointed at a reader" — this is accurate with the new `switch` + `never` guard. Good fix from the original overstated claim.

ccip-sdk/src/cct/errors.ts
```diff
@@ -120,15 +120,26 @@ export class CCTTxNotConfirmedError extends CCIPError {
...
+  /**
+   * Creates a contract-type-invalid error. `reason` is appended to the message and kept in
+   * `context`; pass it when `actual` is a recognized type rejected on its own grounds, so the
+   * message does not read as "wrong address".
+   */
+  constructor(
+    address: string,
+    expected: string,
+    actual: string,
+    reason?: string,
```
aelmanaa, line 132: [N1 — `reason` parameter position] The new optional `reason` is the 4th positional param, before `options?` (SDK-public). A TS caller passing `options` positionally now gets a compile error, but an untyped/JS caller could misbind. Consider folding `reason` into `options` or documenting the signature change.

ccip-sdk/src/cct/solana/index.ts
```diff
@@ -793,8 +798,14 @@ export class SolanaTokenManager extends TokenManager<typeof ChainFamily.Solana>
   }
 
   /**
-   * Reads a Burn/Mint, Lock/Release, or custom token pool's state account.
-   * Pass `poolProgramAddress` instead of `poolType` for a custom pool program.
+   * Reads a Lock/Release token pool's state account, whose config also reports its liquidity
+   * fields (`rebalancer`, `canAcceptLiquidity`).
+   *
+   * @remarks The EVM counterpart, `EVMTokenManager.getTokenPoolState`, returns a different shape:
```
aelmanaa, line 804: [M2 — result shape divergence (RESOLVED)] The `@remarks` added here document the cross-family shape difference (EVM flat `state.owner` vs Solana `state.config.owner`). Good catch that the shared subset — `owner`/`rateLimitAdmin`/`router` — is identically named.

aelmanaa (approved):
✅ **UPDATED APPROVAL** at head `8f66a66` — the `4599c45` "Address PR comments" commit **resolves all 3 major points (M1, M2, M3)** + adds 2 new test cases. **Unanimous APPROVE, 0 blockers.** Residual notes are polish only.

---

## ✅ Resolutions

**M1 (version string/number collision)** → `4599c45` adds cross-family `@remarks` documenting the intentional divergence: EVM flat `{version=semver}` ⇄ Solana `version:number`. *Documented, not unified* — the right call (normalizing would be lossy; the `TokenManager<F>` base enforces no shape parity).

**M2 (flat vs `config` nesting)** → `4599c45` adds mutual `@remarks` on both facades documenting shape divergence (EVM flat `owner`, Solana `config.owner`; field renames `token`↔`mint`, `tokenDecimals`↔`decimals`, `rmnProxy`↔`rmnRemote`). The shared subset (`owner`/`rateLimitAdmin`/`router`) is identically named, no trap.

**M3 (siloed-pool error clarity)** → `4599c45` adds optional `reason` param to `CCTContractTypeInvalidError`; siloed rejection now explains why ("siloed pools escrow per remote chain; read per-lane lockboxes via `getLockBox(remoteChainSelector)`"). New test asserts reason presence.

**Bonus:** EVM version dispatch → **exhaustive `switch` + `never` guard** (a 5th `TokenPoolVersion` now fails to compile until given a reader; was silently routed to legacy reader before).

**Verified:** `npm run check` exit 0; **20/20** get-token-pool-state tests both families (incl. 2 new EVM tests).

**Live-proven side-by-side:**
- EVM Sepolia `BurnMintTokenPool 1.5.1` [`0x4CcbDd6CF18800360161E4D2A519A2047176bDF0`](https://sepolia.etherscan.io/address/0x4CcbDd6CF18800360161E4D2A519A2047176bDF0) → flat, `version="1.5.1"`, no `proposedOwner`
- Solana devnet burn-mint [`J9jfqxhsGM2U7V4aQ6Y2R37hDhVMMVCLYEoQMBzeJCbF`](https://explorer.solana.com/address/J9jfqxhsGM2U7V4aQ6Y2R37hDhVMMVCLYEoQMBzeJCbF?cluster=devnet) → nested `config`, `version=1`, **has `config.proposedOwner`**

---

## From Original Review (`33e48dd`) — All Current**

**Strengths:**
- **DRY** — `cct/query.ts` (`prepare → read`) is clean cross-family read base; `getTypedContract` centralizes the single ethers→abitype cast without hiding call-site types
- **Solidity-exact** — every returned field maps to a real getter (verified line-by-line). Siloed rejection is legit — `SiloedLockReleaseTokenPool.getLockBox(uint64)` takes an arg, so bare getter doesn't exist
- **Live-proven:** EVM Sepolia burn-mint/lock-release match raw `eth_call`; Solana devnet burn-mint/lock-release match raw `getAccountInfo` decode. New `validate→read` gating rejects bad addresses pre-RPC
- **Per-version validity:** v1.5.0/1.5.1/1.6.1 (shared `readLegacyTokenPool`, six getters are standalone functions in each ABI); v2.0.0 (v2-only getters never called on legacy). Rejections correct both live-confirmed

**Cross-family polymorphism:** `Query<Chain,Params,Result,Parsed>` is chain-generic; `EVMQuery`/`SolanaQuery` bind `Chain` and forward; `getTokenPoolState` is same facade name on both TokenManagers. Breaks only at data shape (M1/M2) — by design.

---

## Residual Notes (N1–N4, non-blocking polish)

**N2** — cross-family `@remarks` cover shared/renamed fields but not the **one-sided** ones. Solana returns `config.proposedOwner` (proven live) while EVM's doc says "not readable here" — contradiction. Suggest: *"One-sided fields (no counterpart): EVM-only `feeAdmin`/`finalityDepth`/`finalitySafe`/`lockBox`/`supportedChains`; Solana-only `rebalancer`/`canAcceptLiquidity`/`listEnabled`/`allowList`/`proposedOwner`."*

**N3** — teaching `reason` points at `getLockBox(remoteChainSelector)` (raw ABI getter; SDK does **not** wrap it). Say so: *"…via the pool's raw `getLockBox(uint64)` getter (this SDK does not wrap it)."* so agents don't hunt for nonexistent SDK method.

**N4** — new test asserts `err.context.reason === err.message.split(' — ')[1]`; em-dash in a future `reason` breaks it. Assert `context.reason` directly; document it as the parse target (not the split).

**Optional hardening** — cross-family parity (identical method names, facade behavior) is convention, not compiler-enforced at base. Future drift could go unnoticed. Tying into the deferred CI dep-graph guard: a shared abstract signature or type-level parity test would catch cross-family drift at compile/CI.

ccip-sdk/src/cct/evm/token-pool/operations/get-token-pool-state.ts
aelmanaa: Verified — this is resolved

PR #333, opened 2026-08-06
feat(cct-sdk): Add register token evm op

apedrob:
## What

- Add `registerAdmin` to EVMTokenManager + generateUnsignedRegisterAdmin for offline signing. It proposes a token's administrator through a `RegistryModuleOwnerCustom`, supporting owner, ccip-admin and access-control-default-admin (with versioned discriminated types)
- Add shared `token-admin-registry/contracts.ts` layer with similar approach as other contracts: cached interfaces, resolvers and readers. With this, all token admin ops will share same interface

## Why

- First step for M2 registry flow (registerAdmin, transferAdmin, acceptAdmin, setPool)
- Also we need pre-tx checks turn reverts into clear errors: the module is registered, its version is known, sender holds the authority the chosen method reads, and the token is not already registered

## Testing

- Unit tests cover the three registration paths

## Notes
- `access-control-default-admin` registers msg.sender, so on the offline path whoever signs becomes the administrator, not whoever was named as sender. This is called out on the builder


ccip-sdk/src/cct/evm/index.test.ts
mervin-link, line 1: This index test will grow significantly if we add multiple cases per operation here. Since the operation tests already cover validation and error paths independently, could we limit this file to verifying the methods are exposed on `EVMTokenManager` plus one happy-path delegation test per operation? I'll also update Solana ops test files to sync with this.

ccip-sdk/src/cct/evm/token-admin-registry/operations/register-admin.ts
```diff
@@ -0,0 +1,259 @@
...
+        const tokenGetter = REGISTRATION[method].tokenGetter!
+        const tokenGetterInterface = new Interface([
+          `function ${tokenGetter}() view returns (address)`,
+        ])
+        const admin = (await new Contract(
+          p.tokenAddress,
+          tokenGetterInterface,
+          chain.provider,
+        ).getFunction(tokenGetter)()) as string
+        if (getAddress(admin) !== getAddress(p.sender)) {
```
mervin-link, line 206: We can have `isSameAddress` validator for this.

ccip-sdk/src/cct/evm/token-admin-registry/operations/register-admin.ts
```diff
@@ -0,0 +1,259 @@
...
+      p.tokenAddress,
+    )
+    if (administrator !== ZeroAddress) {
+      throw new CCTParamsInvalidError(
+        this.name,
+        'tokenAddress',
+        `token already has registry administrator ${administrator} — use transferAdmin to hand the role over, or setPool if you are already the admin`,
+      )
+    }
+    if (pendingAdministrator !== ZeroAddress) {
```
mervin-link, line 230: This means that registering again using a different method is not possible.

For example:
- `ccipAdmin` is different from `owner`
- `owner` registers first without accepting
- `ccipAdmin` wants to register itself instead of the owner.

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -103,6 +104,34 @@ export abstract class EVMOperation<P extends { sender?: string }> extends Operat
...
+   * The chain gates on `msg.sender`, and {@link submit} clears any builder-set `tx.from` before
+   * populating the tx (so ethers' own from/signer guard never fires) — an explicit `sender` that
+   * differs from the wallet would therefore let an op's pre-tx checks authorize one address while
+   * a different one actually signs, passing every local guard and reverting on-chain. Ops that
+   * gate on an on-chain role call this from `execute`; build with `generateUnsigned*` instead
+   * when the eventual signer isn't known yet, where `sender` is trusted as given.
+   * @throws {@link CCIPWalletInvalidError} if `wallet` is not a valid signer
+   * @throws {@link CCTParamsInvalidError} if `sender` is given and is not the wallet's address
+   */
+  protected async senderBoundToWallet(wallet: unknown, sender?: string): Promise<string> {
```
mervin-link, line 118: The name sounds like it returns a `boolean`. How about `resolveWalletSender`?

apedrob: Sounds good

ccip-sdk/src/cct/evm/token-admin-registry/operations/register-admin.ts
apedrob: If the contract allows it, make sense to remove the guard here

ccip-sdk/src/cct/evm/operation.ts
```diff
@@ -103,6 +104,34 @@ export abstract class EVMOperation<P extends { sender?: string }> extends Operat
...
+   * Resolves the address a signed submission is authorized against: the signing wallet's own.
+   * The chain gates on `msg.sender`, and {@link submit} clears any builder-set `tx.from` before
+   * populating the tx (so ethers' own from/signer guard never fires) — an explicit `sender` that
+   * differs from the wallet would therefore let an op's pre-tx checks authorize one address while
+   * a different one actually signs, passing every local guard and reverting on-chain. Ops that
+   * gate on an on-chain role call this from `execute`; build with `generateUnsigned*` instead
+   * when the eventual signer isn't known yet, where `sender` is trusted as given.
+   * @throws {@link CCIPWalletInvalidError} if `wallet` is not a valid signer
+   * @throws {@link CCTParamsInvalidError} if `sender` is given and is not the wallet's address
+   */
```
aelmanaa, line 117: **senderBoundToWallet security fix** — excellent addition to the base class. This guards against the "authorization leak": pre-tx checks pre-flighting one address (A), wallet signing as different address (B), all local guards passing, on-chain revert. Now: if pre-flighted sender ≠ wallet address, execute throws locally before signing.

ccip-sdk/src/cct/evm/token-admin-registry/contracts.ts
```diff
@@ -0,0 +1,185 @@
...
+ * @packageDocumentation
+ */
+
+import { Interface, getAddress } from 'ethers'
+import type { TypedContract } from 'ethers-abitype'
+
+import type { EVMChain } from '../../../evm/index.ts'
+import { resultToObject } from '../../../evm/types.ts'
+import { CCTContractTypeInvalidError, CCTContractVersionUnsupportedError } from '../../errors.ts'
+import REGISTRY_MODULE_OWNER_CUSTOM_V1_5_0_ABI from '../artifacts/abi/V1_5_0/registry-module-owner-custom.ts'
```
aelmanaa, line 25: **Shared contracts.ts layer** — follows the pattern from token-pool/contracts.ts. Cached interfaces per version, resolvers throwing typed errors, single `getTypedContract` cast. Establishes canonical structure for all token-admin ops.

ccip-sdk/src/cct/evm/token-admin-registry/operations/register-admin.ts
```diff
@@ -0,0 +1,259 @@
...
+
+/**
+ * Per-method wiring: the RegistryModuleOwnerCustom function this op calls. `owner`/`ccip-admin`
+ * also carry the token getter whose return value the module registers as administrator and
+ * checks against the caller (`_registerAdmin`'s `admin != msg.sender` revert) — used here to
+ * pre-flight that same equality. `access-control-default-admin` has no such getter: unlike the
+ * other two, `registerAccessControlDefaultAdmin` never derives an address from the token at all —
+ * it checks `AccessControl(token).hasRole(DEFAULT_ADMIN_ROLE(), msg.sender)` and then registers
+ * `msg.sender` itself, so it's pre-flighted as a role check in {@link RegisterAdmin.buildUnsigned}
+ * rather than through a `tokenGetter` here.
```
aelmanaa, line 48: **Per-method pre-flight checks** — authorization is method-scoped: owner getter, ccip-admin getter, access-control role. Each independently validated before encoding, so authorization mismatches fail here with clear errors, not as opaque on-chain reverts.

ccip-sdk/src/cct/evm/token-admin-registry/operations/register-admin.test.ts
```diff
@@ -0,0 +1,656 @@
...
+const CCIP_ADMIN_DATA = GOLDEN_MODULE_INTERFACE.encodeFunctionData('registerAdminViaGetCCIPAdmin', [
+  TOKEN,
+])
+const ACCESS_CONTROL_DATA = GOLDEN_MODULE_INTERFACE.encodeFunctionData(
+  'registerAccessControlDefaultAdmin',
+  [TOKEN],
+)
+
+// TAR-side selectors probed by pre-tx validation.
+const IS_REGISTRY_MODULE_SELECTOR =
```
aelmanaa, line 50: **Golden vectors** — module calldata encoded against hand-written independent Interface (not vendored ABI), so swapped args or wrong moduleFn would fail. Selector-aware stub records every call, verifying the right token getter was probed for each method.

aelmanaa (approved):
## ✅ APPROVE — EVM registerAdmin op (f097528)

Reviewed for type-safety, architecture, pre-tx gates, and the three registration paths (owner, ccip-admin, access-control-default-admin). Strong work — the op unifies three self-service auth methods, handles version dispatch safely, and adds a critical security fix to the base `EVMOperation` class.

---

## Strengths

**Unified registration paths with method-scoped pre-flight checks**
- Three methods (owner, ccip-admin, access-control-default-admin) each have their own getter (or role check) validation, proved by independent selector assertions in tests
- Token getter pre-flight (owner/ccip-admin) checks `admin == sender` before encoding, so authorization mismatches fail here with clear errors, not as on-chain reverts
- AccessControl role check mirrors the module's own logic (reads `DEFAULT_ADMIN_ROLE`, then `hasRole`), ensuring parity

**Version dispatch with compile-time safety**
- Discriminated union on `registryModuleVersion` (defaults to V1_6_0): v1.5.0 drops `access-control-default-admin` at the type level (compile error if used)
- On-chain version resolved and compared against declared version, so a wrong declaration fails with clear error before encoding

**Pre-tx validation gates (turning opaque reverts into clear errors)**
- Registry module registered check (`isRegistryModule`)
- Registry module version check (resolved, compared)
- Token not already registered (administrator + pendingAdministrator both zero)
- Pending proposal check (rejects silent overwrites)
- Authorization check per method (getter parity or role presence)
All gated ahead of encoding, all errors named and scoped.

**Security: senderBoundToWallet helper in EVMOperation base class**
- New guard prevents pre-tx checks authorizing one address while a different signer actually signs
- Validates wallet is a valid signer, then gates on wallet address == sender if given
- Prevents the "authorization leak" where pre-tx checks pass for address A, but signing wallet B signs and passes every local guard, reverting only on-chain
- Documented with example that guides toward `generateUnsigned*` when signer isn't known upfront
- Applies to all `EVMOperation` subclasses, not just registerAdmin

**Shared contracts.ts layer (token-admin-registry)**
- Cached interfaces per version (no per-call `new Interface`)
- Version resolvers that throw typed errors (type mismatch → `CCTContractTypeInvalidError`, unknown version → `CCTContractVersionUnsupportedError`)
- Mirrors the pattern from token-pool/contracts.ts; establishes the canonical structure for admin ops

**Test coverage (656 lines, 20+ test cases)**
- Golden vectors: Module calldata encoded against hand-written independent Interface, not the vendored ABI (catches swapped args, wrong moduleFn)
- Selector-aware stub: Records every call and its selector, verifies the right token getter was probed for each method
- Per-method authorization: owner getter, ccip-admin getter, access-control role, all independently validated
- Version dispatch: V1_5_0 rejects `access-control-default-admin`, V1_6_0 accepts it; version mismatch errors
- State machine: Not registered, pending, already-registered — all gated correctly

---

## Per-Method Analysis

**Method: owner (V1_5_0+)**
- Calls `registerAdminViaOwner(token)`
- Pre-flight: `token.owner()` must equal `sender`
- State gate: token must be unregistered (no active or pending admin)
- ✅ Verified in test "registers via owner" + selector probe

**Method: ccip-admin (V1_5_0+)**
- Calls `registerAdminViaGetCCIPAdmin(token)`
- Pre-flight: `token.getCCIPAdmin()` must equal `sender`
- State gate: same
- ✅ Verified in test "registers via ccip-admin" + selector probe

**Method: access-control-default-admin (V1_6_0 only)**
- Calls `registerAccessControlDefaultAdmin(token)`
- Pre-flight: `AccessControl(token).hasRole(DEFAULT_ADMIN_ROLE, sender)` must be true
- **Key insight**: The module registers `msg.sender`, NOT a derived address, so role check at pre-flight time ensures the caller (who will sign) can create the entry
- State gate: same
- ✅ Verified in test "registers via AccessControl" + role + selector probes

---

## Calldata Golden Vectors

Hand-encoded module calldata independently checked; swapping token/registryModule or wrong moduleFn would fail:
- owner: `registerAdminViaOwner(token)`
- ccip-admin: `registerAdminViaGetCCIPAdmin(token)`
- access-control: `registerAccessControlDefaultAdmin(token)`

All three verified against captured module calls in test.

---

## Version Safety

- Declared version (default V1_6_0) resolved on-chain at call time
- Mismatch → clear error (e.g. "v1.5.0 module, v1.6.0 declared")
- V1_5_0 type union excludes `access-control-default-admin` — compile error if misused
- V1_6_0 is the default for the optional param, so omitting it gives the right version for most callers

---

## Architecture: Base Class Security Fix

`EVMOperation.senderBoundToWallet(wallet, sender?)`
- Verifies wallet is a valid signer (not just any object)
- If `sender` is given, asserts it equals wallet address
- On execute, registerAdmin calls this before super.execute, ensuring pre-tx checks and signing happen for the same address
- **Why this matters**: Without it, a caller could:
  1. Pre-flight checks for address A (satisfies authorization)
  2. Pass a different wallet B to execute (signs as B)
  3. All local guards pass (sender was pre-flighted, wallet was validated in isolation)
  4. On-chain `msg.sender` is now B, revert if it expects A
- Now: if `sender` pre-flight is for A and wallet is B, execute throws locally before even signing

This is a critical fix for the M2 registry flow (registerAdmin → transferAdmin → acceptAdmin → setPool), where each step gates on a specific on-chain role. Excellent addition to the base class.

---

## Nits (non-blocking)

- **@example parity**: The `generateUnsignedRegisterAdmin` example is solid; might bring the `registerAdmin` example up to similar detail (show state result on success, e.g. "token now in pendingAdministrator state")

- **AccessControl role check replication**: The op correctly mirrors `AccessControlDefaultAdminRules` logic (reads DEFAULT_ADMIN_ROLE from AccessControl, not AccessControlDefaultAdminRules), but this deserves a comment linking to the Solidity contract so future maintainers understand why we don't call `defaultAdmin()` directly

- **senderboundToWallet docstring clarity**: Say explicitly that this guards against the "authorization leak" (pre-tx checks pass for one address, signing wallet is different). The docstring touches on it but could be bolder

---

## Test Gaps (optional follow-ups)

- Non-registry-module-owner address calling registerAdmin (TAR's `OnlyRegistryModuleOrOwner` revert is now well pre-flighted; confirming the module isn't in the set is good)
- Already-registered token with active admin (currently gated; could add explicit happy-path assertion of the rejection)

Both are covered by pre-flight validation and would be redundant to smoke-test against stubs, so these are polish-only.

---

## npm run check status

✅ Exit 0 — no type errors.

---

## Verdict

**APPROVE.** All three registration paths are correct, pre-tx gates are comprehensive, the senderBoundToWallet security fix in EVMOperation strengthens the base class for all write ops, and test coverage is solid. Ready for merge.

PR #335, opened 2026-08-06
feat(cct-sdk): Add get token admin registry config evm query

apedrob:
## What

- Add `getTokenAdminRegistry` query to EVMTokenManager. Reads a token's entry in the TokenAdminRegistry and returns its administrator, any pending administrator, and its registered pool (read-only, so there is no wallet and no tx)

## Why

- Gives callers a way to see the registry state before acting on it

## Testing

- Unit tests cover the read

## Notes

- `chain.getRegistryTokenConfig` cannot be used for this, because it throws when the administrator is the zero address, which is exactly the state a token sits in after registerAdmin and before acceptAdmin

### DIFF
--- a/ccip-sdk/src/cct/evm/index.test.ts
+++ b/ccip-sdk/src/cct/evm/index.test.ts
@@ -600,4 +600,72 @@ describe('EVMTokenManager (cct/evm)', () => {
       )
     })
   })
+  describe('getTokenAdminRegistry', () => {
+    const GET_TOKEN_CONFIG_IFACE = new Interface([
+      'function getTokenConfig(address token) view returns (tuple(address administrator, address pendingAdministrator, address tokenPool))',
+    ])
+    const ADMINISTRATOR = '0x' + '77'.repeat(20)
+
+    /**
+     * Chain stub whose provider answers `getTokenConfig` with `administrator`/zeroed others —
+     * but only for a call to `TAR` decoding to `TOKEN`. Mirrors the target/argument assertions in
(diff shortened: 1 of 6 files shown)


aelmanaa:
## ✅ APPROVE — `getTokenAdminRegistry` (EVM)

Head `c9b5697` · base `feature/DAPP-10819-evm-accept-admin`. **✅ APPROVE — 0 blockers.**

**Tested live on Sepolia** (TAR [`0x95F29FEE…B37B82`](https://sepolia.etherscan.io/address/0x95F29FEE11c5C55d26cCcf1DB6772DE953B37B82)):

- configured token [`0x6a2674…597FE`](https://sepolia.etherscan.io/address/0x6a2674C18B57a7664e63f11b3e372ff5740597FE) → `administrator=0x78AEb4…`;

- **unconfigured token → `administrator=0x0…0` reported, not thrown** (the gap this op fills vs the throwing `getRegistryTokenConfig`).

- Solana-parity: same op name/params/result base. 5/5 tests · `npm run check` exit 0.

aelmanaa:
✅ **Feedback implemented** (commit fb19f2f)

Added `@remarks` caveat to `GetTokenAdminRegistryResult` type alias documenting that `administrator` may be `ZeroAddress` and must be tested with `=== ZeroAddress`, not truthiness. Guides callers away from the silent-miss bug.

Type check: ✅ `npm run check` exit 0

aelmanaa (approved):

PR #360, opened 2026-08-18
feat: canton cct sdk for registry pools

stackman27:


ccip-sdk/examples/canton-pool-example.ts
```diff
@@ -0,0 +1,191 @@
+/**
```
RodrigoAD, line 1: We should move these scripts somewhere else. If it's only for local testing, let's remove them. If they are useful for testing, we can convert them in integration tests. If useful for users, we can turn them into documentation

ccip-sdk/CCT_CANTON_INTEGRATION.md
```diff
@@ -0,0 +1,51 @@
+# CCT Canton SDK — integration guide
```
RodrigoAD, line 1: Considering we can't keep the previous scripts, I don't think these docs make too much sense

ccip-sdk/src/canton/gateway-ledger-fetch.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/** Monotonic JSON-RPC id counter (Math.random is avoided to stay deterministic). */
+let jsonRpcId = 0
+
+/**
+ * Create a `fetch` that proxies ledger reads through the gateway `ledgerApi`
+ * method. Returns a function with the standard `fetch` signature so it can be
+ * passed straight to `CantonChain.fromUrl({ fetch })` / `createCantonClient`.
+ */
+export function createGatewayLedgerFetch(
```
RodrigoAD, line 45: This is only used by the new scripts. do we really need it?

If we do, let's please refactor it, it has many weird implementation decisions (`jsonRpcId` counter, the hardcoded `/livez` response...)

ccip-sdk/src/canton/networks.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Registry of well-known configuration per Canton chain ID
+ * (`canton:LocalNet` / `canton:DevNet` / `canton:TestNet` / `canton:MainNet`).
+ *
+ * `canton:TestNet` values verified against CV1 (prod testnet) deployment
+ * state, Aug 2026. On CV1 the operator parties share one fingerprint across
+ * the ccipOwner/rmnOwner/ccvOwner hints.
+ *
+ * LocalNet/DevNet have no stable deployments — pass explicit overrides there.
+ */
+export const CANTON_NETWORKS: Readonly<Record<string, CantonNetworkConfig>> = {
```
RodrigoAD, line 53: We should avoid hardcoded addresses or URLs. I see it's only used in a test, but why would we test against testnet or mainnet? I think we should remove this

ccip-sdk/src/canton/index.ts
```diff
@@ -670,6 +681,273 @@ export class CantonChain extends Chain<typeof ChainFamily.Canton> {
...
+   * by contract ID (the ACS stream can surface the same contract via multiple
+   * parties).
+   *
+   * @param templateId - Full template ID (`#<pkg>:<Module>:<Entity>`).
+   * @param parties - Parties whose ACS visibility to query.
+   * @param match - Predicate over the decoded `createArgument`. Pass `() => true`
+   *   to return every active contract of the template.
+   * @returns All matching active contracts (may be empty).
+   */
+  async findActiveContractsByTemplate(
```
RodrigoAD, line 694: This is almost a duplicate of `findActiveContractByTemplate`. Let's reuse it here

ccip-sdk/src/canton/index.ts
```diff
@@ -670,6 +681,273 @@ export class CantonChain extends Chain<typeof ChainFamily.Canton> {
...
+   * query state via non-consuming choices rather than ACS snapshots.
+   *
+   * @param templateId - Full template ID of the target contract.
+   * @param contractId - Target contract ID.
+   * @param choice - Read choice name (e.g. `GetRequiredCCVs`, `Get`).
+   * @param choiceArgument - Daml choice argument record.
+   * @param caller - Acting party (`actAs`).
+   * @returns The raw `exerciseResult` value, or `null` if no matching exercised event.
+   */
+  async submitReadChoice(
```
RodrigoAD: No callers of this one. We can remove it?

ccip-sdk/src/canton/index.ts
```diff
@@ -670,6 +681,273 @@ export class CantonChain extends Chain<typeof ChainFamily.Canton> {
...
+   * The returned {@link CantonActiveContract} carries `createdEventBlob` +
+   * `synchronizerId` so the caller can embed it as a disclosed contract in a
+   * later submission — this is what unblocks real `execute()` for CCT ops.
+   *
+   * @param templateId - Full template ID, e.g. `#ccip-core-v2:CCIP.CoreV2.TokenAdminRegistry:TokenAdminRegistry`.
+   * @param parties - Parties whose ACS visibility to query (ActAs first, then ReadAs).
+   * @param match - Predicate over the decoded `createArgument` record.
+   * @returns The first matching active contract, or `null` if none match.
+   */
+  async findActiveContractByTemplate(
```
RodrigoAD: This isn't used either

ccip-sdk/src/canton/index.ts
```diff
@@ -1406,7 +1684,7 @@ export class CantonChain extends Chain<typeof ChainFamily.Canton> {
    *   2. Decode the hash and call `signer.sign(hashBytes)`.
    *   3. Execute the signed transaction (`/v2/interactive-submission/executeAndWaitForTransaction`).
    */
-  private async submitCommands(
+  async submitCommands(
```
RodrigoAD: Making this public allows users to send arbitrary txs, allowing them to bypass validations for the execute/send commands, and making the SDK surface way bigger and confusing. Let's make it private again. We can extract the logic somewhere else if needs to be reused

ccip-sdk/src/cct/canton/index.ts
```diff
@@ -0,0 +1,158 @@
...
+   */
+  async generateUnsignedDeployTokenPool(
+    opts: GenerateDeployTokenPoolParams,
+  ): Promise<GenerateDeployTokenPoolResult> {
+    return this.#deployTokenPool.generate(this.chain, opts)
+  }
+
+  /** Atomically deploys and initializes a `BurnMintTokenPool`/`LockReleaseTokenPool` (registry-pools family). */
+  async deployTokenPool(opts: ExecuteDeployTokenPoolParams): Promise<ExecuteDeployTokenPoolResult> {
+    return this.#deployTokenPool.execute(this.chain, opts) as Promise<ExecuteDeployTokenPoolResult>
```
RodrigoAD: We can't do these types of casting. It's false that the response has a `poolCid` field declared. It happens more time in this class

RodrigoAD (changes requested):
Left some comments. The `cct/canton` is quite big, I couldn't go through all. 

As a general comment, let's please reduce the verbosity of the comments.

ccip-sdk/src/canton/networks.ts
stackman27: Not test-only, and not a live-network test either: `getCantonNetworkConfig('canton:TestNet')` is just an in-memory lookup on the `CANTON_NETWORKS` table, and `resolvePoolFactoryDeps` uses it in
real production code to default deployTokenPool's deps on known networks. 

Removing it would force every caller to pass tokenAdminRegistry/feeQuoter/rmnRemote explicitly, even on TestNet/MainNet. I'd rather keep it

ccip-sdk/src/canton/gateway-ledger-fetch.ts
stackman27: Kept it (examples still need it), refactored both issues: jsonRpcId is now per-instance instead of module-global, and /livez genuinely round-trips through the gateway instead of faking a 200

ccip-sdk/examples/canton-pool-example.ts
stackman27: Keeping these as user-facing examples/docs rather than removing or converting to integration tests,  I think this is genuinely useful reference material for integrating via the Wallet Gateway, not just local scratch scripts

ccip-sdk/src/canton/index.ts
stackman27: extracted it into a standalone submitCantonCommands() function so CantonChain no longer exposes it at all, while `operation.ts` (which needed cross-module access) can still call the same logic

ccip-sdk/src/cct/canton/index.ts
stackman27: fixed, `DeployTokenPool/ApplyChainUpdates` now override execute() to actually extract poolCid (and rateLimiterCids/tokenConfigCid/poolInstanceAddress) from the transaction response, throwing if the ledger didn't produce what's expected

ccip-sdk/CCT_CANTON_INTEGRATION.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+9 checks (pool config, TAR registration, lane addresses + CCVs + limiter
+references, factory wiring, all 3 rate limiters); exits non-zero on any failure.
+
+## What the integrator provides
+
+1. **Signer**: wallet connection consuming `unsigned.commands` (`{ commands, actAs, disclosedContracts, commandId }`) → `prepareExecute` → approve → execute.
+2. **Ledger reads at compose time**: the caller's connection needs `CanReadAs` over the issuer party (composition resolves TAR/pool contracts via ACS).
+3. **Network constants** (if not CV1 testnet): `ccipOwner`, TAR/FeeQuoter/RMNRemote addresses, EDS + ledger URLs — add to `networks.ts` or pass via env (`TAR_RAW`, `FEE_QUOTER_RAW`, `RMN_REMOTE_RAW`).
```
RodrigoAD: > pass via env (`TAR_RAW`, `FEE_QUOTER_RAW`, `RMN_REMOTE_RAW`).

This isn't true currently. I think we should make more use of the `canton-config.json`

ccip-sdk/src/canton/explicit-disclosures/eds.ts
```diff
@@ -235,6 +235,40 @@ export class EdsDisclosureProvider {
...
+        '/ccip/v1/global/disclosure/batch',
+        EDS_HEADERS,
+        this.timeoutMs,
+        { addresses: [instanceAddress] },
+      )
+      const contract = resp.disclosures?.[0]
+      if (!contract?.contractId || !contract.createdEventBlob) return null
+      return edsContractToSdk(contract)
+    } catch {
+      return null
```
RodrigoAD: Let's return the error here instead

ccip-sdk/src/cct/canton/token-pool/shared.ts
```diff
@@ -0,0 +1,499 @@
...
+ * Normalize an address for on-ledger storage in a pool's remote chain config:
+ * strip an optional `0x` prefix and left-pad 20-byte (EVM) addresses to 32
+ * bytes. Canonical form is 32-byte-padded bare hex: inbound executes compare
+ * the stored value verbatim against the message's 32-byte-padded
+ * sourcePoolAddress/sourceTokenAddress, while outbound sends strip the padding
+ * again for sub-32-byte destination chains (validateDestChainAddress).
+ */
+export function normalizeRemoteAddress(s: string): string {
+  const bare = s.startsWith('0x') ? s.slice(2) : s
+  return bare.length === 40 ? bare.padStart(64, '0') : bare
```
RodrigoAD: This doesn't guarantee 32 B if we pass any different than 20 B (40 char)

ccip-sdk/src/cct/canton/token-pool/operations/deploy-token-pool.ts
```diff
@@ -0,0 +1,363 @@
...
+        {
+          CreateAndExerciseCommand: {
+            templateId,
+            createArguments,
+            choice: 'Initialize',
+            choiceArgument,
+          },
+        },
+      ],
+      commandId: `cct-deploy-${p.poolType}-pool-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`,
```
RodrigoAD:
```suggestion
      commandId: `cct-deploy-${p.poolType}-pool-${randomUUID()}`,
```

I see many places with this logic. If there's some concurrency, this could cause some collision. I'd recommend using `randomUUID` for it. You do it already in tests

ccip-sdk/src/cct/canton/token-pool/operations/get-rate-limiter-state.test.ts
```diff
@@ -0,0 +1,168 @@
...
+  return {
+    network: { family: ChainFamily.Canton },
+    logger: { debug() {}, info() {}, warn() {}, error() {} },
+    async findActiveContractByInstanceAddress(
+      _t: string,
+      instanceAddress: string,
+    ): Promise<CantonActiveContract | null> {
+      return contract && instanceAddress === RL_INSTANCE_ADDRESS ? contract : null
+    },
+  } as unknown as CantonChain
```
RodrigoAD: We shouldn't do this. Why doesn't `as CantonChain` work? Let's fix that instead

stackman27: fixed,  mocks now build a real instance via Object.assign(Object.create(CantonChain.prototype), {...}), no cast anywhere (as CantonChain can't work since the class has private fields)

ccip-sdk/src/cct/canton/token-pool/shared.ts
stackman27: fixed, now always pads to 32 bytes and throws on non-hex or >32B input instead of passing through

RodrigoAD (approved):

PR #368, opened 2026-08-24
feat(ccip-sdk): OIDC authentication for Canton Ledger API

SyedAsadKazmi:
Adds OAuth 2.0 / OIDC auth for Canton, split across `ccip-sdk` (runtime-agnostic protocol primitives) and `ccip-cli` (Node-specific orchestration).

**SDK — runtime-agnostic core** (`ccip-sdk/src/canton/authentication/`):
- `oauth4webapi` for OAuth2 protocol mechanics (discovery, PKCE, grants, token validation) — pure `fetch`/WebCrypto, no `node:*` imports
- `createMemoizedTokenFetcher()` — token caching with concurrent fetch coalescing and expiry-based invalidation, backed by `micro-memoize` (`{ async: true }`)
- `createAuthProvider()` discriminated union selector (static / clientCredentials)
- Authorization-code protocol helpers: `buildAuthorizationRequest()`, `validateAuthorizationCallback()`, `exchangeAuthorizationCode()`, `refreshAuthorizationCodeToken()`
- `AuthorizationCodeProvider` — caching provider wrapping caller-supplied fetch/refresh callbacks
- `CantonConfig.jwt` accepts `string | (() => Promise<string>)` — static token or per-request getter for refreshable tokens
- `CANTON_AUTH_ERROR` error code + recovery hint
- 36 unit tests

**CLI — Node orchestration** (`ccip-cli/src/providers/canton/`):
- `auth.ts` — local `node:http` callback server, `open`/`xdg-open` browser launching, `CANTON_CLIENT_ID`/`CANTON_CLIENT_SECRET` env-var resolution, process-wide auth provider cache (prevents double login across `send` → `showRequests`)
- `config.ts` — `loadCantonConfig` accepts `auth`; `jwt` optional when `auth` present. CLI resolves `auth` upfront into `jwt` (string or getter) before handing config to the SDK
- `wallet.ts` — Ed25519 transaction signer + wallet loading
- Error surfacing: `Promise.allSettled` results inspected to surface `CCIPError` instead of generic `RPC_NOT_FOUND`


ccip-api-ref/docs-cli/configuration.mdx
```diff
@@ -281,22 +282,70 @@ Canton operations require a JSON config file passed via `--canton-config <path>`
...
+export CANTON_CLIENT_ID="my-client-id"
+export CANTON_CLIENT_SECRET="my-client-secret"
+```
+
+**Authorization code example** (interactive browser login for human users):
+
+```json
+{
+  "auth": {
+    "type": "authorizationCode",
```
friedemannf, line 323: Doesn't the `authorizationCode` flow still require a ClientID? Might make sense to add this to the docs here just like for `clientCredentials` just above

ccip-api-ref/docs-sdk/guides/manual-execution.mdx
```diff
@@ -309,19 +309,19 @@ const dest = await CantonChain.fromUrl('https://ledger.example.com/api/json', {
 // Execute by message ID (CCIP API provides verifications)
 const execution = await dest.execute({
   messageId: '0x1234...abcd',
-  wallet: cantonWallet,
+  wallet: { party: 'receiver::1220...' }, // Canton wallet (party ID, no signer needed)
```
friedemannf: Does this actually work? `ccip-cli/src/providers/canton.ts` has removed the wallet from `loadCantonWallet` entirely.

Also, this would be the same `party` as already specified in `cantonConfig`?

ccip-cli/README.md
```diff
@@ -293,8 +296,43 @@ Canton requires a config file with connection parameters via `--canton-config <p
...
+  "transferInstructionUrl": "https://transfer-instruction.example.com"
+}
+```
+
+```bash
+export CANTON_CLIENT_ID="my-client-id"
+export CANTON_CLIENT_SECRET="my-client-secret"
+```
+
+When `auth` is set, the CLI resolves a JWT on demand via the SDK's `@chainlink/ccip-sdk` authentication package (backed by [`oauth4webapi`](https://github.com/panva/oauth4webapi)). If both `jwt` and `auth` are present, `jwt` takes precedence.
```
friedemannf: Is the plan to eventually deprecate the `jwt` field? I get that we should leave it in to not do a breaking change, but setting `jwt` is the same as `static`, right?

ccip-sdk/src/canton/authentication/static.ts
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * These wrap a pre-obtained JWT in a {@link StaticTokenSource} (no refresh),
+ * mirroring the Go `authentication.NewStaticProvider` /
+ * `authentication.NewInsecureStaticProvider`.
+ *
+ * The TLS-vs-insecure distinction in the Go providers only affects gRPC
+ * transport credentials; the TS SDK talks to the Canton JSON Ledger API over
+ * HTTP/2 where TLS is negotiated by the runtime, so both variants collapse to
+ * the same token-source implementation here. The `type` field is preserved on
+ * the provider so callers can distinguish them for logging/validation.
```
friedemannf: Please don't just copy the Go implementation blindly, AI even mentions that `insecureStatic` has no effect as it is only related to http2/gRPC.

Just remove `insecureStatic`

ccip-sdk/src/canton/authentication/types.ts
```diff
@@ -0,0 +1,167 @@
...
+ * Shared types for the Canton authentication providers.
+ *
+ * @packageDocumentation
+ *
+ * This module mirrors the Go `chainlink-canton/deployment/authentication` package,
+ * providing OAuth 2.0 authentication for the Canton Ledger API (JSON API / HTTP).
+ *
+ * Four auth schemes are supported (matching `commonconfig/auth.go`):
+ * - `static`           — a pre-obtained JWT (TLS transport)
+ * - `insecureStatic`   — a pre-obtained JWT (no TLS, for local devnet)
```
friedemannf: Overall, all the comments on TLS don't really apply here, in the sake of reducing confusion let's remove all `TLS` references

ccip-sdk/src/canton/authentication/types.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+ * An authentication provider for the Canton Ledger API.
+ *
+ * Mirrors the Go `authentication.Provider` interface, adapted for HTTP/JSON
+ * API usage (the TS SDK talks to the Canton JSON Ledger API over HTTP/2, not gRPC,
+ * so transport/per-RPC credentials collapse into a single token source).
+ */
+export interface AuthProvider {
```
friedemannf, line 52: As mentioned in the comment, Transport Credentials don't apply over HTTP, is this split in `TokenSource`/`AuthProvider` really needed?

There really is no need to copy the Go implementation as-is, for Go we're interfacing with gRPC and need additional transport credentials options.

ccip-api-ref/docs-cli/configuration.mdx
SyedAsadKazmi: Good catch, yes, `authorizationCode` requires a non-empty `clientId` (validated in `authorization-code.ts` and resolved from `CANTON_CLIENT_ID` via `mergeAuthEnvVars`). 

Added the `export CANTON_CLIENT_ID` example to match the `clientCredentials` example above it.

ccip-api-ref/docs-sdk/guides/manual-execution.mdx
SyedAsadKazmi: You're right on both. The SDK example works (`CantonChain.execute` validates the wallet via `isCantonWallet` and uses `wallet.party` for `actAs`/`payer`), and the CLI's `loadCantonWallet` sources `party` entirely from `cantonConfig.party` , so it's the same value. 

Fixed the example to extract `cantonConfig` into a variable and reuse `cantonConfig.party` instead of duplicating the string literal.

ccip-cli/README.md
SyedAsadKazmi: Correct, the top-level `jwt` is shorthand for `auth: { type: "static", jwt }` (`createAuthProvider` defaults to `Static` when `type` is omitted). And yes, it should be left as is, to avoid breaking change.

Added a note stating the equivalence and that `jwt` takes precedence when both are present.

ccip-sdk/src/canton/authentication/static.ts
SyedAsadKazmi: Done. Removed `insecureStatic` entirely: `InsecureStaticProvider`, `createInsecureStaticProvider`, the `AuthType.InsecureStatic` constant, the switch case, re-exports, tests, and doc entries. 

As you noted, the TLS distinction only matters for gRPC transport credentials and has no effect over the HTTP/2 JSON API.

ccip-sdk/src/canton/authentication/types.ts
SyedAsadKazmi: Done. Stripped all TLS/gRPC/transport-credentials references from code comments and docs. The `allowInsecureRequests` option in `token-source.ts` was kept, that's a real `oauth4webapi` option for allowing HTTP (non-HTTPS) OAuth endpoints in tests, not a TLS transport comment.

ccip-sdk/src/canton/authentication/types.ts
SyedAsadKazmi: Agreed, the split was a blind copy of Go's design where gRPC needs separate transport credentials. Over HTTP there's nothing in between, so every consumer just did `provider.tokenSource().token()`. Collapsed `AuthProvider` to expose `token()` directly. `CachingTokenSource`/`StaticTokenSource` remain as internal primitives (the auth-code refresh flow still needs `getCachedToken()`), but `TokenSource` is no longer part of the public API.

ccip-api-ref/docs-sdk/guides/manual-execution.mdx
friedemannf: Ah, I mixed up CLI and SDK. The SDK still accept and expects this `wallet` `{party: 'asdf'}`. Your change still looks good 👍

friedemannf (dismissed):

.oxlintrc.json
```diff
@@ -48,7 +48,7 @@
       "excludeFiles": ["**/*.test.ts", "**/__tests__/**", "**/__mocks__/**"],
       "rules": {
         "ccip/restricted-syntax": "error",
-        "import/no-nodejs-modules": ["error", { "allow": ["buffer"] }]
+        "import/no-nodejs-modules": ["error", { "allow": ["buffer", "node:child_process", "node:http"] }]
```
andrevmatos: No node modules on ccip-sdk, pls

andrevmatos (changes requested):
Sorry, but we can't have this much node dependencies in the ccip-sdk;
I can see 2 usecases for this oauth worfklow for Canton: ccip-cli and an external provider (e.g. think of someone who manages to pack the Canton lib on their web-based or Electron-based app).
- For ccip-cli, this whole server-spawning and browser `open` call should be handled on the cli side, and then whatever result of that needed is then handed over the `cantonConfig`
- For (web?) clients embedding this workflow, there may be an intersection of the above which they may be interested on, but they probably would implement their own callback REST endpoint, and/or some form of redirect on their application.

Strictly speaking, the SDK should be handed only what it needs, and these previous steps are mostly utilities. But the way I see this, we may be able to extract the reusable core intersection of the above to some functions which MAY be exported from the ccip-sdk, and implement the nodeJS-specific bits in the cli, maybe make `providers/canton/` a folder.

andrevmatos:
Concrete proposal for the split:

**SDK — runtime-agnostic core** (`canton/authentication/`):
keep `types`, `metadata` (RFC 8414 discovery), `token-source` (caching + PKCE via WebCrypto), `client-credentials`, `static` — all pure `fetch`, no `node:*`, no `process.env`. Revert the `.oxlintrc.json` allowances; the docusaurus webpack stubs then become unnecessary too. For auth-code, export only the protocol pieces an embedder can't safely rewrite: build-authorize-URL, callback validation (state + PKCE), code→token exchange, refresh grant. No callback server, no browser open.

**CLI — Node orchestration** (`providers/canton/` folder):
owns everything environment-specific: local `node:http` callback server, `open`/`xdg-open`, `CANTON_CLIENT_ID`/`CANTON_CLIENT_SECRET` resolution, timeouts and terminal UX. It resolves auth upfront and hands the result to `cantonConfig` — the SDK never orchestrates a flow, it only consumes what it's given.

**Wiring:** `CantonConfig` takes `jwt: string`, or — if we want refresh to actually work (today `fromUrl` resolves once and every client holds a static string, so the refresh path is unreachable) — an injected `() => Promise<string>` token getter the clients call per request. Web/Electron embedders compose the SDK protocol functions with their own redirect/callback handling.

ccip-sdk/src/canton/transfer-instruction/client.ts
```diff
@@ -73,8 +73,20 @@ export interface TransferInstructionErrorResponse {
...
+   * a static token. Prefer `tokenGetter` when the token may expire.
+   */
   jwt?: string
+  /**
+   * Optional token getter invoked per request to obtain a fresh JWT.
+   *
+   * When set, each request awaits the getter and uses the returned JWT in the
+   * `Authorization` header (enabling automatic refresh).
+   */
+  tokenGetter?: () => Promise<string>
```
andrevmatos: Is the result of `tokenGetter` used exactly like `jwt`? If so, I'd say it'd be a simpler interface to simply type `jwt?: string | () => Promise<string>`, i.e. if a string, it's static; otherwise, it's async-dynamic, and just called whenever needed, like right before a request, to build a dynamic jwt

ccip-sdk/src/canton/authentication/token-source.ts
```diff
@@ -0,0 +1,187 @@
...
+ * caller-supplied `fetcher` when the cached token is expired or missing.
+ *
+ * The first `token()` call fetches; subsequent calls return the cached value
+ * until it expires, at which point a new fetch is triggered. Concurrent callers
+ * share a single in-flight fetch promise to avoid duplicate token requests.
+ */
+export class CachingTokenSource implements TokenSource {
+  private current: AccessToken | undefined
+  private readonly fetcher: () => Promise<AccessToken>
+  private inFlight: Promise<AccessToken> | undefined
```
andrevmatos: I think this whole `TokenSource` thing can be replaced by a simple `memoize` call, with `{ async: true, expires: ... }`; it already handles promises, returning the single inflight promise ootb, and removing the "cache" if promise rejects. You can also simply provide it a `() => string` in the static case

andrevmatos (approved):
Good job!

PR #372, opened 2026-08-25
test: classify networked suites and lock live-network runs

andrevmatos:
## What

A small test-infra refactor: every networked test suite is classified by filename suffix and serializes its live-network access through OS-level locks, so `node --test`'s concurrent file runs stop rate-limiting shared public RPCs.

## Changes

- **Naming** — CLI e2e suites renamed `show.test.ts → show.e2e.test.ts`, `lane.test.ts → lane.e2e.test.ts`. Test files are classified purely by suffix (`*.integration.test.ts`, `*.e2e.test.ts`, `*.fork.test.ts`), so a bare `integration.test.ts` counts as integration exactly like `logs.integration.test.ts`.
- **`npm run test:unit`** — new script (root + both workspaces) running only offline unit tests via `scripts/run-tests.mjs` (filename-only filtering; `--exclude`/`--tags`/`--dry`). `npm test` is unchanged and still runs everything.
- **Network locks** — `scripts/useResource.ts`. Networked suites declare the networks they talk to:
  ```ts
  import { useResource } from '../../../scripts/useResource.ts'
  await useResource(['sepolia', 'fuji'])
  ```
  Each tag is a directory under `<tmpdir>/ccip-tools-ts-network-locks`, created atomically with `fs.mkdir`; exactly one process holds a tag at a time, and a waiting suite starts the moment the holder finishes — there is no batch/wave scheduling, the filesystem is the queue. Acquisition is all-or-nothing (deadlock-free), locks auto-release on process exit, and crashed holders self-heal via pid-liveness (a lock held by a live process is never stolen). All 13 networked suites are wired with tags verified against the endpoints they actually call.
- **CI** — unchanged: `npm test` still runs everything in one job, and network contention is now resolved by the locks themselves (locks are per-machine, so the networked suites intentionally stay on the single CI runner rather than splitting across jobs).
- **Docs** — `CONTRIBUTING.md` gained "Test Suite Layout" + "Networked tests and the `useResource` lock" sections; `CHANGELOG.md` Unreleased entry added.

## Validation

- `npm run check` clean (lint 0 warnings, typecheck, `generate:lint` up-to-date)
- `npm run test:unit` — 1313 tests, 0 failures
- Full `npm test` (CI parity) — 1327 sdk + 117 cli, 0 failures; lock contention observed and resolved during the run (e.g. `waiting for "fuji"` heartbeats), zero lock residue afterwards
- Fork suites self-skip without anvil/surfpool (existing behavior; CI installs both)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,68 +14,71 @@ jobs:
   build-and-test:
     runs-on: ubuntu-latest
     steps:
-    - name: Check out repository
-      uses: actions/checkout@v5.0.0
+      - name: Check out repository
+        uses: actions/checkout@v7.0.1
 
-    - name: Set up Node.js
-      uses: actions/setup-node@v6.0.0
-      with:
-        node-version: 24
+      - name: Set up Node.js
+        uses: actions/setup-node@v7.0.0
+        with:
+          node-version: 24
 
-    - name: Install Foundry (anvil for fork tests)
-      uses: foundry-rs/foundry-toolchain@8789b3e21e6c11b2697f5eb56eddae542f746c10 # v1.7.0
+      # Older foundry-toolchain pins ran foundryup via `bash` and broke when it became
+      # a compiled binary ("cannot execute binary file"); v1.9.1 executes it directly
+      # and pins the foundryup installer internally, so no vendor script is piped here.
+      - name: Install Foundry (anvil for fork tests)
+        uses: foundry-rs/foundry-toolchain@908c540300062bd5a7e473851cdb4282204cee09 # v1.9.1
 
-    - name: Verify anvil is available
-      run: anvil --version
+      - name: Verify anvil is available
+        run: anvil --version
 
-    - name: Install Surfpool v1.1.2 (for SVM fork tests)
-      run: |
-        curl -sSL https://github.com/solana-foundation/surfpool/releases/download/v1.1.2/surfpool-linux-x64.tar.gz \
-          | tar xz -C /usr/local/bin
+      - name: Install Surfpool v1.1.2 (for SVM fork tests)
+        run: |
+          curl -sSL https://github.com/solana-foundation/surfpool/releases/download/v1.1.2/surfpool-linux-x64.tar.gz \
+            | tar xz -C /usr/local/bin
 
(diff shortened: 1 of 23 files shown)


aelmanaa (dismissed):

aelmanaa (approved):

andrevmatos:
Addressed all review points in `ee7c30bb`:

**Copilot (suppressed)**
- `release()` idempotency — fixed: each handle now tracks its own released state; a second release is a no-op and can no longer delete a lock another process re-acquired.
- Pre-held resource refcounting — fixed: every requested resource (pre-held or freshly acquired) is counted against each handle, so a handle's release withdraws exactly its own share.
- `show.e2e` missing `api` — fixed: the suite now also holds the `api` lock (the `show` command resolves 32-byte hashes through the default CCIP API).

**Copilot (inline)**
- TOCTOU in stale-steal (`useResource.ts:155`) — fixed with an atomic rename-to-tombstone: only one waiter can claim a dead lock (a concurrent stealer's rename fails ENOENT), so nobody can delete another's freshly re-created live lock. Abandoned tombstones are swept in `ensureLockRoot`.
- `run-tests.mjs` suffix boundary — fixed: a name classifies as integration only when it *is* `integration.test.ts` or ends with a dot-delimited `.integration.test.ts` (`notintegration.test.ts` no longer matches).
- `curl | bash` supply chain — agreed and fixed: reverted to the SHA-pinned `foundry-rs/foundry-toolchain@908c540… (# v1.9.1)`. Verified v1.9.1 executes the now-compiled `foundryup` directly (`run(foundryup, args)`, no `bash`) and pins its foundryup installer internally — the "cannot execute binary file" breakage was specific to older pins.

**CodeQL (github-advanced-security[bot])**
- js/insecure-temporary-file — fixed: the owner marker is written with `mode: 0o600` (owner-only, per the rule's own `isSecureMode` predicate), on top of the existing owned-dir/symlink validation. CodeQL now passes.

First CI attempt hit transient upstream flakiness (staging API HTTP 503 on the latency endpoint; toncenter 429s on the TON v2 walk); the rerun on the same commit is fully green.

michelemin (approved):
Approving publish.yml

PR #377, opened 2026-08-26
Fix/publish

andrevmatos:
- v1.13.0 failed to publish due to some env var required by a transient dependency on `npm ci`
- also, include a quick hotfix for `fetch` AdaptiveSync spamming on aptos paths methods


aelmanaa (approved):

michelemin (approved):
Approving publish.yml

PR #390, opened 2026-08-30
feat(cli): fetch CCV attestations directly from a verifier's aggregator

aelmanaa:
This pull request introduces major enhancements to the CCIP CLI and its documentation, focusing on flexible and robust support for CCV attestation sources during manual execution. It adds new CLI options and backend logic to fetch CCV attestations directly from verifiers or to supply them manually, ensuring messages can be executed even when attestations are not available via the default indexers or API. The CLI builder and documentation are updated to reflect these new capabilities, and tests are added to ensure correct grouping and rendering of options.

**Manual Execution & Verification Enhancements:**

- Added `--verifier`/`--verifier-endpoint` and `--ccv-data` CLI options, allowing users to fetch CCV attestations directly from verifier aggregators over gRPC or supply them out-of-band, ensuring execution is possible even when indexers lack the required CCV. The command logic ensures that these sources are used as fallbacks only when managed sources are insufficient, and prevents unverified execution that would revert onchain. [[1]](diffhunk://#diff-9b7385a4a8b934b4c3c047eb7f85e8cf4b1167e3e732954065caabf0d0d22deeR222-R294) [[2]](diffhunk://#diff-9b7385a4a8b934b4c3c047eb7f85e8cf4b1167e3e732954065caabf0d0d22deeR369-R417) [[3]](diffhunk://#diff-d031752a3dc854d38b0b2f7ba015c3d626d241205c336dfe5e0821d4656bff05R116-R164) [[4]](diffhunk://#diff-9b7385a4a8b934b4c3c047eb7f85e8cf4b1167e3e732954065caabf0d0d22deeR41) [[5]](diffhunk://#diff-9b7385a4a8b934b4c3c047eb7f85e8cf4b1167e3e732954065caabf0d0d22deeR26)

- Updated CLI option parsing and error handling to robustly process new verification flags, including improved coercion and validation for `--verifier`, `--ccv-data`, and `--indexer`.

**Documentation & CLI Builder Updates:**

- Expanded documentation for manual execution to detail the new verification source options, their formats, usage scenarios, and fallback behaviors. [[1]](diffhunk://#diff-f8054793b86b61f31b6f5670e3355675827311a85511663ce150884599dbf490R38-R85) [[2]](diffhunk://#diff-d509fd26ff682f49f0c054dcca5422c10bed45e76de696409168b0c5f5e2240bL212-R212) [[3]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR28-R29)

- Updated the CLI Builder UI to include a new "Verification Sources" group, ensuring the new options are visible and properly grouped in the interface. Added tests to guarantee correct rendering and grouping. [[1]](diffhunk://#diff-91fd72e2e3d04038423a0eb30bdf532c4dbb598118b6446d6092cacf2ec9d721R28-R56) [[2]](diffhunk://#diff-c12b5a66d6bc0194d0d25e3e729b3c985f2b459ed4570c56359fa67cf4f86893R54-R57) [[3]](diffhunk://#diff-2e37d950b8c9ec963fb3efd2dc511b970d563eb090c4139fe9212c283838d034L138-R139) [[4]](diffhunk://#diff-2e37d950b8c9ec963fb3efd2dc511b970d563eb090c4139fe9212c283838d034R23) [[5]](diffhunk://#diff-7a4272b06f8bed5b1b2f7059561b8a9c7175090df41098875758f6d937a846a0R1-R16) [[6]](diffhunk://#diff-1bfb49d5a9e9ec02b551b0f2b0e393e3594242650d761bcad513b4649d47b3b2R1-R15)

**Build & Dependency Changes:**

- Added new dependencies for gRPC support (`@grpc/grpc-js`, `@grpc/proto-loader`) and updated the build process to include necessary protocol buffer assets. [[1]](diffhunk://#diff-519884ba22c306f1ee7807a3c4296720af372fa1175f9b75350d8f55aea9088bR58-R59) [[2]](diffhunk://#diff-519884ba22c306f1ee7807a3c4296720af372fa1175f9b75350d8f55aea9088bL31-R36)

These changes significantly improve the flexibility and reliability of CCIP manual execution, especially for advanced or recovery scenarios where attestations are not available from standard sources.

### DIFF
--- a/ccip-cli/src/commands/manual-exec.ts
+++ b/ccip-cli/src/commands/manual-exec.ts
@@ -23,6 +23,7 @@ import {
   type CCIPRequest,
   type Chain,
   CCIPAPIClient,
+  CCIPArgumentInvalidError,
   CCIPDestSimulationUnavailableError,
   CCIPInteractiveRequiredError,
   CCIPMessageIdNotFoundError,
@@ -37,6 +38,7 @@ import type { Argv } from 'yargs'
 
 import type { GlobalOpts } from '../index.ts'
 import { fetchChainsFromRpcs, loadChainWallet, resolveIndexer } from '../providers/index.ts'
+import { assertCoverage, collectDirectVerifications } from '../verifiers/direct.ts'
 import { type Ctx, Format } from './types.ts'
 import {
   getCtx,
@@ -217,16 +219,79 @@ async function manualExec(
   if (argv.estimateGasLimit != null && !source)
     source = await getChain(request.lane.sourceChainSelector)
 
+  // The verification block below lives under `if (source)`, and a bare messageId positional never
+  // sets `source`. Without this, --verifier/--ccv-data were silently ignored and execution fell
+  // through to `dest.execute({ messageId })` with NO verifications, broadcasting an unverified
+  // execute that reverts RequiredCCVMissing and costs gas. Resolve the source chain so the
+  // verifier fetch and the coverage assert actually run.
+  const verifierEntries = (argv as { verifier?: readonly string[] }).verifier ?? []
+  const ccvDataEntries = (argv as { ccvData?: readonly string[] }).ccvData ?? []
+
+  // A messageId positional does not set `source`, so resolve it here to reach the verifier fetch
+  // and the coverage assert. It is optional: a v2 execution needs only the encoded message and the
+  // OffRamp, both of which the CCIP API returns, so a missing source RPC falls through to the
+  // API-only branch below rather than failing.
+  if (!source && (verifierEntries.length > 0 || ccvDataEntries.length > 0))
+    source = await getChain(request.lane.sourceChainSelector).catch(() => undefined)
+
   let inputs
   if (source) {
     offRamp ??= await discoverOffRamp(source, dest, request.lane.onRamp, source)
     const indexer = resolveIndexer(argv, dest, logger, source)
-    const verifications = await dest.getVerifications({
-      ...argv,
-      indexer,
-      offRamp,
-      request,
-    })
+
+    // Source order: the managed sources first, then the verifiers themselves. `--verifier` /
+    // `--ccv-data` are a FALLBACK used only when the API/indexer could not cover the required set.
+    // That keeps routine calls off partner infrastructure and only discloses the messageId to a
+    // verifier operator when there is no alternative. For a CCV no indexer has onboarded the
+    // managed sources never cover it, so the fallback always fires.
+    let verifications
+    if (verifierEntries.length > 0 || ccvDataEntries.length > 0) {
+      verifications = await dest
+        .getVerifications({ ...argv, indexer, offRamp, request })
+        .catch(() => undefined)
+      // Only families whose policy is keyed by DEST address can be coverage-checked here.
+      // Canton derives `requiredCCVs` from the source-side `message_ccv_addresses` while its
+      // results carry dest addresses, so comparing them would reject a complete attestation set.
+      const destKeyedPolicy =
+        verifications &&
+        'verificationPolicy' in verifications &&
+        typeof dest.getCCVsForEncodedMessage === 'function'
+      if (verifications && 'verificationPolicy' in verifications && destKeyedPolicy) {
+        try {
+          assertCoverage(verifications.verifications, verifications.verificationPolicy)
+          logger.info('managed sources covered the required CCV set; not contacting verifiers')
+        } catch {
+          logger.info(
+            'managed sources did not cover the required CCV set; falling back to --verifier',
+          )
+          verifications = undefined
+        }
+      } else if (verifications) {
+        // pre-v2 shape carries no policy to check; keep it
+        logger.debug('non-v2 verifications; --verifier not applicable')
+      }
+    }
+
+    if (verifications === undefined && (verifierEntries.length > 0 || ccvDataEntries.length > 0)) {
+      // Direct verifier fetch: read attestations from the verifiers themselves rather than the
+      // indexer. This is the only path available for a CCV that no indexer has onboarded.
+      verifications = await collectDirectVerifications({
+        dest,
+        offRamp,
+        encodedMessage: (request.message as { encodedMessage?: unknown }).encodedMessage,
+        messageId: request.message.messageId,
+        verifierEntries,
+        ccvDataEntries,
+        logger,
+      })
+    } else if (verifications === undefined) {
+      verifications = await dest.getVerifications({
+        ...argv,
+        indexer,
+        offRamp,
+        request,
+      })
+    }
 
     let estimated: number | undefined
     if (argv.estimateGasLimit != null) {
@@ -301,8 +366,55 @@ async function manualExec(
       }
     }
 
+    // Refuse to broadcast `execute(msg, [], [])`: with no attestation for any required CCV the
+    // OffRamp reverts RequiredCCVMissing, costing gas and moving the message to FAILURE. Placed
+    // after the estimate/`--only-estimate` block so a read-only diagnostic is never blocked by a
+    // broadcast-safety check.
+    if (
+      'verificationPolicy' in verifications &&
+      verifications.verificationPolicy.requiredCCVs.length > 0 &&
+      verifications.verifications.length === 0
+    ) {
+      throw new Error(
+        `no attestation was found for any of the ${verifications.verificationPolicy.requiredCCVs.length} ` +
+          `required CCV(s); OffRamp.execute would revert RequiredCCVMissing. Supply them with ` +
+          `--verifier <ccv>=<scheme>://<host>[:port] or --ccv-data <ccv>=<0x-hex>`,
+      )
+    }
+
     const input = await source.getExecutionInput({ ...argv, request, verifications })
     inputs = { input, offRamp }
+  } else if (verifierEntries.length > 0 || ccvDataEntries.length > 0) {
+    // No source chain, but attestations were supplied or can be fetched. A v2 execution needs only
+    // the encoded message and the OffRamp, both of which the CCIP API returns, so the source chain
+    // is not required: `getExecutionInput` on v2 merges the verifications and reads no chain.
+    if (!apiClient) {
+      throw new CCIPArgumentInvalidError(
+        'rpcs',
+        'reading attestations without a source-chain RPC needs the CCIP API; pass --rpcs for the ' +
+          'source chain, or drop --no-api',
+      )
+    }
+    const messageId = request.message.messageId
+    const execInput = await apiClient.getExecutionInput(messageId)
+    offRamp ??= execInput.offRamp
+    if (!('encodedMessage' in execInput)) {
+      throw new CCIPArgumentInvalidError(
+        'txHashOrId',
+        'this message predates CCIP v2.0; manual execution needs a source-chain RPC to build its ' +
+          'merkle proof. Pass --rpcs for the source chain.',
+      )
+    }
+    const collected = await collectDirectVerifications({
+      dest,
+      offRamp,
+      encodedMessage: execInput.encodedMessage,
+      messageId,
+      verifierEntries,
+      ccvDataEntries,
+      logger,
+    })
+    inputs = { input: { encodedMessage: execInput.encodedMessage, ...collected }, offRamp }
   }
 
   const [walletAddr, wallet] = await loadChainWallet(dest, argv, logger)
--- a/ccip-cli/src/index.ts
+++ b/ccip-cli/src/index.ts
@@ -113,11 +113,55 @@ const globalOpts = {
     describe:
       'Path to Canton config JSON file (party, ccipParty, jwt, edsUrl, transferInstructionUrl, etc.)',
   },
+  verifier: {
+    type: 'array',
+    string: true,
+    alias: 'verifier-endpoint',
+    describe:
+      'Fetch CCV attestations from a verifier when the CCIP API and indexer cannot cover the ' +
+      'required CCV set (e.g. a CCV no indexer has onboarded): ' +
+      '[<ccv-address>=]<scheme>://<host>[:port], repeatable. Schemes: grpc (aggregator over TLS), ' +
+      'grpc+plaintext (aggregator, no TLS). Repeat an address to give it failover endpoints, tried ' +
+      'in order. Omit the address to apply one endpoint to every required CCV.',
+    coerce: (entries: unknown[]): string[] =>
+      entries.flatMap((e) => {
+        if (typeof e !== 'string') {
+          throw new Error(`--verifier expects strings, got ${typeof e}`)
+        }
+        return e.split(',').map((s) => s.trim())
+      }),
+  },
+  'ccv-data': {
+    type: 'array',
+    string: true,
+    describe:
+      'Supply a CCV attestation directly as <ccv-address>=<0x-hex>, repeatable. The bottom of the ' +
+      'source ladder: use it when the CCIP API, the indexer and the verifier endpoint are all ' +
+      'unavailable, or to execute bytes obtained out of band. Validity is decided onchain by the ' +
+      "CCV's verifyMessage, so wrong bytes can only waste gas.",
+    coerce: (entries: unknown[]): string[] =>
+      entries.flatMap((e) => {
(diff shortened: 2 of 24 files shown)


ccip-cli/src/commands/manual-exec.ts
```diff
@@ -32,11 +33,16 @@ import {
   isSupportedTxHash,
   jsonStringify,
 } from '@chainlink/ccip-sdk/src/index.ts'
+import {
+  assertCoverage,
+  collectDirectVerifications,
+} from '@chainlink/ccip-sdk/src/verifiers/index.ts'
```
andrevmatos: In CLI, we shouldn't import from sdk other than src/index.ts; gotta export those functions

ccip-cli/src/commands/manual-exec.ts
```diff
@@ -217,16 +227,79 @@ async function manualExec(
   if (argv.estimateGasLimit != null && !source)
     source = await getChain(request.lane.sourceChainSelector)
 
+  // The verification block below lives under `if (source)`, and a bare messageId positional never
+  // sets `source`. Without this, --verifier/--ccv-data were silently ignored and execution fell
+  // through to `dest.execute({ messageId })` with NO verifications, broadcasting an unverified
+  // execute that reverts RequiredCCVMissing and costs gas. Resolve the source chain so the
+  // verifier fetch and the coverage assert actually run.
+  const verifierEntries = (argv as { verifier?: readonly string[] }).verifier ?? []
+  const ccvDataEntries = (argv as { ccvData?: readonly string[] }).ccvData ?? []
```
andrevmatos: I think we can avoid these casts; there should be an easy way to have them come typed from `argv`

ccip-cli/src/index.ts
```diff
@@ -113,11 +113,55 @@ const globalOpts = {
...
+      'in order. Omit the address to apply one endpoint to every required CCV.',
+    coerce: (entries: unknown[]): string[] =>
+      entries.flatMap((e) => {
+        if (typeof e !== 'string') {
+          throw new Error(`--verifier expects strings, got ${typeof e}`)
+        }
+        return e.split(',').map((s) => s.trim())
+      }),
+  },
+  'ccv-data': {
```
andrevmatos: This is not a global option; it's specific to `manual-exec`, only it cares about ccvData/verifications; move it there, pls

ccip-cli/src/commands/manual-exec.ts
```diff
@@ -37,6 +37,7 @@ import type { Argv } from 'yargs'
 
 import type { GlobalOpts } from '../index.ts'
 import { fetchChainsFromRpcs, loadChainWallet, resolveIndexer } from '../providers/index.ts'
+import { ccvDataOption, verifiersOption } from '../verifiers/options.ts'
```
andrevmatos: > In CLI, we shouldn't import from sdk other than `src/index.ts`; gotta export those functions

Addressed in 560bce52. Apart from `VerifierTransport` and `grpcWebTransport`, the CLI now needs nothing verifier-specific from the SDK. Both are exported from `src/index.ts`, along with `readVerifier`, `VERIFIER_METHOD`, `VerificationPolicy` and `VerificationSourcesOpts`. That made the `@chainlink/ccip-sdk/verifiers` subpath export, its typedoc entry point and the extra `patch-dist` sed rule unnecessary, so they're removed. `manual-exec.ts` is back to `main` plus these option imports.

ccip-cli/src/commands/manual-exec.ts
```diff
@@ -68,6 +69,8 @@ export const builder = (yargs: Argv) =>
     })
     .check((argv) => isSupportedTxHash(argv.txHashOrId ?? argv['tx-hash-or-id']))
     .options({
+      ...verifiersOption,
+      ...ccvDataOption,
```
andrevmatos: > I think we can avoid these casts; there should be an easy way to have them come typed from `argv`

Addressed in 560bce52. The options now live in the command builder, so yargs infers `argv.verifiers: string[] | undefined` and `argv.ccvData: Record<string, string> | undefined`. The `--ccv-data` coerce parses `<ccv>=<0x-hex>` into that record and rejects malformed entries up front. The names match the SDK options, so the existing `{ ...argv }` spreads into `dest.getVerifications()` and `dest.execute()` hand them over with no casts and no extra wiring.

That also removes every command-level branch: resolving the source chain, the managed-coverage check plus `destKeyedPolicy`, `collectDirectVerifications`, the no-source API branch, and the empty-set refusal. The SDK does all of that now.

ccip-cli/src/verifiers/options.ts
```diff
@@ -0,0 +1,66 @@
...
+      throw new CCIPArgumentInvalidError(flag, `expects strings, got ${typeof value}`)
+    }
+    return value
+      .split(',')
+      .map((s) => s.trim())
+      .filter(Boolean)
+  })
+
+/** `--verifier`: extra verifier endpoints to read CCV attestations from. */
+export const verifiersOption = {
```
andrevmatos: > This is not a global option; it's specific to `manual-exec`, only it cares about ccvData/verifications; move it there, pls

Addressed in 560bce52: both options are removed from the global options.
- `--verifier` (`verifiersOption`) is shared by `manual-exec` and `show`. `show` also reads verifications, and with `--wait` it now waits until the policy is covered.
- `--ccv-data` (`ccvDataOption`, below) is on `manual-exec` only.

I left the `--indexer` coerce and the CI libusb steps as they were. They're unrelated to this feature and would be better as a separate PR.

ccip-sdk/src/commits.ts
```diff
@@ -78,44 +150,131 @@ export async function fetchVerifications(
...
+    policy?.optionalCCVs.filter((c) => collected.has(key(c))).length ?? 0
+  const covered = () =>
+    policy
+      ? !missingRequired().length && optionalCovered() >= policy.optionalThreshold
+      : collected.size > 0
+  // results outside the policy are kept, as sources serve them: a family's policy read may not
+  // resolve every CCV the OffRamp enforces (e.g. Solana's token pool CCVs)
+  const selected = () => [...collected.values()]
+
+  collect(known)
```
andrevmatos: **Core of the refactor.** Verifiers are no longer a parallel pipeline. They are the last source of the existing `fetchVerifications`, which now takes the destination `policy`:
1. `known` (`--ccv-data`) is seeded first and wins over fetched results.
2. The API and indexers are raced as before. A *partial* answer is merged per dest CCV instead of winning the race (L231), and the race settles only once the policy is covered.
3. If the policy is still uncovered, the `verifiers` are tried **in order**, one at a time, stopping as soon as it is covered. The order is the failover order (primary, then backup), and managed infrastructure stays the default path.
4. If it's still uncovered after that, it throws `CCIPMessageNotVerifiedYetError`. That error is transient, so `watch`/`--wait` polling keeps working. Its context carries `missingCCVs`, `optionalCovered` and `verifierFailures`, and the CLI prints it, replacing `formatCoverageFailure`/`assertCoverage`.

Without `policy`, external callers keep the old first-source-wins behavior. Results outside the policy are kept rather than dropped, because Solana's policy read doesn't resolve pool CCVs yet (see the TODO in `solana/index.ts`).

**Behavior change:** a v2 `getVerifications` now throws instead of returning a set that can't cover the policy (that set would revert `RequiredCCVMissing` onchain). As a result, `show` without `--wait` logs "Verifications unavailable: …" instead of printing a partial list.

ccip-sdk/src/verifiers/transport.ts
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @example Delegate to the default for URLs a custom transport doesn't handle
+ * ```ts
+ * const web = grpcWebTransport()
+ * const transport: VerifierTransport = (call) =>
+ *   call.url.startsWith('grpc://') ? myGrpcClient.unary(call) : web(call)
+ * const chain = await EVMChain.fromUrl(rpc, { verifierTransport: transport })
+ * ```
+ */
+export type VerifierTransport = (call: VerifierCall) => Promise<Uint8Array>
```
andrevmatos: **Minimal port**, as discussed: a single function, `(call: { url, method, body, signal? }) => Promise<Uint8Array>`. The SDK owns the schema and decides when to call; a transport only frames bytes and does the I/O. The URL is opaque to the SDK, and its scheme is the transport's to interpret. This replaces the SDK-side `VerifierEndpoint` model (`type`/`target`/`tls`/`raw`) and the scheme table.

The default, `grpcWebTransport` (L101), assumes a grpc-web proxy in front of the verifier:
- it POSTs to `<url><method>` with grpc-web framing, through `fetchWithTimeout`;
- it handles the status from trailer frames and from trailers-only responses;
- it rejects non-`http(s)` URLs with a pointer to `ChainContext.verifierTransport`.

Trailers are parsed line by line rather than with a regex, to avoid another CodeQL ReDoS finding on server input.

ccip-sdk/src/verifiers/codec.ts
```diff
@@ -0,0 +1,160 @@
...
+    n = Math.floor(n / 0x80)
+  }
+  out.push(n)
+}
+
+/**
+ * Iterate the fields of one protobuf message: varints as numbers, length-delimited fields as byte
+ * views into `buf`. Fixed-width fields are skipped; no field this codec reads uses them.
+ */
+function* readFields(buf: Uint8Array): Generator<[field: number, value: number | Uint8Array]> {
```
andrevmatos: **A hand-written protobuf codec** (~150 lines) replaces `protobufjs`, the generated `descriptor.ts` and `gen-verifier-descriptor.mjs`. We read only 5 fields, so a wire reader is cheap, and the SDK takes no protobuf runtime dependency: no bundle cost and no `Root.fromJSON` at module load.
- Unknown fields and wire types are skipped, so additive schema changes are safe.
- Malformed input throws `MESSAGE_DECODE_FAILED`.
- Per-message `google.rpc.Status` errors are decoded too, and surface when a verifier returns no results.

`codec.test.ts` cross-checks the codec against protobufjs loading the real `proto/verifier.proto`, through `verifiers/__mocks__/verifier.ts`. protobufjs is now an SDK **devDependency** only.

ccip-sdk/src/verifiers/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    await transport({
+      url,
+      method: VERIFIER_METHOD,
+      body: encodeGetVerifierResultsRequest(messageId),
+      signal,
+    }),
+  )
+  const verifications: VerifierResult[] = []
+  for (const r of results) {
+    if (!r.ccvData?.length || !r.destAddress?.length) continue
```
andrevmatos: **Bug fix.** Results are now attributed only by `metadata.verifier_dest_address`, rendered in the dest family's format through `Chain.getAddress`. The old code had two problems:
- In `direct.ts`, `chosen = mine.length > 0 ? mine : read.results` gave B the blob of A. An endpoint that served only A's attestation, when queried for B, was reported as `status: 'ok'` with verifications `[A, A]`. I reproduced this before the refactor; the "never counts one CCV's attestation for another" test in `commits.test.ts` now covers it.
- `schema.ts` fell back to `message_ccv_addresses[0]`, which is a **source-side** address. `execute` passes `destAddress` as `ccvs[]` (`evm/index.ts` ~L1843), so that fallback would revert.

The 20-byte-only address check is gone as well, so Solana destinations work.

ccip-sdk/src/chain.ts
```diff
@@ -1727,6 +1768,29 @@ export abstract class Chain<F extends ChainFamily = ChainFamily> {
       opts_ = opts
     } else if (!this.apiClient) {
       throw new CCIPApiClientNotAvailableError()
+    } else if (
+      'messageId' in opts &&
```
andrevmatos: **The no-source path moved into the SDK**, replacing the CLI's `else if (verifierEntries.length || ccvDataEntries.length)` branch. `execute({ messageId, verifiers?, ccvData? })` reads the encoded message and the OffRamp from the API (`getEncodedMessage`), then runs this destination's own `getVerifications`: the policy read plus every source. With no source RPC, the CLI already calls `dest.execute({ ...argv, messageId })`, so `--verifier` works there without CLI code. When no extra sources are given, the API-only path is unchanged.

ccip-sdk/src/chain.ts
```diff
@@ -1878,10 +1953,67 @@ export abstract class Chain<F extends ChainFamily = ChainFamily> {
...
+  /**
+   * The MessageV1Codec-encoded form of a CCIP v2.0 request's message: taken from the request when
+   * it carries one (decoded from source logs), else fetched from the CCIP API (whose
+   * `getMessageById` requests don't).
+   *
+   * @param request - The v2.0 request
+   * @returns The 0x-prefixed encoded message
+   * @throws {@link CCIPApiClientNotAvailableError} if it must be fetched and the API is disabled
+   */
+  protected async resolveEncodedMessage(
```
andrevmatos: Requests from `getMessageById` carry no `encodedMessage`. On `main`, that breaks v2 `getVerifications` and `getExecutionInput` whenever a message ID comes with a source RPC, e.g. `manual-exec <id> --estimate-gas-limit` or `show <id>`. This PR's `request.message.encodedMessage` had the same problem.

This helper falls back to the new `CCIPAPIClient.getEncodedMessage`, which reads `/execution-inputs` without requiring the verifications to be complete, and shares that memoized fetch with `getExecutionInput`.

ccip-sdk/src/chain.ts
```diff
@@ -1878,10 +1953,67 @@ export abstract class Chain<F extends ChainFamily = ChainFamily> {
...
+   * `opts.ccvData`, then the CCIP API raced against the indexers, then `opts.verifiers` (read via
+   * {@link ChainContext.verifierTransport}). For families' `getVerifications` implementations.
+   *
+   * @param messageId - The message id
+   * @param policy - This destination's CCV policy for the message
+   * @param opts - The `getVerifications` options
+   * @returns The results covering `policy`
+   * @throws {@link CCIPMessageNotVerifiedYetError} if the sources can't cover `policy`
+   */
+  protected fetchCCVResults(
```
andrevmatos: Shared by the EVM and Solana v2 `getVerifications`; it replaces their duplicated `fetchVerifications` + `watch` blocks.

The destination family's `getAddress` canonicalizes every address involved (the policy, the API and indexer results, the `ccvData` keys and raw verifier bytes), so they match whatever encoding each source uses.

`getCCVsForEncodedMessage` is removed. It duplicated the policy read already in `EVMChain.getVerifications`, and it was the reason for the CLI's `destKeyedPolicy` special-casing of Solana and Canton. Canton keeps its own indexer path, so `verifiers` don't apply there.

ccip-cli/src/verifiers/grpc-transport.ts
```diff
@@ -0,0 +1,75 @@
...
+    const { protocol, host } = new URL(call.url)
+    const tls = GRPC_SCHEMES[protocol]
+    if (tls === undefined) {
+      if (protocol === 'http:' || protocol === 'https:') return web(call)
+      throw new CCIPArgumentInvalidError(
+        'verifier',
+        `unsupported scheme "${protocol}//": use https:// or http:// (grpc-web), grpc:// or grpcs:// (gRPC over TLS), or grpc+plaintext://`,
+      )
+    }
+    const { Client, credentials } = await (grpc$ ??= import('@grpc/grpc-js'))
```
andrevmatos: The CLI transport:
- `grpc://` and `grpcs://` (TLS) and `grpc+plaintext://` go through grpc-js with identity (de)serializers.
- `http(s)://` is delegated to the SDK's `grpcWebTransport`, so one `--verifier` list can mix both kinds.
- `@grpc/grpc-js` is imported lazily, on the first native call. Before, the static import loaded it for **every** CLI command, because `commandDir` loads all command modules. I checked this with a resolve hook on the built CLI.
- The abort signal cancels the call.
- grpc-js's empty trailing "Resolution note:" is stripped here, where it belongs, instead of in the SDK.

It is set once, in `getCtx`, so every chain gets it. `grpc-transport.integration.test.ts` covers it against a real local grpc-js server: a read, failover past a dead endpoint, abort, and `http(s)` delegation.

andrevmatos (approved):

tt-cll (approved):

timweri (approved):

PR #392, opened 2026-08-31
fix(canton): resolve CCIPReceiver package + CCV raw address for manual-exec

SyedAsadKazmi:
Fixes manual-exec on Canton destinations failing with "no active contract found" and "caller is not an allowed CCV" errors.

### CCIPReceiver template ID hardcoded

`createReceiverForFinality`/`updateReceiverRequiredCCVs` used `#ccip-receiver` instead of the `packages.ccipReceiver` override from canton-config (e.g. `ccip-receiver-v2`). The created receiver was invisible to the ACS filter, so a new one was created every run with a different contract ID. Added `ccipReceiver` to `ccipPackages` and use it for the template ID.

### CCV address format mismatch

The attestation CCV from the indexer/canton-config is a keccak256 InstanceAddress hash, but `CCIPReceiver.requiredCCVs` expects the raw `RawInstanceAddress.unpack` string. The hash was stored directly, so the Daml `Execute` choice never matched the CCV caller. Resolve the raw form via EDS (`fetchCcvExecuteDisclosure.rawInstanceAddress`) before creating/updating the receiver.

### DIFF
--- a/ccip-sdk/src/canton/index.ts
+++ b/ccip-sdk/src/canton/index.ts
@@ -236,9 +236,9 @@ export class CantonChain extends Chain<typeof ChainFamily.Canton> {
     this.senderInstanceId =
       ctx?.cantonConfig?.senderInstanceId?.trim() || DEFAULT_CANTON_SENDER_INSTANCE_ID
     this.ccipPackages = {
-      perPartyRouter: ctx?.cantonConfig?.packages?.perPartyRouter ?? 'ccip-perpartyrouter',
-      ccipSender: ctx?.cantonConfig?.packages?.ccipSender ?? 'ccip-sender',
-      ccipReceiver: ctx?.cantonConfig?.packages?.ccipReceiver ?? 'ccip-receiver',
+      perPartyRouter: ctx?.cantonConfig?.packages?.perPartyRouter ?? 'ccip-runtime-v2',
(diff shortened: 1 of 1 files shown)


ccip-sdk/src/canton/index.ts
```diff
@@ -234,6 +238,7 @@ export class CantonChain extends Chain<typeof ChainFamily.Canton> {
     this.ccipPackages = {
       perPartyRouter: ctx?.cantonConfig?.packages?.perPartyRouter ?? 'ccip-perpartyrouter',
       ccipSender: ctx?.cantonConfig?.packages?.ccipSender ?? 'ccip-sender',
+      ccipReceiver: ctx?.cantonConfig?.packages?.ccipReceiver ?? 'ccip-receiver',
```
friedemannf: Lets default to the current version:

```suggestion
      perPartyRouter: ctx?.cantonConfig?.packages?.perPartyRouter ?? 'ccip-runtime-v2',
      ccipSender: ctx?.cantonConfig?.packages?.ccipSender ?? 'ccip-sender-v2',
      ccipReceiver: ctx?.cantonConfig?.packages?.ccipReceiver ?? 'ccip-receiver-v2',
```

friedemannf (approved):

andrevmatos (approved):

PR #393, opened 2026-08-31
feat(cct-sdk): Add set chain rate limiter configs op

apedrob:
## What

- [DAPP-11281](https://smartcontract-it.atlassian.net/browse/DAPP-11281)
- Add `setChainRateLimiterConfigs` and `generateUnsignedSetChainRateLimiterConfigs` to `EVMTokenManager`. They set the inbound and outbound rate limits of lanes that are already configured, in one transaction
- The encoding is version-specific, and each version keeps the one-op-one-transaction rule:
  - v1.5.0 only has the singular `setChainRateLimiterConfig`, so it takes one lane per call. Passing more than one lane is rejected rather than split into several transactions
  - v1.5.1 and v1.6.1 use the batch `setChainRateLimiterConfigs`
  - v2.0.0 uses the reshaped `setRateLimitConfig`, which adds the `fastFinality` flag (v2.0.0 only)
- The write is gated on the pool `owner` or its `rateLimitAdmin`. When `sender` is given, `buildUnsigned` checks it against both roles, so the offline and multisig path fails at build time instead of reverting after signing

## Why

- Rate limits are the last lane setting a pool owner needs to change after lanes are configured. Without this op the SDK can enable a route but not re-limit it
- This is the only owner-gated write that also accepts a delegated `rateLimitAdmin`, so the check is a disjunction rather than the owner-only `assertPoolOwner`

## Testing

- Unit tests cover the three rate limit calldata

[DAPP-11281]: https://smartcontract-it.atlassian.net/browse/DAPP-11281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/evm/token-pool/operations/set-chain-rate-limiter-configs.ts
```diff
@@ -0,0 +1,352 @@
...
+function parseUpdates(
+  operation: string,
+  updates: unknown,
+  allowFastFinality: boolean,
+  version: TokenPoolVersion | null,
+): ParsedChainRateLimitUpdate[] {
+  if (!Array.isArray(updates) || updates.length === 0)
+    throw new CCTParamsInvalidError(operation, 'updates', 'must be a non-empty array')
+  // `.map` below skips holes, so reject a sparse array before it can smuggle one past validation
+  assertDenseArray(operation, 'updates', updates)
```
mervin-link, line 109: can we reuse `validateArray` instead of creating a new one?

```suggestion
  validateArray(operation, 'updates', updates, 1)
```

mervin-link (approved):
## Approved ✅

Tested the set chain rate-limit configurations:
- [v1.5.0](https://sepolia.etherscan.io/tx/0x994e0e710942e85d1d374eea2ddce1cd14edffa25226e5d3141621e71a7686df)
- [v1.6.1](https://sepolia.etherscan.io/tx/0x5cc2b0fd53ac7b8dc01871faf6103f3394217b1b0862dc702afc0d279ee2bfcb)
- [v2.0.0](https://sepolia.etherscan.io/tx/0x64d5c45f2641dd9b29ad4fda97021407ff465131acd4afb9186ff884047bf2e6)
- [with fast finality](https://sepolia.etherscan.io/tx/0xde50650d55eae5a809a4bbe6c601dd04fa6a5538152be3fa983a04a4c0172616)

 ### Nit

 Could we reuse `validateArray` instead of introducing `assertDenseArray`?

ccip-sdk/src/cct/evm/token-pool/operations/set-chain-rate-limiter-configs.ts
apedrob: Added into https://github.com/smartcontractkit/ccip-tools-ts/pull/397

PR #396, opened 2026-09-01
feat(cct-sdk): Add apply allowlist updates evm op

apedrob:
## What
- [DAPP-11283](https://smartcontract-it.atlassian.net/browse/DAPP-11283)
- Add `applyAllowlistUpdates` / `generateUnsignedApplyAllowlistUpdates`: replaces a token pool's sender allowlist in one call
- Removes apply before adds; an address in both arrays is rejected rather than silently allowlisted
## Why
- The allowlist is the last sender-side control a pool owner needs after lanes are configured; without it the SDK can enable a route but not restrict who uses it
## Testing
- Unit tests cover the calldata, removes-before-adds byte parity, and the empty/duplicate/overlap rejections

## Notes

- v1.5.0 to v1.6.1 only; v2.0.0 has no allowlist, so it reports unsupported

[DAPP-11283]: https://smartcontract-it.atlassian.net/browse/DAPP-11283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ccip-sdk/src/cct/evm/index.ts
```diff
@@ -1178,6 +1183,68 @@ export class EVMTokenManager extends TokenManager<typeof ChainFamily.EVM> {
...
+   * ```typescript
+   * const { hash } = await cct.applyAllowlistUpdates({
+   *   poolAddress: '0xPool...',
+   *   removes: ['0xRevoked...'],
+   *   adds: ['0xNewSender...'],
+   *   wallet,
+   * })
+   * ```
+   */
+  applyAllowlistUpdates(
```
mervin-link, line 1265: Nit: Add here that removals run before additions and overlapping addresses are rejected, matching the unsigned method’s TSDoc.

ccip-sdk/src/cct/evm/token-pool/operations/apply-allowlist-updates.ts
```diff
@@ -0,0 +1,200 @@
...
+   * should surface as {@link CCTOperationUnsupportedError} rather than burning an RPC on a pool
+   * this op can never target. The owner check is skipped entirely when `sender` is omitted —
+   * there is nothing to compare against, and `generateUnsignedApplyAllowlistUpdates` is expected
+   * to be usable before the eventual signer is known. {@link execute} always supplies one.
+   * @throws {@link CCTOperationUnsupportedError} if the pool is v2.0.0
+   * @throws {@link CCTContractTypeInvalidError} if the address is not a supported pool type
+   * @throws {@link CCTContractVersionUnsupportedError} if the pool reports an unknown version
+   * @throws {@link CCTParamsInvalidError} if `sender` is given and is not the pool owner
+   */
+  protected async buildUnsigned(
```
mervin-link, line 204: `allowlistEnabled` must be true for this operation; otherwise the contract reverts with `AllowListNotEnabled`.

We can either document this requirement or add a preflight check via `getAllowListEnabled()`.

ccip-sdk/src/cct/evm/token-pool/operations/apply-allowlist-updates.ts
```diff
@@ -0,0 +1,200 @@
...
+   * - **an address in BOTH `adds` and `removes`** — rejected: removes apply first, so the address
+   *   would end up *allowlisted*, and no caller can reasonably have meant both.
+   *
+   * Comparisons are on checksummed form, so the same address in two different casings still
+   * counts as a duplicate / an overlap.
+   * @throws {@link CCTParamsInvalidError} if `poolAddress` is invalid, either array is not an
+   * array or is sparse, both are empty, an entry is not a valid address (reported as `adds[i]` /
+   * `removes[i]`), an array holds duplicates, or an address appears in both arrays
+   */
+  protected override parse(params: ApplyAllowlistUpdatesParams): ParsedApplyAllowlistUpdatesParams {
```
mervin-link, line 148: There are other no-op updates we could catch before submission:

- Removing addresses that are not currently allowlisted
- Adding or removing the zero address

A preflight against the current allowlist would provide clearer errors and avoid submitting transactions that cannot change pool state.

mervin-link (approved):
## Approved ✅

Tested against all three supported contract versions; works as expected.

- [v1.5.0](https://sepolia.etherscan.io/tx/0x4339a5b5a00c25da23597d1ff8529c2c421938dbbb687e44653dd39d4eb8c6e0) ✅ 
- [v1.6.1](https://sepolia.etherscan.io/tx/0x8e19b31c1d79c25dc25e8aa45fd2fe7459b16ee8b2bb8f2ea0e32e752d6f7776) ✅ 
- v2.0 🔴 Rejected

There are a few small nits, but nothing blocking.

ccip-sdk/src/cct/evm/token-pool/operations/apply-allowlist-updates.ts
apedrob: good catch, added preflight check

ccip-sdk/src/cct/evm/token-pool/operations/apply-allowlist-updates.ts
apedrob: it now reads getAllowlist and getAllowlistEnabled and op rejects a removes entry that isnt allowlisted and adds entry that already is

PR #413, opened 2026-09-08
feat(cct-sdk): Add lock-release liquidity + rebalancer evm ops

apedrob:
## What

- Add the EVM LockRelease liquidity surface to `EVMTokenManager`:
  - `provideLiquidity`, `withdrawLiquidity`, `transferLiquidity`
  - `setRebalancer`, `getRebalancer`
  - `approveToken`: ERC-20 `approve`, without it the liquidity flow cannot be completed through the SDK alone; also it matches the solana op of the same name
- Pool ops target LockRelease `1.5.0` to `1.6.1`

## Why

- Closes the liquidity and rebalancer flows needed for setting up cross-chain pools on EVM

## Testing

- Unit tests 
- Verified with [`LockReleaseTokenPool 1.6.1`](https://sepolia.basescan.org/address/0xf60217d486f11EfD1d31d126441699710e054F4d) + [`Token`](https://sepolia.basescan.org/token/0xa7f29d90f2ad98d6c2972fc72c4b613270c3ef4d):
  - [`approveToken`](https://sepolia.basescan.org/tx/0x43e36eb53fe1b560b69c173a349657557c2bb4a76b99ef41e424e40435467015)
  - [`provideLiquidity`](https://sepolia.basescan.org/tx/0x5d8b07766b5ad45dea0d53fb257f3854ea8b92947e26c547818ce31b1f3f366e)
  - [`withdrawLiquidity`](https://sepolia.basescan.org/tx/0x67c4ccf762001ff95d37234029a51ff7be9d96a014dd2c3fa042387a95c02c72)
  - [`setRebalancer` for new pool](https://sepolia.basescan.org/tx/0x6b230dee62571a4306072735f4076196c1adb8be586f4076bfa9f28618b99546)
  - [`transferLiquidity` between pools](https://sepolia.basescan.org/tx/0xdb15f73780a70494f92ecc1b95c572e4f10ebd142a423b4605b356672ad8fd01)
  
## Notes
 I suggest reviewing commit by commit to make things easier. I kept everything in the same PR since these ops are tightly coupled and needed for the full flow to work

PR #402, opened 2026-09-04
feat(cct-sdk): Add mint/burn role management ops

apedrob:
## What
- Add five mint/burn role ops for BurnMintERC677 tokens (v1.5 / v1.6), each as
`generateUnsigned<Op>` and `<op>`, plus the `isMinter`/`isBurner` and `owner` reads they gate on

## Why
- Burn/mint pool cannot bridge until the token owner grants it both roles

## Testing
- Unit tests

PR #416, opened 2026-09-09
fix(cct-sdk): Prevent partial Solana CCT commits

mervin-link:
## What

- For more info: [DAPP-11383](https://smartcontract-it.atlassian.net/browse/DAPP-11383)
- Add CCT-local Solana submission that splits only locally oversized transaction slices
- Rethrow simulation/program errors without truncating instructions
- Preserve `setTokenAuthority` atomicity and validate selected mint/freeze authorities before submission
- Report confirmed split transaction hashes in `CCTTxFailedError.context.committedHashes`

## Why

- Prevent a failed multi-instruction CCT operation from committing an earlier instruction while returning an error

## Notes

- CCT uses a local `simulateAndSendTxs` helper because shared Solana version treats every simulation failure as a size failure and splits instructions. The local path splits only after a deterministic size check, preserving CCT operation semantics without changing shared SDK behavior.
- Operations can pass requireSingleTransaction to reject oversize rather than split when they require atomic execution.

[DAPP-11383]: https://smartcontract-it.atlassian.net/browse/DAPP-11383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ccip-sdk/src/cct/errors.ts
+++ b/ccip-sdk/src/cct/errors.ts
@@ -45,7 +45,9 @@ export class CCTParamsInvalidError extends CCIPError {
  * Thrown when a CCT write fails before broadcast, the transaction reverts after mining,
  * or it mines without the expected effect (e.g. a deployment that produced no contract
  * address). Pre-broadcast failures (signing/RPC) may set `isTransient: true` for network
- * errors; reverts and post-mining anomalies are permanent and include `context.txHash`.
+ * errors; reverts and post-mining anomalies are permanent and include `context.txHash`. For a
+ * split Solana operation, `context.committedHashes` contains confirmed prior transaction hashes;
+ * state is partially applied and those transactions are permanent.
  *
  * @example
  * ```typescript
@@ -62,17 +64,23 @@ export class CCTTxFailedError extends CCIPError {
   override readonly name = 'CCTTxFailedError'
   /** Creates a tx-failed error. */
   constructor(operation: string, reason: string, options?: CCIPErrorOptions) {
-    super(CCIPErrorCode.CCT_TX_FAILED, `${operation} failed: ${reason}`, {
-      ...options,
-      isTransient: options?.isTransient ?? false,
-      context: { ...options?.context, operation, reason },
-    })
+    super(
+      CCIPErrorCode.CCT_TX_FAILED,
+      `${partialApplicationPrefix(options?.context)}${operation} failed: ${reason}`,
+      {
+        ...options,
+        isTransient: options?.isTransient ?? false,
+        context: { ...options?.context, operation, reason },
+      },
+    )
   }
 }
 
 /**
  * Thrown when a transaction was broadcast but not confirmed within the timeout.
- * Transient — it may still mine; check `context.txHash` before resubmitting.
+ * Transient — it may still mine; check `context.txHash` before resubmitting. For a split Solana
+ * operation, `context.committedHashes` contains confirmed prior transaction hashes; state is
+ * partially applied and those transactions are permanent.
  *
  * @example
  * ```typescript
@@ -91,7 +99,7 @@ export class CCTTxNotConfirmedError extends CCIPError {
   constructor(operation: string, txHash: string, options?: CCIPErrorOptions) {
     super(
       CCIPErrorCode.CCT_TX_NOT_CONFIRMED,
-      `${operation} transaction not confirmed within timeout: ${txHash}`,
+      `${partialApplicationPrefix(options?.context)}${operation} transaction not confirmed within timeout: ${txHash}`,
       {
         ...options,
         isTransient: true,
@@ -100,6 +108,13 @@ export class CCTTxNotConfirmedError extends CCIPError {
       },
     )
   }
+}
+
+function partialApplicationPrefix(context?: Record<string, unknown>): string {
+  const committedHashes = context?.committedHashes
+  return Array.isArray(committedHashes) && committedHashes.length
+    ? `partially applied: ${committedHashes.length} transaction(s) confirmed; `
+    : ''
 }
 
 // Contract version dispatch
--- a/ccip-sdk/src/cct/solana/submit.test.ts
+++ b/ccip-sdk/src/cct/solana/submit.test.ts
@@ -2,6 +2,7 @@ import assert from 'node:assert/strict'
 import { describe, it } from 'node:test'
 
 import {
+  AddressLookupTableAccount,
   ComputeBudgetProgram,
   Keypair,
   SendTransactionError,
@@ -125,15 +126,30 @@ describe('Submit error mapping (cct/solana)', () => {
   it('reports confirmed slices when a later slice fails', async () => {
     const hash = 'confirmed-slice'
     let simulations = 0
+    const simulationLookupCounts: number[] = []
+    const lookupAddress = Keypair.generate().publicKey
+    const lookupTable = new AddressLookupTableAccount({
+      key: Keypair.generate().publicKey,
+      state: {
+        deactivationSlot: 0xffff_ffff_ffff_ffffn,
+        lastExtendedSlot: 0,
+        lastExtendedSlotStartIndex: 0,
+        authority: undefined,
+        addresses: [lookupAddress],
+      },
+    })
     const chain = {
       logger: { debug() {}, info() {}, warn() {}, error() {} },
       connection: {
-        simulateTransaction: async () => ({
-          value: {
-            err: ++simulations === 1 ? null : { InstructionError: [1, { Custom: 4 }] },
-            logs: [],
-          },
-        }),
+        simulateTransaction: async (tx: { message: { addressTableLookups: unknown[] } }) => {
+          simulationLookupCounts.push(tx.message.addressTableLookups.length)
+          return {
+            value: {
+              err: ++simulations === 1 ? null : { InstructionError: [1, { Custom: 4 }] },
+              logs: [],
+            },
+          }
+        },
         getLatestBlockhash: async () => ({
           blockhash: Keypair.generate().publicKey.toBase58(),
           lastValidBlockHeight: 1,
@@ -149,7 +165,7 @@ describe('Submit error mapping (cct/solana)', () => {
     const instruction = () =>
       new TransactionInstruction({
         programId: Keypair.generate().publicKey,
-        keys: [],
+        keys: [{ pubkey: lookupAddress, isSigner: false, isWritable: false }],
         data: Buffer.alloc(700),
       })
 
@@ -160,14 +176,17 @@ describe('Submit error mapping (cct/solana)', () => {
         {
           family: ChainFamily.Solana,
           instructions: [instruction(), instruction()],
+          lookupTables: [lookupTable],
           mainIndex: 0,
         },
         OP,
       ),
       (error: unknown) =>
         error instanceof CCTTxFailedError &&
+        error.message.startsWith('partially applied: 1 transaction(s) confirmed;') &&
         Array.isArray(error.context.committedHashes) &&
         error.context.committedHashes[0] === hash,
     )
+    assert.deepEqual(simulationLookupCounts, [1, 1])
   })
 })
--- a/ccip-sdk/src/cct/solana/submit.test.ts
+++ b/ccip-sdk/src/cct/solana/submit.test.ts
@@ -1,16 +1,26 @@
 import assert from 'node:assert/strict'
 import { describe, it } from 'node:test'
 
-import { SendTransactionError, TransactionExpiredTimeoutError } from '@solana/web3.js'
+import {
+  ComputeBudgetProgram,
+  Keypair,
+  SendTransactionError,
+  TransactionExpiredTimeoutError,
+  TransactionInstruction,
+} from '@solana/web3.js'
 
+import { ChainFamily } from '../../networks.ts'
+import type { SolanaChain } from '../../solana/index.ts'
 import { CCTTxFailedError, CCTTxNotConfirmedError } from '../errors.ts'
-import { createCCTSubmitError } from './submit.ts'
+import { createCCTSubmitError, submit } from './submit.ts'
 
 const OP = 'setPool'
 
 describe('Submit error mapping (cct/solana)', () => {
   it('maps post-broadcast confirmation errors with a signature to not-confirmed', () => {
-    const cause = Object.assign(new Error('transaction was not confirmed'), { signature: 'abc' })
+    const cause = Object.assign(new Error('transaction was not confirmed'), {
+      signature: 'abc',
+    })
     const err = createCCTSubmitError(OP, cause)
 
     assert.ok(err instanceof CCTTxNotConfirmedError)
@@ -38,7 +48,9 @@ describe('Submit error mapping (cct/solana)', () => {
   })
 
   it('maps signed on-chain failures to permanent tx failed', () => {
-    const cause = Object.assign(new Error('custom program error: 0x1'), { signature: 'jkl' })
+    const cause = Object.assign(new Error('custom program error: 0x1'), {
+      signature: 'jkl',
+    })
     const err = createCCTSubmitError(OP, cause)
 
     assert.ok(err instanceof CCTTxFailedError)
@@ -70,5 +82,92 @@ describe('Submit error mapping (cct/solana)', () => {
 
     assert.ok(err instanceof CCTTxFailedError)
     assert.equal(err.isTransient, false)
+  })
+
+  it('does not submit a slice when its simulation rejects', async () => {
+    let sends = 0
+    const chain = {
+      logger: { debug() {}, info() {}, warn() {}, error() {} },
+      connection: {
+        simulateTransaction: async () => ({
+          value: { err: { InstructionError: [2, { Custom: 4 }] }, logs: [] },
+        }),
+        sendTransaction: async () => {
+          sends++
+          return 'unused'
+        },
+      },
+    } as unknown as SolanaChain
+    const wallet = {
+      publicKey: Keypair.generate().publicKey,
+      signTransaction: async <T>(tx: T) => tx,
+    }
+
+    await assert.rejects(
+      submit(
+        chain,
+        wallet,
+        {
+          family: ChainFamily.Solana,
+          instructions: [
+            ComputeBudgetProgram.setComputeUnitLimit({ units: 1 }),
+            ComputeBudgetProgram.setComputeUnitLimit({ units: 2 }),
+          ],
+          mainIndex: 0,
+        },
+        OP,
+      ),
+      CCTTxFailedError,
+    )
+    assert.equal(sends, 0)
+  })
+
+  it('reports confirmed slices when a later slice fails', async () => {
+    const hash = 'confirmed-slice'
+    let simulations = 0
+    const chain = {
+      logger: { debug() {}, info() {}, warn() {}, error() {} },
+      connection: {
+        simulateTransaction: async () => ({
+          value: {
+            err: ++simulations === 1 ? null : { InstructionError: [1, { Custom: 4 }] },
+            logs: [],
+          },
+        }),
+        getLatestBlockhash: async () => ({
+          blockhash: Keypair.generate().publicKey.toBase58(),
+          lastValidBlockHeight: 1,
+        }),
+        sendTransaction: async () => hash,
+        confirmTransaction: async () => ({ value: { err: null } }),
+      },
+    } as unknown as SolanaChain
+    const wallet = {
+      publicKey: Keypair.generate().publicKey,
+      signTransaction: async <T>(tx: T) => tx,
+    }
+    const instruction = () =>
+      new TransactionInstruction({
+        programId: Keypair.generate().publicKey,
+        keys: [],
+        data: Buffer.alloc(700),
+      })
+
+    await assert.rejects(
+      submit(
+        chain,
+        wallet,
+        {
+          family: ChainFamily.Solana,
+          instructions: [instruction(), instruction()],
+          mainIndex: 0,
+        },
+        OP,
+      ),
+      (error: unknown) =>
+        error instanceof CCTTxFailedError &&
+        Array.isArray(error.context.committedHashes) &&
+        error.context.committedHashes[0] === hash,
+    )
   })
 })
--- a/ccip-sdk/src/cct/solana/submit.ts
+++ b/ccip-sdk/src/cct/solana/submit.ts
@@ -65,12 +65,13 @@ function getTransactionSlice(unsigned: UnsignedSolanaTx, start: number, end: num
   return {
     includesMain,
     instructions: unsigned.instructions.slice(start, end),
-    lookupTables: includesMain ? unsigned.lookupTables : undefined,
+    // Lookup tables may be needed by instructions in any slice.
+    lookupTables: unsigned.lookupTables,
   }
 }
 
 /**
- * Submits CCT instruction slices, shrinking only locally oversized slices.
+ * Submits CCT instruction slices, splitting only after local size-overflow detection.
  *
  * A simulation failure aborts its slice without submitting it. Confirmed earlier slices are
  * attached as `committedHashes` when a later slice fails. Set `requireSingleTransaction` for
--- a/ccip-sdk/src/cct/solana/submit.ts
+++ b/ccip-sdk/src/cct/solana/submit.ts
@@ -8,31 +8,177 @@
  */
 
 import {
+  ComputeBudgetProgram,
+  PACKET_DATA_SIZE,
+  PublicKey,
   TransactionExpiredBlockheightExceededError,
   TransactionExpiredNonceInvalidError,
   TransactionExpiredTimeoutError,
+  TransactionMessage,
+  VersionedTransaction,
 } from '@solana/web3.js'
 
 import { CCIPWalletInvalidError, shouldRetry } from '../../errors/index.ts'
 import type { SolanaChain } from '../../solana/index.ts'
-import { type UnsignedSolanaTx, isWallet } from '../../solana/types.ts'
-import { simulateAndSendTxs } from '../../solana/utils.ts'
+import { type UnsignedSolanaTx, type Wallet, isWallet } from '../../solana/types.ts'
+import { simulateTransaction } from '../../solana/utils.ts'
 import { CCTTxFailedError, CCTTxNotConfirmedError } from '../errors.ts'
 import type { TransactionResult } from '../operation.ts'
 
-/** Signs, simulates, sends, and confirms a Solana CCT transaction. */
+const MAX_COMPUTE_UNITS = 1_400_000
+
+function buildTransaction(
+  payer: PublicKey,
+  instructions: UnsignedSolanaTx['instructions'],
+  lookupTables: UnsignedSolanaTx['lookupTables'],
+  computeUnits?: number,
+  recentBlockhash = PublicKey.default.toBase58(),
+): VersionedTransaction {
+  return new VersionedTransaction(
+    new TransactionMessage({
+      payerKey: payer,
+      recentBlockhash,
+      instructions: [
+        ...(computeUnits === undefined
+          ? []
+          : [ComputeBudgetProgram.setComputeUnitLimit({ units: computeUnits })]),
+        ...instructions,
+      ],
+    }).compileToV0Message(lookupTables),
+  )
+}
+
+function exceedsTransactionSize(tx: VersionedTransaction): boolean {
+  try {
+    return tx.serialize().length > PACKET_DATA_SIZE
+  } catch (error) {
+    if (error instanceof Error && error.message.includes('encoding overruns Uint8Array')) {
+      return true
+    }
+    throw error
+  }
+}
+
+function getTransactionSlice(unsigned: UnsignedSolanaTx, start: number, end: number) {
+  const includesMain =
+    unsigned.mainIndex != null && start <= unsigned.mainIndex && unsigned.mainIndex < end
+  return {
+    includesMain,
+    instructions: unsigned.instructions.slice(start, end),
+    lookupTables: includesMain ? unsigned.lookupTables : undefined,
+  }
+}
+
+/**
+ * Submits CCT instruction slices, shrinking only locally oversized slices.
+ *
+ * A simulation failure aborts its slice without submitting it. Confirmed earlier slices are
+ * attached as `committedHashes` when a later slice fails. Set `requireSingleTransaction` for
+ * operations that must never be split.
+ */
+async function simulateAndSendCCTTxs(
+  chain: SolanaChain,
+  wallet: Wallet,
+  unsigned: UnsignedSolanaTx,
+  operation: string,
+  computeUnits?: number,
+  requireSingleTransaction = false,
+): Promise<string> {
+  let mainHash: string | undefined
+  const committedHashes: string[] = []
+  try {
+    for (let start = 0; start < unsigned.instructions.length;) {
+      let end = unsigned.instructions.length
+      let slice = getTransactionSlice(unsigned, start, end)
+
+      while (
+        exceedsTransactionSize(
+          buildTransaction(
+            wallet.publicKey,
+            slice.instructions,
+            slice.lookupTables,
+            MAX_COMPUTE_UNITS,
+          ),
+        )
+      ) {
+        if (requireSingleTransaction || --end === start) {
+          throw new CCTTxFailedError(
+            operation,
+            `transaction exceeds Solana's ${PACKET_DATA_SIZE}-byte limit`,
+          )
+        }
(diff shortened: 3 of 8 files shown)


aelmanaa:
## 🟢 Core Atomicity Preservation — Excellent Design

**Location:** `submit.ts` lines 1-157, `set-token-authority.ts` lines 174-202

**Summary:**
This PR delivers a well-designed solution to prevent partial Solana CCT commits. The new `simulateAndSendCCTTxs` function splits transactions only on deterministic local size checks (via `exceedsTransactionSize`), **not** on simulation or execution errors. SetTokenAuthority enforces atomicity by calling `submit(..., true)` with `requireSingleTransaction=true`, guaranteeing that either all selected roles update or none do.

**Why This Matters:**
- **Before:** Multi-role operations could split mid-transaction; if role #1 succeeded and role #2 failed on-chain, role #1 would remain committed with no rollback mechanism
- **After:** SetTokenAuthority's atomic guarantee means all roles update together or the entire operation fails pre-broadcast

**Key Implementation Details:**
1. `exceedsTransactionSize(instructions, blockheight)` deterministically checks whether instructions exceed PACKET_DATA_SIZE before any chain call
2. Split loop in `simulateAndSendCCTTxs` (lines ~108-145) only enters when `exceedsTransactionSize` returns true AND `requireSingleTransaction` is false
3. When a split fails post-broadcast, `committedHashes` array in `CCTTxFailedError.context` tracks which prior slices succeeded
4. SetTokenAuthority always calls `submit(..., true)`, rejecting oversized input before RPC

**Tests:** All 16 SetTokenAuthority tests passing; submit.ts has 96.94% coverage with split/committedHashes behavior validated.

✅ **Recommendation:** Approve. This is production-ready atomicity semantics.

aelmanaa:
## 🟡 Documentation Gap: Explain WHY SetTokenAuthority Forbids Splitting

**Location:** `set-token-authority.ts` lines 64-78 (class JSDoc)

**Current Text:**
```typescript
/**
 * Immediately sets mint authority, freeze authority, or both for an SPL Token mint; there is no
 * propose-and-accept step.
 *
 * @remarks
 * ⚠️ **IRREVERSIBLE:** Setting `newAuthority` to null permanently revokes the selected roles.
 * Once revoked, a revoked mint or freeze authority **cannot be recovered, transferred, or restored**.
 *
 * Once confirmed, the current authority loses the selected roles. All selected roles must have the
 * same current authority. Supply `multisigSigners` when that authority is an SPL Token multisig.
 * **Atomic:** All selected roles update in one transaction. If any selected update fails, none are
 * committed.
 */
```

**Issue:**
The JSDoc correctly states atomicity ("All selected roles update in one transaction") but omits *why* this matters. Readers unfamiliar with the split-on-size concept won't understand the risk that this design prevents.

**Suggested Revision:**
```typescript
/**
 * Immediately sets mint authority, freeze authority, or both for an SPL Token mint; there is no
 * propose-and-accept step.
 *
 * @remarks
 * ⚠️ **IRREVERSIBLE:** Setting `newAuthority` to null permanently revokes the selected roles.
 * Once revoked, a revoked mint or freeze authority **cannot be recovered, transferred, or restored**.
 *
 * Once confirmed, the current authority loses the selected roles. All selected roles must have the
 * same current authority. Supply `multisigSigners` when that authority is an SPL Token multisig.
 *
 * **Atomic Enforcement:** All selected roles update in one transaction. If any update fails, none
 * are committed. This prevents partial authority changes: if mint authority succeeds but freeze
 * authority fails on-chain, both must fail pre-submission. (Solana limits transactions to
 * PACKET_DATA_SIZE bytes; multi-instruction operations exceeding this limit can be split across
 * multiple transactions on-chain. For role updates, partial splits are unrecoverable, so this
 * operation forbids splitting and rejects oversized input before RPC.)
 */
```

**Why:**
New operators may not know that Solana splits oversized transactions, or that partial authority updates create unrecoverable state. This extended note teaches both constraints and rationale.

**Tests:** No change needed; tests already validate atomicity.

✅ **Recommendation:** Apply this revision to class JSDoc.

aelmanaa:
## 🟡 Error Context Clarity: Document committedHashes Semantics

**Location:** `submit.ts` lines ~130-150 (error handling in simulateAndSendCCTTxs)

**Current Behavior:**
When `simulateAndSendCCTTxs` splits instructions across multiple transactions and a later slice fails post-broadcast, it populates `CCTTxFailedError.context.committedHashes` with the hashes of successfully committed prior slices.

**Issue:**
The committedHashes semantic is not documented in:
1. `submit.ts` JSDoc on `simulateAndSendCCTTxs`
2. `CCTTxFailedError` error type definition
3. SetTokenAuthority JSDoc (which uses `requireSingleTransaction=true`, so never hits committedHashes in normal flow)

This leaves operators unclear about when committedHashes is populated and what recovery actions are safe.

**Suggested Fix:**

1. **In `submit.ts` JSDoc for simulateAndSendCCTTxs:**
```typescript
/**
 * Submits Solana transaction instructions, splitting if oversized.
 * 
 * @param requireSingleTransaction - If true, reject oversized transactions before RPC
 * instead of splitting. Ensures atomicity for all-or-nothing operations.
 * 
 * @throws CCTTxFailedError - If any transaction fails. If this operation split
 * instructions across multiple transactions and a later slice failed on-chain,
 * error.context.committedHashes lists transaction signatures that succeeded before
 * the failure. Caller must assume those successfully committed changes are permanent
 * and unrollable on-chain.
 */
```

2. **In CCTTxFailedError JSDoc:**
```typescript
/**
 * @property context.committedHashes - Present only when a multi-slice transaction
 * split and a later slice failed on-chain. Lists signatures of prior slices that
 * succeeded. These committed changes are permanent and cannot be rolled back.
 * When this field is present, caller should assume partial state mutation occurred.
 */
```

3. **In SetTokenAuthority:** No change needed (requireSingleTransaction=true prevents split paths).

**Example Flow:**
- SetTokenAuthority(['mint', 'freeze']) called, both fit in one tx → atomicity preserved, committedHashes never set
- Hypothetical operation with requireSingleTransaction=false: 5 instructions, split into 2 txs; tx1 succeeds, tx2 fails → committedHashes = [tx1_sig]

✅ **Recommendation:** Add these JSDoc clarifications so operators understand committed state when errors occur.

aelmanaa:
## 🔵 Code Quality & Test Coverage Nits

**Location:** Multiple files

### 1. Test Coverage Gaps
**File:** `submit.ts`
- Coverage: 96.94% (lines 58-59, 105-109 uncovered)
- Lines 58-59: Likely empty lines or non-executable code
- Lines 105-109: Error path in split logic (e.g., when simulation fails before submission)
- **Action:** Not critical; these are error recovery paths less likely to trigger in happy-path tests. Consider adding a test that simulates the error path (instruction parsing error mid-slice).

### 2. SetTokenAuthority Test Suite
- **Coverage:** All 16 tests passing, 100% code coverage
- **Breadth:** Covers mint+freeze, multisig, revocation, defaults, validation, execution wallet guards
- **Strength:** Test names are clear ("builds selected mint and freeze authority updates", "requires the authority for every selected role")
- ✅ **Action:** No changes needed; test suite is exemplary.

### 3. Deterministic Size Check Quality
- `exceedsTransactionSize` function is well-named and deterministic
- Uses `PACKET_DATA_SIZE` constant from @solana/web3.js (standard 1280 bytes limit)
- ✅ **Action:** Approved; this is production-ready boundary validation.

### 4. Log Clarity (Minor)
**File:** `set-token-authority.ts` line 159-167
```typescript
chain.logger.debug(
  `${this.name}: token = ${opts.tokenAddress.toBase58()}, 
   authorityTypes = ${opts.authorityTypes.join(',')}, 
   newAuthority = ${opts.newAuthority?.toBase58() ?? 'revoked'}`
)
```
- ✅ Good: Log shows key parameters (token, roles, new authority or revocation)
- Minor nit: Consider adding `authority` (current authority) to log for debugging multi-role scenarios
- **Not blocking:** Current log is sufficient for troubleshooting

### 5. Multisig Guard Clarity
**File:** `set-token-authority.ts` lines 187-192
```typescript
if (parsed.multisigSigners.length > 0) {
    this.name,
    'multisigSigners',
    'requires externally signed transactions; use generateUnsignedSetTokenAuthority',
  )
}
```
- ✅ Good: Clear error directing users to unsigned generation for multisig workflows
- **Action:** No changes needed.

✅ **Recommendation:** Approve all code quality aspects. Test suite is exemplary; coverage is production-ready.

aelmanaa:
## 🔵 Cross-Family Consistency: Solana Atomicity vs EVM Txs

**Location:** Reference against PR #413 (EVM liquidity ops)

### Architectural Parity

Both Solana and EVM operations follow the identical parse→buildUnsigned→execute pattern:

| Phase | Solana | EVM |
|-------|--------|-----|
| **Parse** | Validate params, resolve pubkeys | Validate params, checksum addresses |
| **buildUnsigned** | Resolve on-chain state, build instructions, pre-flight checks | Resolve on-chain state, build calldata, pre-flight checks |
| **Execute** | Wallet guard, submit with optional split | Wallet guard, submit (atomic) |

### Atomicity Guarantees

**Solana (PR #416):**
- Multi-instruction operations may exceed PACKET_DATA_SIZE (1280 bytes)
- Can be split across multiple on-chain transactions
- SetTokenAuthority forbids splitting via `requireSingleTransaction=true`
- If a split fails post-broadcast, `committedHashes` tracks partially committed prior slices

**EVM (PR #413):**
- Single transaction per operation (calldata + state changes)
- Atomicity guaranteed by EVM execution model (all-or-nothing per tx)
- No splitting concept; oversized calldata is pre-detected and rejected

### Implication for Multi-Chain Operators

Developers building cross-chain token management workflows must understand:
1. **Solana:** SetTokenAuthority guarantees all roles update atomically *or* fail pre-RPC
2. **EVM:** All role operations (grantMintRole, revokeMintRole, etc. from PR #402) are atomic by design
3. **Error Handling:** Solana may populate committedHashes; EVM errors are simple (success or revert)

### Verification

- Solana: SetTokenAuthority.execute() calls `submit(..., true)` → requireSingleTransaction=true
- EVM: No split logic in any PR #413 operations (liquidity, rebalancer, approve-token all submit single tx)

✅ **Recommendation:** Approve. Cross-family consistency is preserved. SDK documentation should note the atomicity guarantee difference when documenting multi-chain token workflows.

aelmanaa:
## Summary & Recommendation

**PR #416: Prevent Partial Solana CCT Commits**

### What This PR Delivers

1. **New `simulateAndSendCCTTxs` function** in submit.ts
   - Deterministically checks transaction size before any RPC call
   - Splits only on local PACKET_DATA_SIZE boundary, never on simulation/execution errors
   - Reports committedHashes when split transactions fail post-broadcast
   - Supports `requireSingleTransaction` flag for all-or-nothing operations

2. **New `SetTokenAuthority` operation** for Solana token roles
   - Atomically updates mint authority, freeze authority, or both
   - Validates roles before submission, resolves mint state
   - Enforces wallet identity for single-signer mode
   - Blocks multisig in execute() (directs to generateUnsignedSetTokenAuthority)
   - Always calls submit(..., true) to prevent splitting

### Key Strengths

✅ **Production-Ready Design:**
- Atomicity semantics are clear: all roles update or none do
- Split logic is deterministic and well-tested
- Error handling properly reports committed state via committedHashes

✅ **Test Coverage:**
- 16 tests for SetTokenAuthority (all passing)
- Submit test suite validates error mapping and multi-slice behavior
- 96.94% code coverage on submit.ts

✅ **Cross-Family Consistency:**
- Follows identical parse→buildUnsigned→execute pattern as EVM ops
- Atomicity guarantee is Solana-specific; EVM operations are atomic by design

### Areas for Improvement (Non-Blocking)

🟡 **Documentation:**
- Add explanation of WHY SetTokenAuthority forbids splitting (partial state is unrecoverable)
- Clarify committedHashes semantics in JSDoc (when populated, what it means)

🔵 **Code Quality:**
- Submit.ts has 96.94% coverage (3.06% uncovered is acceptable for error paths)
- SetTokenAuthority test suite is exemplary; no changes needed

### Recommendation

**✅ APPROVE** with optional documentation clarifications:
1. Expand SetTokenAuthority class JSDoc to explain the atomicity requirement and why splits are forbidden
2. Add CCTTxFailedError.context.committedHashes documentation clarifying when/how it's populated

**Blocking Items:** None. Code is production-ready as-is.

**Follow-Up Actions:**
- Consider live Devnet test to verify SetTokenAuthority atomicity on real on-chain split scenarios (if multi-role updates are large enough to trigger split without requireSingleTransaction)
- Document in SDK guides that SetTokenAuthority is Solana-only; EVM equivalent is individual grant/revoke calls (PR #402)

aelmanaa:
## The Bug & The Fix

### The Bug

`ccip-sdk/src/solana/utils.ts` `simulateAndSendTxs` splits an instruction list that is too large by catching a simulation failure and dropping the last instruction, then retrying. The catch does not distinguish *why* simulation failed. A program rejection — for example SPL Token `OwnerMismatch` — is treated the same as an over-size/over-budget failure: the failing instruction is dropped and the **valid prefix is signed and committed as its own transaction**.

**Concrete case:** a token has wallet A as mint authority and wallet B as freeze authority. A calls `setTokenAuthority({ authorityTypes: ['mint', 'freeze'], newAuthority: C })`. The combined transaction fails simulation on the freeze instruction with `OwnerMismatch`. The old code drops the freeze instruction, commits the mint-authority change (A → C), retries the freeze update, fails, and returns an error — leaving the mint authority already changed while telling the caller the operation failed.

### The Fix

One shared send routine (`sendInstructionSlices`) replaces the truncate-on-any-error loop and is used by both the CCT ops and the primary send path.

| File | Change |
|---|---|
| `solana/utils.ts` | New `sendInstructionSlices`: (1) shrink a slice by serialized **size** *before* simulating; (2) on a simulation failure, split only if it is a **compute-budget** error, otherwise rethrow; (3) never truncate on a program rejection. `simulateAndSendTxs` now delegates to it, so `ccipSend` and cleanup get the fix too. |
| `cct/solana/submit.ts` | Removes the duplicated send loop; `submit` calls `sendInstructionSlices` with `requireSingleTransaction`, and still reports `committedHashes` on failure. |
| `errors/{codes,recovery,specialized,index}.ts` | Adds `CCIPSolanaTransactionTooLargeError`, thrown when a single instruction does not fit or an atomic op is too large to split (replaces a raw `throw new Error`). |
| `cct/solana/submit.test.ts` | Two unit tests: a compute-budget error splits into two txs; a program rejection on a later instruction is not truncated and commits nothing. |
| `cct/solana/token/operations/set-token-authority.integration.test.ts` (new) | Integration test on Surfpool (local validator, real SPL Token): the on-chain VM rolls both authority changes back, and `setTokenAuthority` rejects without changing the mint authority. |

Compute-budget vs program failure is decided on the **structured** Solana error, not on message text:

```ts
function isComputeBudgetError(error: unknown): boolean {
 if (!(error instanceof SendTransactionError)) return false
 const structured = (error as { simulationError?: unknown }).simulationError
 if (typeof structured === 'string') return COMPUTE_BUDGET_ERRORS.has(structured)
 if (structured && typeof structured === 'object' && 'InstructionError' in structured) {
   const detail = (structured as { InstructionError?: unknown }).InstructionError
   return Array.isArray(detail) && typeof detail[1] === 'string' && COMPUTE_BUDGET_ERRORS.has(detail[1])
 }
 return false
}
```

The split-vs-rethrow decision in the slice loop:

```ts
try {
 unitsConsumed = (await simulateTransaction(ctx, { payerKey, instructions: slice.instructions, ... })).unitsConsumed || 0
 break
} catch (error) {
 // split only on a compute-budget (resource) error; rethrow every program rejection
 if (!requireSingleTransaction && isComputeBudgetError(error) && end - 1 > start) {
   slice = getInstructionSlice(unsigned, start, --end)
   continue
 }
 throw error
}
```

### Properties the Fix Guarantees

| Property | How |
|---|---|
| Truncate only for size and compute failures; rethrow everything else | Size shrink runs before simulation; compute split is gated on the structured error; all program rejections rethrow |
| Classify by structured error type, not message text | Classifier reads `simulationError` (the Solana `TransactionError`), not `logs` or the message |
| Validate fit before submission | `exceedsTransactionSize` size probe runs before each send |
| `setTokenAuthority` mint+freeze with a failing freeze commits nothing | Authority pre-check rejects before submit; the submit path also rethrows the program error without truncating |
| Fix covers the primary send path | The fix is in `simulateAndSendTxs`, so `ccipSend` and cleanup are covered |
| Atomic operations stay atomic | `setTokenAuthority` passes `requireSingleTransaction: true` |
| Size splitting still works | Proven on devnet (below) |

---

## Proof

### P1 — The bug is real (devnet)

The old truncating loop, given `[setMint (valid), setFreeze (fails OwnerMismatch)]`, committed the mint-authority change to the new key and then returned an error. **Evidence:** after the run, the mint's on-chain authority had changed to the new key (partial commit). This is the defect being fixed.

### P2 — The fix prevents the partial commit at the send path (devnet)

The same `[setMint (valid), setFreeze (fails)]` sent directly through the fixed `simulateAndSendTxs` (no authority pre-check involved) throws and commits nothing. **Evidence:** the mint's on-chain authority is unchanged after the run. (No success transaction exists to link — the observable is the mint state.)

### P3 — `setTokenAuthority` is atomic (devnet)

`setTokenAuthority({ authorityTypes: ['mint','freeze'], newAuthority: C })` changes both authorities in one transaction. **Evidence:** tx [4T11U1qt…95jX](https://explorer.solana.com/tx/4T11U1qtzFkwU5LMBp6r3NrnwjHbFzjo76UcfU63BRdtw3PMgvjNZkuYx825eZHiDdPgz5NNcmAiTTbmsxBz95jX?cluster=devnet); both `mintAuthority` and `freezeAuthority` read as C on-chain after that single transaction.

### P4 — The primary send path (`ccipSend`) still works (devnet)

A single-instruction send through `simulateAndSendTxs` sends and confirms. **Evidence:** tx [4jhMS4uU…ukuHX](https://explorer.solana.com/tx/4jhMS4uUN7nhZtnMC2fNNeTRPWrpZzTRuf7n3vLTtMMS81Pya9j5Ha853rU9dDpac8v2tFXussJq3CCaQMmukuHX?cluster=devnet).

### P5 — Size splitting still works (devnet)

8 memo instructions (~2000 bytes, over the 1232-byte limit) split into two confirmed transactions, no truncation. **Evidence:** tx [2ykDP9Fu…w9vb](https://explorer.solana.com/tx/2ykDP9FuZezSG88rK5yBK5vHCpLKLsDqM1Ao6mMSkUgDFbPxcqzxguT5W9gnETFsAdqaYgZNmjWY6JQ2pqtYw9vb?cluster=devnet) and tx [3xAY59Tn…uif3](https://explorer.solana.com/tx/3xAY59Tn4AfBaWcN3L6W8V4f2HkCHiwBaJFm3sMzsJFtdU3Zgj9p7TmJjvgwSwDKZDsVAZg8PeHkSKGg7ZdHuif3?cluster=devnet).

### P6 — On-chain atomicity holds independent of the SDK (Surfpool integration test, control case)

A single transaction with both authority changes fails on the freeze instruction (`OwnerMismatch`) and the validator rolls both changes back. **Evidence:** `set-token-authority.integration.test.ts` control case passes on Surfpool.

### P7 — `setTokenAuthority` commits nothing on a failed combined update (Surfpool integration test, regression case)

**Evidence:** `set-token-authority.integration.test.ts` regression case — after the rejected call, the mint authority is unchanged.

### P8 — A compute-budget failure splits, a program rejection does not (unit)

**Evidence:** `submit.test.ts` — one test drives a `ComputationalBudgetExceeded` simulation error and asserts two transactions are sent; another drives a program rejection on a later instruction and asserts zero transactions are sent. A live >1.4M-compute-unit operation cannot be constructed on-chain, so the compute-split path is proven by unit test only.

### P9 — Nothing regresses; CI is green

**Evidence:** `npm ci` exits 0 (install + build); `tsc`, `oxfmt`, `oxlint` clean on all changed files; 402 Solana/CCT unit tests, 101 error/submit tests, and 2 Surfpool integration tests all pass.

---

## Grounding

- `PACKET_DATA_SIZE = 1232` bytes is the Solana serialized-transaction limit; an oversized transaction fails to serialize (`encoding overruns Uint8Array`).
- SPL `set_authority` changes one authority per instruction and requires the current authority to sign, so a wrong authority reverts `OwnerMismatch`. The pre-check reads the mint and rejects a mismatch before building the transaction, so the same rejection happens off-chain instead of on-chain. Reassign and revoke (`null`) are both handled.
- Blast radius: 30 `submit(...)` call sites, all on the 5-arg signature (the 6th `requireSingleTransaction` is optional); only `setTokenAuthority` opts into `true`. Single-instruction ops are unchanged; multi-instruction ops were never atomic before either (they split by size), so the default preserves prior behavior.

aelmanaa:
## Finding #1 — 🟡 Medium (DX): Partial-Failure State is Hard to Act On

On a split failure, `committedHashes` (which transactions already landed) is only documented on the internal `submit` JSDoc, not on `CCTTxFailedError`/`CCTTxNotConfirmedError`, and the message does not say the state is partially changed.

**Impact:** Callers receiving an error don't know whether partial state was committed and cannot reason about recovery (retry, rollback, etc.).

**Fix:** 
1. Document `context.committedHashes` on both error classes (location: `cct/errors.ts`)
2. Prepend "partially applied: N transaction(s) confirmed" to the reason when committedHashes is present (location: `submit.ts`)

**Example:**
- Before: `Error: simulation failed`
- After: `Error: partially applied: 1 transaction(s) confirmed; simulation failed`

This change makes it clear to operators that prior slices succeeded and are permanent.

aelmanaa:
## Finding #2 — 🟡 Medium (DX): Authority Pre-Check Error Message Lacks Context

The authority pre-check error does not name the current authority: "must match the token mint authority" does not say what the correct value is, and the already-revoked (`null`) case shows the same text.

**Impact:** When validation fails, callers don't see the on-chain authority value and must make a separate query to understand what went wrong.

**Fix:** Enhance error message with the actual value (location: `set-token-authority.ts:143-153`):
- Current: `"must match the token mint authority"`
- Suggested: `"authority ${provided} is not the current ${type} authority (${current ?? 'already revoked'})"`

**Alternatively:** Put `currentAuthority` in the error `context` object for programmatic access.

**Example:**
- Before: `Error: authority must match the token mint authority`
- After: `Error: authority ABC... is not the current mint authority (XYZ...)`

This reduces the round-trip to understand validation failures.

aelmanaa:
## Finding #3 — 🟡 Low: Missing @throws Documentation on execute()

The class docs cover the atomicity guarantee, but the public `execute` facade has only a one-line doc.

**Fix:** Add `@throws` for all error types thrown by this operation (location: `set-token-authority.ts:176-208`):
- `CCTParamsInvalidError` (invalid authority types, mismatched authority, multisig on execute)
- `CCIPTokenDataParseError` (malformed mint data)
- `CCTTxFailedError` (on-chain program rejection, e.g., wrong authority)
- `CCTTxNotConfirmedError` (blockhash expired, not confirmed, etc.)

**Example:**
```ts
/**
 * Generate, sign, simulate, send, and confirm with the current authority wallet.
 * 
 * @throws {CCTParamsInvalidError} If authority types are invalid or if authority does not match the executing wallet
 * @throws {CCIPTokenDataParseError} If the mint data cannot be parsed
 * @throws {CCTTxFailedError} If the transaction fails on-chain (e.g., wrong current authority)
 * @throws {CCTTxNotConfirmedError} If the transaction is not confirmed in time
 */
override async execute(...)
```

---

## Finding #4 — Nits (Polish)

**(a) Rename the concept from "truncation"**

An oversized transaction hard-fails at serialize time, it is not silently truncated. "Size overflow" is more accurate.

**(b) Clarify the atomicity wiring**

The atomicity is wired by a bare positional `true` in the `submit(...)` call. Add an inline comment:
```ts
return submit(..., /* requireSingleTransaction */ true)
```

**(c) Document ALT-only-on-main-slice behavior**

The lookup-tables-only-on-the-main-slice behavior silently drops ALTs on non-main slices. Safe today (no CCT op sets `lookupTables`), but add a comment/assert so a future ALT-bearing op keeps ALT-referencing instructions on the main slice.

**(d) Namespace committedHashes to avoid key collisions**

`apply-chain-updates.ts` does `Object.assign(error.context, {committedHashes})`, which could overwrite the `committedHashes` the submit path sets — moot today (batches are packed to one tx), but namespace it to avoid a latent key collision.

aelmanaa:
## Recommendation & Summary

✅ **APPROVE.** The bug is reproduced (P1), the fix prevents it (P2), and both are proven on a local validator and on devnet. The remaining items are documentation/DX polish (findings above), not blockers.

**What ships:**
1. ✅ Atomicity guarantee for `setTokenAuthority` (mint + freeze in one tx, or none)
2. ✅ Program rejection rethrow (no truncation on `OwnerMismatch`, etc.)
3. ✅ Size-based splitting still works (deterministic PACKET_DATA_SIZE check)
4. ✅ All proofs validated (9 scenarios: Devnet txs + integration tests + unit tests)
5. ✅ No regressions (402 Solana/CCT + 101 error/submit + 2 Surfpool integration tests pass)

**Polish items (non-blocking):**
- Finding #1: Document and clarify `committedHashes` on error types
- Finding #2: Enhance authority pre-check error message with actual values
- Finding #3: Add `@throws` to `execute()` facade JSDoc
- Finding #4: Code polish (rename concept, clarify comments, namespace collisions, ALT handling)

**Recommendation:** Ship. Polish can follow in a focused commit if preferred.

mervin-link:
> ## Recommendation & Summary
> ✅ **APPROVE.** The bug is reproduced (P1), the fix prevents it (P2), and both are proven on a local validator and on devnet. The remaining items are documentation/DX polish (findings above), not blockers.
> 
> **What ships:**
> 
> 1. ✅ Atomicity guarantee for `setTokenAuthority` (mint + freeze in one tx, or none)
> 2. ✅ Program rejection rethrow (no truncation on `OwnerMismatch`, etc.)
> 3. ✅ Size-based splitting still works (deterministic PACKET_DATA_SIZE check)
> 4. ✅ All proofs validated (9 scenarios: Devnet txs + integration tests + unit tests)
> 5. ✅ No regressions (402 Solana/CCT + 101 error/submit + 2 Surfpool integration tests pass)
> 
> **Polish items (non-blocking):**
> 
> * Finding [Chore/publish ci #1](https://github.com/smartcontractkit/ccip-tools-ts/pull/1): Document and clarify `committedHashes` on error types
> * Finding [Prepare release v0.1.0 #2](https://github.com/smartcontractkit/ccip-tools-ts/pull/2): Enhance authority pre-check error message with actual values
> * Finding [chore: ccip-data as codeowners #3](https://github.com/smartcontractkit/ccip-tools-ts/pull/3): Add `@throws` to `execute()` facade JSDoc
> * Finding [feat: parseBytes #4](https://github.com/smartcontractkit/ccip-tools-ts/pull/4): Code polish (rename concept, clarify comments, namespace collisions, ALT handling)
> 
> **Recommendation:** Ship. Polish can follow in a focused commit if preferred.

@aelmanaa Addressed all comments.

apedrob (approved):

andrevmatos:
## Plan: move CCT Solana submission onto the shared split modes (#417 merged, #441 open)

`main` now has split modes in the shared `simulateAndSendTxs` (#417), and #441 adds `{ computeUnits, split }` options, the `{ hash, slices }` return, cut-at-failed-instruction, and the **Solana v1 fix**. Once `cct-sdk` picks those up, I suggest replacing the CCT-local `simulateAndSendCCTTxs` here with the shared engine instead of keeping a fork:

```ts
// cct/solana/submit.ts
const { hash, slices } = await simulateAndSendTxs(chain, wallet, unsigned, {
  computeUnits,
  split: requireSingleTransaction ? 'atomic' : 'resource',
})
```

Why not keep the local fork:
- **It's v0-only (1232 B).** Devnet *and mainnet* accept v1 transactions now (4096 B, live-verified; evidence table in #441). The shared engine sends v1 when v0 doesn't fit, with the SIMD-0385 resource limits set correctly. Every v1 simulation from the pre-#441 code fails on real clusters with `MaxLoadedAccountsDataSizeExceeded`. So large CCT ops (Metaplex `deployToken` + premint, `editChainRemoteConfig` / `appendRemotePoolAddresses` with many pools, `createLookupTable` with >1 extend, long `deployTokenPool` allowlists) fit in one tx on v1 networks instead of failing or splitting.
- `'resource'` is the local fork's semantics: split only on size or compute budget, and rethrow program errors with nothing sent. It also cuts compute failures at the failed instruction instead of aborting. `'atomic'` = `requireSingleTransaction`.
- It avoids a third copy of the loop (explorer's `splitAndSerializeTxsSolana` is the second).

### `createCCTSubmitError`
Unwrap `CCIPPartialTransactionSubmissionError`:
- use `cause` for `getSignature` / `isNotConfirmedError` / the transient regex;
- `context.pendingSignature` (sent, not confirmed) → `CCTTxNotConfirmedError` txHash;
- `context.committedSlices` (`{ signature, start, end }`) → keep `committedHashes`, and add the ranges.

Otherwise a confirmation timeout on a later slice surfaces as a permanent `CCTTxFailedError`.

### Split mode per operation

| Op | Instructions | Must stay together? | Mode |
|---|---|---|---|
| `setTokenAuthority` | 1–2 × `setAuthority` | yes (ToB repro) | `atomic` (plus this PR's authority pre-check 👍) |
| `createTokenMultisig` | `createAccountWithSeed` + `initializeMultisig` | **yes: init needs no signer, so a split lets anyone initialize the seeded account first** | `atomic` |
| `deployToken` | create + `initMint2` [+ Metaplex] [+ ATA + `mintTo`] | create+init: yes (same front-run risk); rest is sequential | `resource` (the greedy first slice keeps create+init together). Put `tokenAddress` in the error context. |
| `deployTokenPool` | `initialize` [+ ATA] [+ `configureAllowList(N)`] | yes (otherwise the pool briefly has no allowlist) | `atomic`; long allowlists → `configureAllowlist` separately |
| `provideLiquidity` | [approve] + provide | yes (otherwise an approval is left behind) | `atomic` |
| `mintTokens` | [ATA] + `mintTo` | loosely | `atomic` |
| `createLookupTable` (create + extend) | create + ⌈n/30⌉ extends | order only (CCIP addresses first) | `resource`; return **all** hashes, and put `lookupTableAddress` + committed count in the error |
| `appendToLookupTable` | ⌈n/30⌉ extends | incremental | `resource` |
| `applyChainUpdates` batches | own group-atomic v0 packer | per remote chain | `atomic` per batch. Keeping the packer v0-only makes batches the same on every RPC. |
| single-instruction ops | 1 | — | `atomic` |

Suggested shape: `protected readonly splitMode: SolanaSplitMode = 'atomic'` on `SolanaOperation`, with `execute` passing it to `submit`, and overrides on `deployToken` / `createLookupTable` / `appendToLookupTable`.

### v0-only vs v1 RPCs
- v1 RPC: oversized slices go out as one v1 tx (≤4096 B, ≤64 accounts / 64 instructions / 12 signatures; no ALTs, and no CCT op uses them).
- v0-only RPC: the v1 attempt returns JSON-RPC `-32602`, which becomes `CCIPTransactionTooLargeError`. `resource` then splits, and `atomic` throws `TooLarge`.
- So the number of txs for a `resource` op depends on the RPC; results should always carry every signature.

### Other items found while mapping ops
- Resuming after a partial failure: `deployToken` / `createLookupTable` pick a new random seed or slot on each call, so a retry creates a new mint or ALT. `appendToLookupTable` rejects an ALT that already holds *some* CCIP addresses, so a half-extended create+extend can't be resumed; it should allow appending the missing tail.
- `serializeUnsignedTx` (vault/external signing) compiles one legacy message, ≤1232 B, with no split and no v1. It's atomic by construction, but large ops can't go through it; worth documenting.
- Wallets: the v1 fallback is fine for keypair-style wallets; browser wallets / Squads should use the `generateUnsigned*` path.

andrevmatos:
Addendum to the plan above: **hardware wallets and v1**. Ledger's Solana app can't sign v1 transactions yet. It parses them as malformed v0 messages and fails with `0x6a80` ([LedgerHQ/app-solana#248](https://github.com/LedgerHQ/app-solana/issues/248)).

#441 (`6013402b`) handles this in the shared engine:
- the `Wallet` type takes an optional `supportedTransactionVersions` (wallet-adapter semantics);
- a wallet without `1` never gets a v1 transaction. A slice too large for v0 becomes `CCIPTransactionTooLargeError`: it is split in `'resource'` mode and rejected in `'atomic'` mode;
- the CLI's Ledger wallet declares legacy/v0 only.

CCT operations on the shared engine inherit this. It's one more reason not to keep the local v0-only fork, since both would otherwise need wallet-capability handling. For Ledger/Squads users, `atomic` CCT ops that only fit v1 (e.g. long `deployTokenPool` allowlists, `editChainRemoteConfig` with many pools) fail with `TooLarge`. The error should point them to splitting the operation.

PR #417, opened 2026-09-10
fix(sdk): Add Solana transaction split modes

mervin-link:
## What

- Add `partial`, `resource`, and `atomic` split modes to Solana `simulateAndSendTxs`
- Preserve existing partial-split behavior by default while supporting resource-only and no-split execution
- Surface confirmed prefix transaction signatures through `CCIPPartialTransactionSubmissionError.context.committedHashes`
- Add unit coverage for atomic, resource, and partial program-error behavior
- Addresses [TOB-CLCCT-1](https://smartcontract-it.atlassian.net/browse/DAPP-11383)

## Why

- Provide explicit split-mode support for the CCT follow-up while preserving existing non-CCT behavior.
- Preserve partial as the default for backward compatibility with existing non-CCT workflows that intentionally depend on sequential transaction submission

## Testing

- Typecheck
- Unit tests: `npm --prefix ccip-sdk test -- src/solana/v1.test.ts`

## Notes

- `partial` remains the default for backward compatibility and CCT callers should explicitly select resource or atomic based on their operation contract.
- This PR adds the internal split-mode capability without changing existing non-CCT callers. The CCT follow-up will wire `submit()` to select modes per operation: atomic for coupled admin/deployment operations (including multi-authority updates) and resource only for explicitly incremental workflows such as lookup-table extension.


aelmanaa:
## Grounding & Live Proofs (devnet) — PR #417 Solana Transaction Split Behavior

**Branch** `fix/DAPP-11383-simulate-and-send-tx` → **main**. Head `160851e9`, base `0f3836dd` (net-new commit `3b8a47c3`). **Verdict: APPROVE-WITH-CHANGES.**

Adds `splitMode: 'partial'|'resource'|'atomic'` (default `'partial'`) to the internal Solana `simulateAndSendTxs`, an `isComputeBudgetError` classifier, and a `committedHashes` accumulator raising a new `CCIPPartialTransactionSubmissionError`. Panel (typescript-pro, sdk-cli-architect, dx-engineer) + `chainlink-ccip`/Solana grounding + **live proofs on real Solana devnet**. Every suggested fix below is **applied in the review worktree, built, and proven** (patch: `docs/11383-review-fixes.patch`, 5 files; `npm run check` clean, 16/16 Solana tests, oxlint clean), except three items explicitly marked *Recommended (not applied)* with reasons.

---

### Grounding

`isComputeBudgetError` checked against Solana source (`InstructionError::ComputationalBudgetExceeded` is a unit variant; no top-level `TransactionError` of that name) and **confirmed live** on `api.devnet.solana.com` (payer `2SGSoyjD…C3V3`), simulate-only — the change is simulation-error classification, no broadcast needed. `chainlink-ccip` has no equivalent off-chain splitter/classifier (only a Go CU-limit builder), so nothing upstream to align to.

### Live proofs (real devnet)

| Proof | Input | Real devnet `err` | classified |
|---|---|---|---|
| P1 (real SDK path) | `simulateTransaction(computeUnitsOverride:1)` | `{InstructionError:[0,"ComputationalBudgetExceeded"]}` | **true** ✅ |
| P2 (Finding 2 fix) | raw simulate @300 CU | `{InstructionError:[1,"ProgramFailedToComplete"]}` + `"exceeded CUs meter"` log | **true** (was **false** pre-fix) ✅ |
| P3 (precision) | `ProgramFailedToComplete`, no meter log | — | **false** (not over-broadened) ✅ |
| P4 | `Error`, `{Custom:6000}` | — | false ✅ |

aelmanaa:
## 🟠 Major (correctness) — The new error defeated `execute()`'s retry-escalation — FIXED + PROVEN

`solana/index.ts:1472` escalates oversized-report recovery by string-matching `err.message` (`AlreadyContainsChunk`, `too large`). Once a prefix commits (e.g. `forceBuffer` writes chunks, then `execIx` is still too large), `simulateAndSendTxs` throws `CCIPPartialTransactionSubmissionError` whose message is the fixed `"Transaction submission partially completed"` — matches no keyword, so the buffer→ALT ladder silently dies.

**Fix:** match against the wrapped `cause`.

**Proof:** typecheck clean, 16/16 tests; targeted proof — wrapper message escalates=`false`, cause-aware escalates=`true`, `committedHashes` preserved (`cause` is wired via `CCIPError`).

aelmanaa:
## 🟠 Major (correctness) — `isComputeBudgetError` missed `ProgramFailedToComplete` — FIXED + PROVEN live

A program exhausting budget mid-execution surfaces as `ProgramFailedToComplete`, not `ComputationalBudgetExceeded` (reproduced live @300 CU). Pre-fix the classifier returned `false`, so `'resource'` mode would not split.

**Fix (precise, not over-broad):** also treat `ProgramFailedToComplete` as compute-budget **only when** the logs carry `"exceeded CUs meter"` — the BPF meter line that distinguishes compute exhaustion from other program failures (verified against the live devnet log).

**Proof:** P2 now `true`, P3 (no meter log) stays `false`. Also dropped the redundant `as {InstructionError?…}` cast (ts-pro nit) in the same edit; the dead bare-string branch left as-is (harmless).

aelmanaa:
## 🟠 Major (scope) — `'atomic'`/`'resource'` are half-landed — RECOMMENDED (not applied — design decision)

`simulateAndSendTxs`/`splitMode` are SDK-internal (not re-exported) and no caller passes anything but the default, so the headline capability is unreachable by consumers. There is no single correct snippet — the author should either wire a caller (natural fit: `execute()` → `'resource'`, so exec only re-splits on compute/size, not program errors) **or** state in the PR/CHANGELOG that the modes are an internal foundation for a follow-up.

I did not apply a wiring because it changes `execute()`'s split semantics — the author's call, and I won't prescribe an unproven behavior change.

aelmanaa:
## 🟠 Major (DX) — Recovery hint over-promised; no resume path — FIXED + PROVEN

The hint told callers to inspect `committedHashes` "so confirmed transactions are not submitted again," but the hashes couldn't map to instructions and there was no resume affordance — re-invoking replays everything.

**Fix:** added `committedInstructionCount` (the confirmed instruction prefix length) to the error, reworded the hint to be honest about replay, and added an `@example` (also resolves the ts-pro/DX "no `@example`, `committedHashes` undiscoverable" nit).

**Proof:** error now carries `committedInstructionCount`, `cause` preserved, hint updated, build green, 16/16 tests.

aelmanaa:
## Nits — Applied

- `index.ts:1409,1451` — `@throws {@link CCIPPartialTransactionSubmissionError}` added to `sendMessage`/`execute`. **Applied.**
- `utils.ts` dead code — removed unreachable `if (end <= start) throw lastErr` and the write-only `lastErr` (ts-pro). **Applied + proven** (typecheck/tests green).
- `CHANGELOG.md` — `[Unreleased]` entry added (main-line PR; main logs comparable SDK changes — unlike cct-sdk-stacked branches). **Applied.**

## Nits — Recommended (not applied)

- `utils.ts:625` `splitMode` has no runtime validation (an out-of-union value silently acts as `'atomic'`). **Recommended (not applied)** — the param is typed and unreachable from CLI/typed callers; a guard is defensive-only.
- `'resource'` mode name isn't self-teaching. **Recommended (not applied)** — cosmetic; the `@param` explains it.

## Verified Clean (no action)

- Object-shape `ComputationalBudgetExceeded` classification correct + live-confirmed (P1).
- Default `'partial'` splitting is **behavior-equivalent** to pre-PR (ts-pro traced single/double/truncate); only the *error type* on a partially-committed send changed, which is safe.
- `'resource'` size-split branch is **reachable** (not dead): `simulateTransaction`'s v1 fallback throws `CCIPTransactionTooLargeError` during simulation (`v1.ts:144/180/186`) — I was wrong to suspect a dead branch.
- Error taxonomy structurally correct; no CLI parity gap (`formatCCIPError` renders context + hint generically).

aelmanaa:
## Recommendation

**Approve with changes.** Findings 1, 2, 4 and the applicable nits are fixed and proven here (build + live devnet). Finding 3 (half-landed modes) needs an author decision — wire a caller or document the intent — before the PR delivers the behavior its title implies. `cct-sdk` is unaffected; this is a main-line Solana fix.

mervin-link:
> ## Recommendation
> **Approve with changes.** Findings 1, 2, 4 and the applicable nits are fixed and proven here (build + live devnet). Finding 3 (half-landed modes) needs an author decision — wire a caller or document the intent — before the PR delivers the behavior its title implies. `cct-sdk` is unaffected; this is a main-line Solana fix.

@aelmanaa Addressed all comments except for calling 'atomic'/'resource'. This will be a follow up.

andrevmatos:
## Path to merge (least churn)

Proposal: land this PR as the **foundation only** (`splitMode` + `CCIPPartialTransactionSubmissionError`, default `'partial'`, no behavior change for main callers) and move all wiring/redesign to a follow-up PR stacked on this branch (link below once opened). With that scope, only these small items are left here:

**Must**
1. `ccip-sdk/src/solana/utils.ts:681` — `if (end <= start) throw lastErr` is unreachable (the `catch` either decrements with `end - 1 > start` still holding, or throws), and `lastErr` is write-only. The previous review marked this as applied, but it's still on the branch; please drop both.
2. `CHANGELOG.md` `[Unreleased]` entry — also marked as applied but missing. Something like: *"Solana: `simulateAndSendTxs` split modes (`partial` default, `resource`, `atomic`); a submission failing after earlier transactions confirmed throws `CCIPPartialTransactionSubmissionError` with `context.committedHashes`"*.
3. PR description: say explicitly that main callers (`sendMessage`, `execute`, cleanup) still use `'partial'`, so TOB-CLCCT-1 / DAPP-11383 **stays open** until the follow-up wires them (ticket AC1: truncation only for size/compute failures).

**Optional (fine to leave)**
- Tests use `InstructionError: [0, …]`, which in the v0 path points at the prepended compute-budget instruction (the caller's instructions start at index 1). Harmless today, since the classifier ignores the index; the follow-up uses the index and updates these tests.

**Deferred to the follow-up (no action here)**
- `simulateAndSendTxs(ctx, wallet, unsigned, { computeUnits, split })` returning `{ hash, slices }`; error context `committedSlices` (`{ signature, start, end }`) + `pendingSignature` (sent but unconfirmed)
- `sendMessage` → `'resource'`: token approvals are no longer committed when `ccipSend` fails simulation (the same class as the ToB repro)
- Cut a failing slice right before the failed instruction from the `InstructionError` index (one simulation per split point instead of one per dropped instruction; first-instruction failures throw immediately)
- `computeUnits` applied only to the main instruction's transaction (as its doc says); return the last signature when `mainIndex` is unset (cleanup currently gets `undefined` and re-sends the ALT close)
- `execute()` retry: unwrap `CCIPPartialTransactionSubmissionError` by type instead of joining messages; also escalate on `CCIPTransactionTooLargeError`
- `execute()` stays `'partial'` for now: `'resource'` would throw on `forceLookupTable` flows where an ALT extend and its first use land in the same tx (tx-level ALT error, not compute/size). Needs an explicit "ALT create/extend never shares a tx with its use" split rule + fork coverage for `forceBuffer`/`forceLookupTable` first.

**Unrelated, separate PR (pre-existing on main since 134ccde5)** — v1 compliance: per SIMD-0385 an unset compute-unit-limit / loaded-accounts-data-size config bit budgets **0** (no defaults, unlike v0), and v1 caps are 64 accounts / 64 instructions / 12 signatures. Today v1 sends get CU 0 when simulation used ≤200k, and loaded-data 0 always. Surfpool doesn't enforce this, so the fork test passes.

**Heads-up for `cct-sdk`**: `cct/solana/submit.ts` must unwrap `cause` of `CCIPPartialTransactionSubmissionError` — `getSignature` / `isNotConfirmedError` / the transient regex only look at the top-level error, so a confirmation timeout on a later tx would surface as a permanent `CCTTxFailedError`.

LGTM once 1–3 are done.

mervin-link:
> ## Path to merge (least churn)
> Proposal: land this PR as the **foundation only** (`splitMode` + `CCIPPartialTransactionSubmissionError`, default `'partial'`, no behavior change for main callers) and move all wiring/redesign to a follow-up PR stacked on this branch (link below once opened). With that scope, only these small items are left here:
> 
> **Must**
> 
> 1. `ccip-sdk/src/solana/utils.ts:681` — `if (end <= start) throw lastErr` is unreachable (the `catch` either decrements with `end - 1 > start` still holding, or throws), and `lastErr` is write-only. The previous review marked this as applied, but it's still on the branch; please drop both.
> 2. `CHANGELOG.md` `[Unreleased]` entry — also marked as applied but missing. Something like: _"Solana: `simulateAndSendTxs` split modes (`partial` default, `resource`, `atomic`); a submission failing after earlier transactions confirmed throws `CCIPPartialTransactionSubmissionError` with `context.committedHashes`"_.
> 3. PR description: say explicitly that main callers (`sendMessage`, `execute`, cleanup) still use `'partial'`, so TOB-CLCCT-1 / DAPP-11383 **stays open** until the follow-up wires them (ticket AC1: truncation only for size/compute failures).
> 
> **Optional (fine to leave)**
> 
> * Tests use `InstructionError: [0, …]`, which in the v0 path points at the prepended compute-budget instruction (the caller's instructions start at index 1). Harmless today, since the classifier ignores the index; the follow-up uses the index and updates these tests.
> 
> **Deferred to the follow-up (no action here)**
> 
> * `simulateAndSendTxs(ctx, wallet, unsigned, { computeUnits, split })` returning `{ hash, slices }`; error context `committedSlices` (`{ signature, start, end }`) + `pendingSignature` (sent but unconfirmed)
> * `sendMessage` → `'resource'`: token approvals are no longer committed when `ccipSend` fails simulation (the same class as the ToB repro)
> * Cut a failing slice right before the failed instruction from the `InstructionError` index (one simulation per split point instead of one per dropped instruction; first-instruction failures throw immediately)
> * `computeUnits` applied only to the main instruction's transaction (as its doc says); return the last signature when `mainIndex` is unset (cleanup currently gets `undefined` and re-sends the ALT close)
> * `execute()` retry: unwrap `CCIPPartialTransactionSubmissionError` by type instead of joining messages; also escalate on `CCIPTransactionTooLargeError`
> * `execute()` stays `'partial'` for now: `'resource'` would throw on `forceLookupTable` flows where an ALT extend and its first use land in the same tx (tx-level ALT error, not compute/size). Needs an explicit "ALT create/extend never shares a tx with its use" split rule + fork coverage for `forceBuffer`/`forceLookupTable` first.
> 
> **Unrelated, separate PR (pre-existing on main since [134ccde](https://github.com/smartcontractkit/ccip-tools-ts/commit/134ccde581f45133b41e4e0eb3b52252760d5fe6))** — v1 compliance: per SIMD-0385 an unset compute-unit-limit / loaded-accounts-data-size config bit budgets **0** (no defaults, unlike v0), and v1 caps are 64 accounts / 64 instructions / 12 signatures. Today v1 sends get CU 0 when simulation used ≤200k, and loaded-data 0 always. Surfpool doesn't enforce this, so the fork test passes.
> 
> **Heads-up for `cct-sdk`**: `cct/solana/submit.ts` must unwrap `cause` of `CCIPPartialTransactionSubmissionError` — `getSignature` / `isNotConfirmedError` / the transient regex only look at the top-level error, so a confirmation timeout on a later tx would surface as a permanent `CCTTxFailedError`.
> 
> LGTM once 1–3 are done.

@andrevmatos 

- Addressed the unreachable error path
- Added the CHANGELOG.md entry 
- Clarified the PR description
- Updated the mocked `InstructionError` indices in tests.

andrevmatos (approved):
LGTM — all three items addressed (dead `lastErr`/`if (end <= start)` removed; the loop still exits only via `break`/`throw`), CHANGELOG entry added, and the description now states main callers stay on `'partial'` / TOB-CLCCT-1 remains open. Thanks for also fixing the `InstructionError` indexes in the tests.

Before merging:
- **`sigscanner-check`** fails because `87832d39` is unsigned (`verification.reason: unsigned`); please re-sign it (e.g. `git commit --amend --no-edit -S` + force-push).
- **`build-and-test`**: the 7 failures are all in `evm/fork.test.ts` / `evm/integration.test.ts` (v1.5/v1.6 lane fixtures broken by the Sep-2026 OnRamp 2.0 migration on testnets), unrelated to this PR. A separate test-fixture PR will land on `main` first; merging `main` in after that should turn CI green.

andrevmatos:
Follow-ups are up:
- #440 — EVM fork/integration lane fixtures (split from #390, same patch); lands on `main` first so this PR's `build-and-test` goes green after merging `main` in.
- #441 — the deferred items from my review above (options object + `{ hash, slices }`, `sendMessage` → `'resource'`, cut at the failed instruction, `committedSlices`/`pendingSignature`, main-only `computeUnits`, `execute()` unwrap by type). Stacked on this branch; I'll rebase it onto your re-signed head once you force-push.

PR #420, opened 2026-09-14
fix(cct-sdk): report unmet registry state instead of refusing to build

marek-sadura:
## Problem

`generateUnsignedAcceptAdmin` and `generateUnsignedTransferAdmin` throw if the TokenAdminRegistry isn't already in the required state (no pending admin, or `sender` isn't the current admin). That breaks offline/multisig planning: a plan that registers an admin and then accepts can't be built, because at build time the registry hasn't been written yet.

## Solution

The calldata doesn't depend on that state, so `buildUnsigned` now attaches what it found to the tx instead of discarding it:

```ts
UnsignedEVMTx.preconditions?: { param: string; reason: string }[]
```

`generate*` returns calldata + any unmet requirement. `execute` still throws the same `CCTParamsInvalidError` as before, and doesn't submit. No extra RPC — it rides the read that already happens. `preconditions` is absent on the happy path, so existing consumers are unaffected.

Also: `parse` now rejects a zero `sender`, which was previously only caught as a side effect of the check that no longer throws.

Scope: `acceptAdmin` + `transferAdmin` only. The token-pool ops have a related deploy-then-configure limitation — separate change.

Tests: 2562 passing, lint and typecheck clean.


apedrob (commented):
Thanks @marek-sadura, we could have a skip validation flag, but I agree with the goal of keeping `sender` and the onchain checks, and reporting them. That will be useful for reviewable artifacts and operation bundling

I would suggest changes on how the report is delivered tho:
- `preconditions` on `UnsignedEVMTx`: it's the core type every EVM consumer uses (outside of `/cct`), so I would rather keep CCT concerns out of it
- `preflight` switch: I would keep the op always running the same checks, and the caller decides what to do with the result

Proposal: keep `generateUnsigned*` as is and let the error carry the result. Created this [gist](https://gist.github.com/apedrob/c4fa4df6ba50aa14d237a1fdc7847f08) as an implementation example.

This will change getting the preconditions check from the error instead of as a return value, will that work for you?

PR #382, opened 2026-08-27
feat(cct-sdk): Add apply chain updates evm op

apedrob:
## What

- [DAPP-11279](https://smartcontract-it.atlassian.net/browse/DAPP-11279)
- Add `applyChainUpdates` / `generateUnsignedApplyChainUpdates` to `EVMTokenManager`,
configuring, enabling and removing a token pool's remote lanes: remote token, remote pool(s),
and both directional rate limits
- Params are version-discriminated, because the contract signature changed at v1.5.1:
  - `V1.5.0` takes a single `chains` array with an inline `allowed` bit and a singular
`remotePoolAddress`
  - `V1.5.1` splits into `chainsToAdd` / `remoteChainSelectorsToRemove`
with plural `remotePoolAddresses`, and is also the shape for v1.6.1 and v2.0.0
- The declared version is checked against the pool's on-chain `typeAndVersion`, and `sender`
against its `owner()`, so a wrong shape or an unauthorized caller is a param error before signing
- `getTokenPoolState` now reads v2.0.0 `SiloedLockReleaseTokenPool` (every field but `lockBox`,
which is per-lane) instead of rejecting it

## Why

- Lane configuration is the last write op needed to take a token pool from deployed to live;
without it the SDK can create and register a pool but not enable a route

## Testing

- Unit tests cover both calldata shapes, the version/owner pre-flights, rate-limit bounds per
generation, and the lane-array param rules
- New coverage for `resolveEncoder`'s removal ceilings and for the siloed-pool read path

## Notes

- `Operation` now owns the `validate` → `parse` → `prepare` lifecycle (hoisted from
`SolanaOperation`), so EVM ops can hand `buildUnsigned` already-narrowed params; existing ops
were migrated in place
- `resolveEncoder` accepts an explicit `null` to mark "removed from this version up", so a
removed function can no longer floor-match an older encoder and emit an unimplemented selector

[DAPP-11279]: https://smartcontract-it.atlassian.net/browse/DAPP-11279

PR #423, opened 2026-09-15
feat(cct-sdk): Add finality config ops EVM

mervin-link:
## What

- [DAPP-11552](https://smartcontract-it.atlassian.net/browse/DAPP-11552)
- Add EVM CCT operations to read and set v2.0.0 token-pool allowed finality config
- Expose unsigned, signed, and read APIs on `EVMTokenManager`
- Add validation, version/owner preflight checks, API docs, and unit tests

## Why

- Let pool owners configure FTF depth and FCR/safe-finality support through the SDK, and let clients read the active configuration.

## Notes

- Both operations support v2.0.0 pools only; earlier pool versions return an unsupported-operation error.

[DAPP-11552]: https://smartcontract-it.atlassian.net/browse/DAPP-11552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aelmanaa:
## Grounding & Proof — DAPP-11552 Finality Config Ops EVM

**Head** `93278ace` (single commit on `cct-sdk` @ `aa3968d5`). **Verdict: APPROVE-WITH-CHANGES.**

Adds two v2.0.0-only EVM TokenPool ops — `getAllowedFinalityConfig` / `setAllowedFinalityConfig` (+ `generateUnsigned…`) on `EVMTokenManager` — encoding the pool's `bytes4` finality config from the SDK-level `FinalityAllowed` `{ finalityDepth, finalitySafe? }`. +596 lines, 6 files. Builds green in isolation (`npm run check` clean, 28/28 new tests). All findings and all fixes below are proven — on real Sepolia for behavior, by compiler for type surface.

---

### Grounding (contract truth)

The `bytes4` encoding is **correct**, cross-checked two ways:

- Pinned `FinalityCodec.sol` (`docs-cct-foundry` deps) and the canonical `chainlink-ccip` clone (`d077fd91`, `TokenPool.sol:220-258`, `FinalityCodec.sol`) agree: safe flag = bit 16 (`WAIT_FOR_SAFE_FLAG = 1<<16 = 0x00010000`), depth = low 16 bits (`BLOCK_DEPTH_MASK = 0xFFFF`, max 65535), `WAIT_FOR_FINALITY_FLAG = bytes4(0)`.
- SDK `encodeFinality` (`extra-args.ts`) — `(flags<<16)|(depth&0xffff)`, `flags|=0x1` on safe, depth range `[0,65535]` — matches byte-for-byte. `setAllowedFinalityConfig` is `onlyOwner` on-chain; the SDK's owner-only assumption is correct. Allowed-finality legitimately combines safe + depth (unlike *requested* finality, which is single-mode) — the SDK permits the combination, which is right for this setter.

### DRY / consistency

Clean. The op reuses `extra-args.ts` (`encodeFinality`/`decodeFinalityAllowed`/`FinalityAllowed`) rather than re-packing bits; `readTokenPoolAllowedFinality` follows the `readTokenPoolOwner` idiom; structure, version dispatch, owner-gating, and error taxonomy mirror the sibling pool ops. No duplicated logic, no `new Error()`, no `any`/unsafe cast.

---

### Proof — verified live, not just by reading source

Every mutation a confirmed Sepolia tx against a real `BurnMintTokenPool 2.0.0` ([`0x92af99b7…`](https://sepolia.etherscan.io/address/0x92af99b75a623Fa0Abd164f8EAa6BEF030824D9A), owner = `0x9d08…b717`):

| Step | Set | SDK-encoded | On-chain `bytes4` | Tx |
|---|---|---|---|---|
| Round-trip | `{depth:5}` | `0x00000005` | `0x00000005` ✅ | [`0xd31283c5`](https://sepolia.etherscan.io/tx/0xd31283c5be117b651b64cabdcff6e4b2e9171d417b188bd14eba92fc306022ae) |
| Combined mode | `{depth:32,safe:true}` | `0x00010020` | `0x00010020` ✅ | [`0x97824a38`](https://sepolia.etherscan.io/tx/0x97824a38ac81b12af53918390e0df86260eebeb5c8d644eaef88d74337447328) |
| Full-replace (Comment 2) | `{depth:20}` | `0x00000014` | `0x00000014`; FCR `true→false` ⚠️ | [`0x228cf5b1`](https://sepolia.etherscan.io/tx/0x228cf5b1b36de572010fd4f5cc21885ef8423ab92cdf987d88558f73a01d50d2) |
| Disable/reset | `{depth:0}` | `0x00000000` | `0x00000000` ✅ | [`0x3378f0c1`](https://sepolia.etherscan.io/tx/0x3378f0c1d25650ae2f3da596c036bd042f56c7d85ffd62bb2a38a21aa2bd8027) |

aelmanaa:
## 🟠 Major — Get/Set param & result types not exported from the `cct/evm` entry point

`cct/evm/index.ts` is a published entry (`@chainlink/ccip-sdk/cct/evm`); its bottom `export type` block is the public surface, and every read sibling exports both `…Params` and `…Result` there. `GetAllowedFinalityConfigParams/Result` and `SetAllowedFinalityConfigParams` are only *imported* to type the facade — a consumer can't name the return type of a public method.

**Proof (compiler):** a consumer importing them fails `tsc` with **TS2459** ("declares locally, but not exported").

**Fixed + proven:** added the three re-exports; the same probe now compiles clean.

aelmanaa:
## 🟠 Major — `setAllowedFinalityConfig` is a full `bytes4` replace, but the public docs read as additive

The param JSDoc and both manager `@remarks` said only "Set `finalitySafe` to enable FCR independently of the depth" — which reads like a partial update. It is not: a caller who sets `{ finalityDepth: 20 }` to bump FTF, intending to leave FCR alone, **silently disables FCR**.

**Proof (on-chain):** [tx `0x228cf5b1`](https://sepolia.etherscan.io/tx/0x228cf5b1b36de572010fd4f5cc21885ef8423ab92cdf987d88558f73a01d50d2) — FCR went `true → false` with no error.

**Fixed + proven:** param JSDoc, both manager `@remarks`, and the `FinalityAllowed` type doc now state the full-replace semantics and point at read-modify-write via `getAllowedFinalityConfig`; the live proof confirms the documented behavior is accurate; build stays green.

*Note on fix blast radius:* the `FinalityAllowed` doc change touches `extra-args.ts`, which is shared with the extra-args codec used by `ccipSend` paths — the change is documentation-only, no behavior impact.

aelmanaa:
## 🔵 Nit — Get version-gate uses strict `=== V2_0_0`; set uses floor-match `resolveEncoder`

Identical today (only V2_0_0 exists), but a future `V2_1_0` pool would keep working through `set` while `get` rejects it as unsupported. Forward-latent divergence from the `resolveEncoder` pattern; not fixed (no `V2_1_0` exists to prove against — flagging, not prescribing an unverified change). Consider a read-side floor-match if/when a second pool version lands.

aelmanaa:
## 🔵 Nit — `get` test covers only BurnMint; the LockRelease read path is untested

`readTokenPoolAllowedFinality` hardcoding the BurnMint ABI is correct (arg-less base-`TokenPool` getter, identical selector across both v2.0.0 families), but a LockRelease case in `get-allowed-finality-config.test.ts` would lock it. No code change.

aelmanaa:
## 🔵 Nit — Decode drops reserved flag bits

`FinalityCodec` reserves bits 17–31 ("accepted on the wire"); SDK decode (`flags & 0x1`) ignores them, so a read-modify-write would clear any reserved bit a pool had set. Benign while reserved bits are unassigned; note for when they gain meaning.

aelmanaa:
## No changelog finding — and why

The branch adds no `CHANGELOG.md` entry — but neither does any sibling CCT op: `cct-sdk`'s `[Unreleased]` block has no CCT entries at all (the token/pool/admin ops landed via PRs #382/#386/#393/#402/#406/#413… without one). The CCT stream is changelogged once, when `cct-sdk` merges to main. Requiring 11552 alone to add an entry would impose a convention its siblings don't follow, so this is not flagged.

---

## Fixes applied & proven

 `npm run check` clean, tests pass.

- **Comment 1** — `cct/evm/index.ts`: export the three types. Proof: TS2459 → clean.
- **Comment 2** — `set-allowed-finality-config.ts` param JSDoc, `index.ts` ×2 manager `@remarks`, `extra-args.ts` `FinalityAllowed` doc: full-replace semantics. Proof: live Sepolia tx [`0x228cf5b1`](https://sepolia.etherscan.io/tx/0x228cf5b1b36de572010fd4f5cc21885ef8423ab92cdf987d88558f73a01d50d2) + green build.

## Recommendation

Merge after Comments 1–2 (both fixed + proven here). The encoding is correct against the pinned contract and live on Sepolia; the ops are consistent with siblings. Rebase on `cct-sdk` at merge time — `index.ts` is high-churn. Nits 3–5 are optional follow-ups.

mervin-link:
> ## No changelog finding — and why
> The branch adds no `CHANGELOG.md` entry — but neither does any sibling CCT op: `cct-sdk`'s `[Unreleased]` block has no CCT entries at all (the token/pool/admin ops landed via PRs #382/#386/#393/#402/#406/#413… without one). The CCT stream is changelogged once, when `cct-sdk` merges to main. Requiring 11552 alone to add an entry would impose a convention its siblings don't follow, so this is not flagged.
> 
> ## Fixes applied & proven
> `npm run check` clean, tests pass.
> 
> * **Comment 1** — `cct/evm/index.ts`: export the three types. Proof: TS2459 → clean.
> * **Comment 2** — `set-allowed-finality-config.ts` param JSDoc, `index.ts` ×2 manager `@remarks`, `extra-args.ts` `FinalityAllowed` doc: full-replace semantics. Proof: live Sepolia tx [`0x228cf5b1`](https://sepolia.etherscan.io/tx/0x228cf5b1b36de572010fd4f5cc21885ef8423ab92cdf987d88558f73a01d50d2) + green build.
> 
> ## Recommendation
> Merge after Comments 1–2 (both fixed + proven here). The encoding is correct against the pinned contract and live on Sepolia; the ops are consistent with siblings. Rebase on `cct-sdk` at merge time — `index.ts` is high-churn. Nits 3–5 are optional follow-ups.

@aelmanaa Addressed all comments except for the last nit. I also updated the getAllowedFinalityConfig to use floor-match `resolveEncoder` because I notice that there's already a [2.1.0-beta.0](https://www.npmjs.com/package/@chainlink/contracts-ccip/v/2.1.0-beta.0).

aelmanaa (approved):

PR #431, opened 2026-09-21
fix(cct-sdk): Validate lockbox before authorizing callers

apedrob:
## What
- `authorizeLockboxCallers` now reads `typeAndVersion` on the supplied `lockbox` before building calldata, rejecting an EOA, an undeployed address, or a different contract with a typed error

## Why
- [TOB-CLCCT-3](https://smartcontract-it.atlassian.net/browse/DAPP-11573): `applyAuthorizedCallerUpdates` sent to an address with no code mines successfully and does nothing, so the SDK handed back a confirmed tx hash for an authorization that never happened


aelmanaa:
# Review — PR #431 (DAPP-11573) `feat(cct-sdk): Validate lockbox before authorizing callers`


- **Branch:** `fix/DAPP-11573-validate-lockbox-deployment` → `cct-sdk` · **Reviewed head:** `777be9e1` (1 commit) · **Base:** `cct-sdk` @ `11a6b1bc`
- **Verdict: APPROVE the change as-is** — the added pre-flight is correct, contract-faithful, well-tested, and proven live. Plus **one MAJOR** (owner gate, §5) and **one MINOR** (a missing `@throws`, §6). Proposed changes have been **implemented in a seprate review worktree, `npm run check`'d green, unit-tested (29/29), and live-proven before→after on Sepolia** (§5) — ready to push into this PR. Neither regresses anything; neither blocks merge of the pre-flight the PR already adds.
- **Change:** `AuthorizeLockboxCallers.buildUnsigned` becomes `async` and calls a new `assertLockbox(op, chain, lockbox)` before encoding calldata. `assertLockbox` reads `chain.typeAndVersion(lockbox)` and throws a typed error unless the address is a deployed `ERC20LockBox` at a supported version (`2.0.0`). The read failure test `isMissingFunction` (`CALL_EXCEPTION | BAD_DATA`) is hoisted from `token/contracts.ts` into `query.ts` so both layers share it.
- **Method:** three-lens code review (type-safety, architecture/DRY, DX/errors) + on-chain grounding against the `chainlink-ccip` Solidity + live proofs on Ethereum Sepolia (deploy a real lockbox, then exercise every path).

---

## 1. What the op protects, and why the check is right

`ERC20LockBox.applyAuthorizedCallerUpdates` adds/removes authorized callers; a `LockReleaseTokenPool` (and any liquidity depositor) must be an authorized caller before it can lock/release. The pre-PR op built calldata to a caller-supplied `lockbox` address with **no read**. The EVM answers a call to an address with no code by succeeding and doing nothing — so a lockbox typo, an EOA, or a pool address would mine as a "successful" authorization and the SDK would hand back a confirmed tx hash for an update that never happened; the pool's first `lockOrBurn` is what finally reverts `UnauthorizedCaller`. One `eth_call` (`typeAndVersion`) turns that silent failure into a typed error before any calldata is built — and, because it lives in `buildUnsigned`, the offline/multisig (`generateUnsigned…`) path is guarded too.

---

## 2. On-chain ground truth (`chainlink-ccip`, v2.0.0) — quoted

**`typeAndVersion`** — `chains/evm/contracts/pools/ERC20LockBox.sol:19-21`:

```solidity
function typeAndVersion() external pure virtual override returns (string memory) {
  return "ERC20LockBox 2.0.0";
}
```

Parsed → type `ERC20LockBox`, version `2.0.0`. The SDK's `contractType === 'ERC20LockBox'` + `LOCKBOX_VERSIONS = ['2.0.0']` map 1:1. `ERC20LockBox` is a **v2-only** contract — a repo-wide grep of `chains/evm/contracts` (non-test) returns this single version string, so there is no legacy lockbox to fall back to and a hard-fail on any other version is correct.

**Owner-gated write** — `ERC20LockBox` is `AuthorizedCallers` (`ERC20LockBox.sol:16`), and the method is `onlyOwner` (`shared/access/AuthorizedCallers.sol:44-46`):

```solidity
function applyAuthorizedCallerUpdates(
  AuthorizedCallerArgs memory authorizedCallerArgs
) external onlyOwner {
```

The owner guard reverts `OnlyCallableByOwner()` (`OwnerIsCreator`/`ConfirmedOwnerWithProposal`), selector `0x2b5c74de`.

**The caller invariant** — deposit/withdraw both end in `_validateCaller()` (`ERC20LockBox.sol:102`, via `_validateDepositWithdraw`), which reverts `UnauthorizedCaller(address)` (`AuthorizedCallers.sol:16,78-82`), selector `0xd86ad9cf`. Both selectors are observable directly in the deployed runtime bytecode (§4, tx `0xad39…`). This is the class of silent failure the pre-flight prevents from being mistaken for success.

---

## 3. The fix (inline diff)

`ccip-sdk/src/cct/evm/lockbox/contracts.ts` — new `assertLockbox`:

```ts
const LOCKBOX_TYPE = 'ERC20LockBox'
const LOCKBOX_VERSIONS: string[] = ['2.0.0']

export async function assertLockbox(operation: string, chain: EVMChain, lockbox: string): Promise<void> {
  let contractType: string, version: string
  try {
    ;[contractType, version] = await chain.typeAndVersion(lockbox)
  } catch (err) {
    if (!isMissingFunction(err)) throw err          // transport/rate-limit propagates untouched
    throw new CCTParamsInvalidError(operation, 'lockbox',
      `nothing at ${lockbox} answers typeAndVersion() — …deploy a lockbox with deployLockbox`,
      { cause: err instanceof Error ? err : undefined })
  }
  if (contractType !== LOCKBOX_TYPE) throw new CCTContractTypeInvalidError(lockbox, LOCKBOX_TYPE, contractType)
  if (!LOCKBOX_VERSIONS.includes(version))
    throw new CCTContractVersionUnsupportedError(LOCKBOX_TYPE, version, { context: { address: lockbox } })
}
```

`ccip-sdk/src/cct/evm/lockbox/operations/authorize-callers.ts` — `buildUnsigned` now `async`, asserts first:

```ts
protected async buildUnsigned(chain: EVMChain, { lockbox, addedCallers = [], removedCallers = [] }: …): Promise<UnsignedEVMTx> {
  await assertLockbox(this.name, chain, lockbox)
  const data = LOCKBOX_INTERFACE.encodeFunctionData('applyAuthorizedCallerUpdates', [{ addedCallers, removedCallers }])
  …
}
```

`isMissingFunction` moved from `token/contracts.ts` into `query.ts` (shared, imported by both) — no duplicate left behind.

aelmanaa:
## 4. Live proofs (Ethereum Sepolia, chainId 11155111) — verbatim

Fixtures, each a clickable Sepolia Etherscan link:

| Thing | Address / tx |
|---|---|
| Deployer = lockbox owner | [`0x9d087fC03ae39b088326b67fA3C788236645b717`](https://sepolia.etherscan.io/address/0x9d087fC03ae39b088326b67fA3C788236645b717) |
| `deployLockbox` tx (valid fixture, mined) | [`0xad3921bce9a3aa5e1dc341ec9dbfe514c379722dbc5258d51535643db1a600eb`](https://sepolia.etherscan.io/tx/0xad3921bce9a3aa5e1dc341ec9dbfe514c379722dbc5258d51535643db1a600eb) |
| Deployed `ERC20LockBox 2.0.0` (valid fixture) | [`0x78AF0E243584aB8cb75f7477112046D6F8761095`](https://sepolia.etherscan.io/address/0x78AF0E243584aB8cb75f7477112046D6F8761095) |
| `token` it escrows (CrossChainToken 2.0.0) | [`0x227f0db66e9bf42a28e1e38438c5f19d0adb2df0`](https://sepolia.etherscan.io/address/0x227f0db66e9bf42a28e1e38438c5f19d0adb2df0) |
| Deployed non-lockbox fixture (Router 1.2.0) | [`0x0BF3dE8c5D3e8A2B34D2BEeB17ABfCeBaf363A59`](https://sepolia.etherscan.io/address/0x0BF3dE8c5D3e8A2B34D2BEeB17ABfCeBaf363A59) |

### 4.1 `assertLockbox` outcomes — valid builds, the three bad shapes throw the right typed error (live reads)

```
=== deploy a real ERC20LockBox (valid fixture) ===
deployLockbox tx: 0xad3921bce9a3aa5e1dc341ec9dbfe514c379722dbc5258d51535643db1a600eb
lockbox address:  0x78AF0E243584aB8cb75f7477112046D6F8761095
typeAndVersion:   ERC20LockBox 2.0.0

=== assertLockbox outcomes (all live reads against Sepolia) ===
valid lockbox (just deployed)      : BUILT — tx to 0x78AF0E243584aB8cb75f7477112046D6F8761095, data len 394
no-code EOA (deployer address)     : threw CCTParamsInvalidError ctx={"operation":"authorizeLockboxCallers","param":"lockbox","reason":"nothing at 0x9d087fC03ae39b088326b67fA3C788236645b717 answers typeAndVersion() — …"} cause=BAD_DATA
deployed non-lockbox (Router 1.2.0): threw CCTContractTypeInvalidError ctx={"address":"0x0BF3dE8c5D3e8A2B34D2BEeB17ABfCeBaf363A59","expected":"ERC20LockBox","actual":"Router"}
```

### 4.2 The silent-failure the check closes — `applyAuthorizedCallerUpdates` to a no-code address mines as a no-op (live)

Raw calldata `eth_call`ed against Sepolia, bypassing the guard:

```
=== footgun proof: without the guard, applyAuthorizedCallerUpdates to an EOA mines as a no-op ===
eth_call EOA (0x9d087fC03ae39b088326b67fA3C788236645b717): SUCCEEDED ret=0x estimateGas=23567 — no revert, no code ran (silent no-op)
eth_call undeployed 0x...dead (0x000000000000000000000000000000000000dEaD): SUCCEEDED ret=0x estimateGas=23567 — no revert, no code ran (silent no-op)
```

This is exactly why a confirmed tx hash from the pre-PR op meant nothing — and why the read is worth it.

### 4.3 Verification summary

| Claim | Evidence | Live |
|---|---|---|
| Valid `ERC20LockBox 2.0.0` builds calldata | §4.1 "BUILT", data len 394 | ✅ |
| No-code address → `CCTParamsInvalidError(param=lockbox, cause=BAD_DATA)` | §4.1 | ✅ |
| Deployed wrong contract (Router) → `CCTContractTypeInvalidError(expected=ERC20LockBox, actual=Router)` | §4.1 | ✅ |
| Without the guard, the call to a no-code address succeeds silently (`ret=0x`) | §4.2 `eth_call` | ✅ |
| `typeAndVersion` == `ERC20LockBox 2.0.0`; selectors `0xd86ad9cf`/`0x2b5c74de` in runtime bytecode | §2 + tx `0xad39…` | ✅ |

aelmanaa:
## 5. MAJOR (recommend folding into this PR) — the op still does not gate `sender` against the lockbox owner

`applyAuthorizedCallerUpdates` is `onlyOwner` on-chain (§2). `AuthorizeLockboxCallersParams` already carries a `sender?` field documented **"Lockbox owner; sets `tx.from` for offline / multisig signing"**, and base `generate` writes it to `tx.from`. But `buildUnsigned` never checks that `sender` is the owner — whereas **every sibling owner-gated write does**, in `buildUnsigned`, via a `readXOwner` + assert: `assertPoolOwner` (`set-dynamic-config`, `set-rate-limit-admin`, `apply-allowlist-updates`, `add-remote-pool`, `transfer-liquidity`), `assertTokenOwner` (`grant-mint-role`, `revoke-burn-role`), `assertPoolOwnershipTransfer` / `assertTokenOwnershipTransfer` (the transfer ops).

So an unauthorized `sender` sails through offline-tx generation and reverts `OnlyCallableByOwner` only **after** a multisig has reviewed and signed — the *same* silent-until-too-late failure mode this PR exists to remove, one level up. Proven live against the real lockbox from §4:

```
=== owner-gate gap: authorize builds for a NON-owner sender, then reverts OnlyCallableByOwner on-chain ===
lockbox owner():          0x9d087fc03ae39b088326b67fa3c788236645b717
non-owner sender used:    0x00000000000000000000000000000000000000B2
generateUnsigned with non-owner sender: BUILT — tx.from=0x00000000000000000000000000000000000000B2 (no owner check)
eth_call from non-owner: REVERTED selector=0x2b5c74de (OnlyCallableByOwner()=0x2b5c74de)
```

The gap is **pre-existing** (the old sync `buildUnsigned` gated nothing) and this diff does not regress it, so it is not a merge blocker.

### Fix — implemented, built, tested, and live-proven (in the review worktree)

Added `readLockboxOwner` + `assertLockboxOwner` to `lockbox/contracts.ts` (mirroring `readTokenPoolOwner` + `assertPoolOwner`), gated it in `buildUnsigned` when `sender !== undefined`, and added an `execute` override that defaults `sender` to the signing wallet (mirroring `SetDynamicConfig`) so the broadcast path is gated too:

```ts
// lockbox/contracts.ts
export async function assertLockboxOwner(operation: string, chain: EVMChain, lockbox: string, sender: string): Promise<void> {
  const owner = await readLockboxOwner(chain, lockbox)   // owner() via getTypedContract, checksummed
  if (getAddress(sender) === owner) return
  throw new CCTParamsInvalidError(operation, 'sender', `must be the current lockbox owner (${owner})`)
}

// authorize-callers.ts — buildUnsigned
await assertLockbox(this.name, chain, lockbox)
if (sender !== undefined) await assertLockboxOwner(this.name, chain, lockbox, sender)

// authorize-callers.ts — new execute override defaults sender to the wallet
override async execute(chain, params) {
  const sender = await this.resolveWalletSender(params.wallet, params.sender)
  return super.execute(chain, { ...params, sender })
}
```

`npm run check` → **EXIT=0**; `authorize-callers.test.ts` → **29/29** (the 25 above + 4 new owner-gate tests: reads `owner()` only after `typeAndVersion` and only when `sender` given; builds for the owner; rejects a non-owner `sender`; `execute` rejects and never broadcasts when the wallet is not the owner).

**Live before→after** against the same real lockbox `0x78AF0E24…1095` (on-chain `owner()` = deployer):

```
BEFORE (bug, §5 above): non-owner sender → BUILT, then eth_call REVERTED 0x2b5c74de OnlyCallableByOwner (post-sign only)
AFTER (fix built + run live):
on-chain owner():                        0x9d087fc03ae39b088326b67fa3c788236645b717
non-owner sender (0x…B2):                threw CCTParamsInvalidError param=sender — rejected before any tx is produced
owner sender (0x9d087fC0…b717):          BUILT — tx.from=0x9d087fC0…b717, data len 394  (happy path preserved)
```

The unauthorized `sender` that previously produced a clean tx (reverting only after a multisig signs) is now rejected at build; the real owner still builds. Ready to fold into this PR.

---

## 6. MINOR — public `@throws` omit `CCIPTypeVersionInvalidError`

`assertLockbox` can raise a fourth error: when the contract answers with a string that is not a `type version` pair, `chain.typeAndVersion` throws `CCIPTypeVersionInvalidError` (it is not a `isMissingFunction` shape, so it propagates raw). This is documented on `assertLockbox` itself and explicitly tested ("rejects a contract whose typeAndVersion string is unparseable"), so it is reachable on the public path — but it is missing from the `@throws` of `EVMTokenManager.generateUnsignedAuthorizeLockboxCallers`, `.authorizeLockboxCallers`, and `buildUnsigned`.

**Fix — applied** in the worktree: added `@throws {@link CCIPTypeVersionInvalidError}` to both public manager methods and to `buildUnsigned`, and folded the owner-check throw into their `CCTParamsInvalidError` lines. Doc-only; `npm run check` green (§8). Cosmetic; not blocking.

aelmanaa:
## 7. Verified consistent (no action)

- **Placement:** `buildUnsigned` is the correct layer — every sibling owner/type pre-flight does its async read there (not sync `validate`, not `execute`), and base `EVMOperation.generate` runs `buildUnsigned` for both the offline and execute paths, so the multisig path is guarded exactly as claimed.
- **Hard-fail vs soft:** `assertLockbox` hard-fails a missing/reverting read, unlike `assertOwnable2StepToken` (which soft-returns "proceed as v1.x"). Correct: that op is soft only because v1.5.1 tokens predate `typeAndVersion()`; no legacy lockbox exists, so a missing read genuinely means "no lockbox here." It mirrors `readTokenRole`, which likewise treats `isMissingFunction` as a hard type error and threads `{ cause }`.
- **DRY:** hoisting `isMissingFunction` into `query.ts` (the "EVM CCT reads" module) is the right neutral shared home; both consumers import it, no cross-module method import. `LOCKBOX_VERSIONS = ['2.0.0']` reasonably simplifies the token-pool version-table (one ABI, nothing dispatches on lockbox version) rather than copying it.
- **Errors/DX:** the three messages are actionable and match sibling wording; the `CCTParamsInvalidError` "code that is not a lockbox" phrasing is not misleading — that branch only fires on BAD_DATA/CALL_EXCEPTION (nothing answered the probe); a deployed wrong contract is routed to the distinct `CCTContractTypeInvalidError` (proven in §4.1).

---

## 8. Types / build / tests

Built the branch head standalone in a clean worktree (`npm ci` → `npm run check`; toolchain is oxfmt + oxlint + tsc):

```
npm run check → EXIT=0   (no errors, no warnings)
```

`ccip-sdk/src/cct/evm/lockbox/operations/authorize-callers.test.ts` (node --test):

```
ℹ tests 25
ℹ pass 25
ℹ fail 0
ℹ skipped 0
```

The 25 include the new 7-test "lockbox pre-flight" suite (reads `typeAndVersion` first; rejects no-code/BAD_DATA; reverting CALL_EXCEPTION with `cause` preserved; propagates TIMEOUT unwrapped; unparseable version string → `CCIPTypeVersionInvalidError`; wrong type; unsupported version) and "does not sign or broadcast when lockbox not deployed". No `any`/unsafe `as` in the diff; the sync→async `buildUnsigned` is a permitted narrowing of the base's `Promise<UnsignedEVMTx> | UnsignedEVMTx` return, already `await`ed by `generate`. No type-safety findings.

**After the §5/§6 fixes** (same worktree): `npm run check` → **EXIT=0**; the op suite → **29/29** (adds the 4 owner-gate tests). So both the PR as submitted and the PR + proposed fixes build and test green.

### Verification — findings AND fixes

| Item | Finding proven live | Fix built + `check` green | Fix proven live |
|---|---|---|---|
| §5 owner gate not enforced | ✅ non-owner `sender` builds, reverts `0x2b5c74de` on-chain | ✅ `assertLockboxOwner` + `execute` override; 29/29 | ✅ non-owner `sender` → `CCTParamsInvalidError param=sender` at build; owner still builds |
| §6 `@throws` omits `CCIPTypeVersionInvalidError` | ✅ reachable, test-covered | ✅ added to 2 public methods + `buildUnsigned`; `check` green | n/a (doc-only) |

---

## 9. Recommendation

**Approve.** The added lockbox pre-flight is correct, contract-faithful (§2), DRY/consistent with siblings (§7), and proven live at every path (§4). Two improvements — the **owner-gate MAJOR** (§5) and the **`@throws` MINOR** (§6) — are already implemented in the review worktree, `npm run check` green, unit-tested **29/29**, and (for §5) live-proven before→after on Sepolia; both extend the exact "no silent multisig failure" guarantee this PR is about. Recommend folding them in before merge; if the author prefers, they can ship as a fast follow — nothing here blocks the current diff.

mervin-link (approved):

